PR #79, opened 2015-01-06
Added podspec file

dclelland:
I made a podspec file, so you can install AudioKit using CocoaPods...

Haven't tested on OS X yet, but working just fine on iOS with the following Podfile and the AudioKit repo in a neighbouring directory.

    source 'https://github.com/CocoaPods/Specs.git'
    platform :ios, '8.0'
    pod 'AudioKit', :path => '../AudioKit'

### DIFF
--- a/AudioKit.podspec
+++ b/AudioKit.podspec
@@ -4,7 +4,7 @@ Pod::Spec.new do |spec|
     spec.authors                  = { 'YOUR NAME HERE' => 'YOUR EMAIL HERE' }
     spec.license                  = { :type => 'LGPL' }
     spec.homepage                 = 'http://audiokit.io/'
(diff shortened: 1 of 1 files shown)


aure:
Hi Daniel,

I've wanted to do the CocoaPods thing since release 1.1, but I haven't had time to research what I had to.  So, first of all thanks very much.  Next, I looked at the code, and it looks like it points to the git repo at aure/AudioKit instead of the more canonical audiokit/AudioKit so I wanted to double check with you on that choice.  

Next, what will this mean for the installation of AudioKit... basically what does this give the end user?

Thanks again,
Aure

aure:
Oh I get it, this will automatically create the AudioKit directory side by side to a project that requires it right?  Cool... so just the question about the git URL...

PR #81, opened 2015-01-21
Easy exponential curves

adambmedia:
Added FunctionTable for exponential curves with tunable positive and negative second order derivatives yielding concave up and down curves with varying degrees of bend.

### DIFF
--- a/AudioKit/Core Classes/AKFoundation.h
+++ b/AudioKit/Core Classes/AKFoundation.h
@@ -109,6 +109,7 @@
 #import "AKLowFrequencyOscillator.h"
 #import "AKOscillator.h"
 #import "AKPhasor.h"
+#import "AKPhasorControl.h"
 #import "AKVCOscillator.h"
 
 // Operations - Signal Generators - Physical Models
--- a/AudioKit/Core Classes/AKFoundation.h
+++ b/AudioKit/Core Classes/AKFoundation.h
@@ -48,6 +48,7 @@
 #import "AKAdditiveCosineTable.h"
 #import "AKArrayTable.h"
 #import "AKExponentialCurves.h"
+#import "AKExponentialCurvesVariableGrowth.h"
 #import "AKFunctionTable.h"
 #import "AKLineSegments.h"
 #import "AKRandomDistributionTable.h"
--- a/AudioKit/Core Classes/AKTypes.h
+++ b/AudioKit/Core Classes/AKTypes.h
@@ -128,6 +128,7 @@ typedef NS_OPTIONS(NSUInteger, AKFunctionTableType)
     AKFunctionTableTypeSoundFile = 1,
     AKFunctionTableTypeArray=2,
     AKFunctionTableTypeAdditiveCosines=11,
+    AKFunctionTableTypeExponentialCurvesVariableGrowth=16,
     AKFunctionTableTypeWeightedSumOfSinusoids=19,
     AKFunctionTableTypeWindows=20,
     AKFunctionTableTypeRandomDistributions=21,
--- a/AudioKit/Operations/Function Tables/Generators/AKExponentialCurvesVariableGrowth.h
+++ b/AudioKit/Operations/Function Tables/Generators/AKExponentialCurvesVariableGrowth.h
@@ -8,6 +8,28 @@
 
 #import "AKFunctionTable.h"
 
+/** Constructs concatenated power functions with variable rates of growth or decay.
+ */
 @interface AKExponentialCurvesVariableGrowth : AKFunctionTable
+
+/// Size of the table (default 4096)
+@property int size;
+
+/// Start the power curves at a specific value
+/// @param value Initial value of the first segment
+-(instancetype)initWithValue:(float)value;
+
+/// Add a junction point
+/// @param value The value at the given index
+/// @param index The index at which the value will be set
+- (void)addValue:(float)value atIndex:(int)index;
+
+/// Add a junction point
+/// @param value The value at the given index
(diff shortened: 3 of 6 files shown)


aure:
Reviewing this now.  AKPhasingControl was removed from AudioKit because AKPhasor can be used at k-rate without issue as part of rate agnosticism.  I don't think its needed.  Please rename AKExponentialCurvesVariableGrowth toe AKVariableGrowthExponentialCurves.

adambmedia:
I couldn’t use AKPhasor as the index argument in AKTableValueControl as follows:

AKTableValueControl *tableRead;
tableRead = [[AKTableValueControl alloc] initWithFunctionTable:myTable atIndex:myPhasorControl];

-adam

--
Adam Boulanger
co-founder, Hear For Yourself
t. 617-990-2704
e. adam@hearforyourself.com
w. www.adamboulanger.com

On January 21, 2015 at 3:50:31 PM, Aurelius Prochazka (notifications@github.com) wrote:

Reviewing this now. AKPhasingControl was removed from AudioKit because AKPhasor can be used at k-rate without issue as part of rate agnosticism. I don't think its needed. Please rename AKExponentialCurvesVariableGrowth toe AKVariableGrowthExponentialCurves.

adambmedia:
Don’t you think AKVariableGrowthExponentialCurves is inferior in terms of autocomplete?  I’m reaching for an exponential curve not a variable growth.
-adam

--
Adam Boulanger
co-founder, Hear For Yourself
t. 617-990-2704
e. adam@hearforyourself.com
w. www.adamboulanger.com

On January 21, 2015 at 3:50:31 PM, Aurelius Prochazka (notifications@github.com) wrote:

Reviewing this now. AKPhasingControl was removed from AudioKit because AKPhasor can be used at k-rate without issue as part of rate agnosticism. I don't think its needed. Please rename AKExponentialCurvesVariableGrowth toe AKVariableGrowthExponentialCurves.

aure:
Do you need to have "VariableGrowth" in these names at all, why not just AKExponentialCurves?

adambmedia:
Hmmm, there is a better name here but not VariableGrowth.  I needed something where I could customize the second order derivative of an otherwise exponential curve.  The existing exponential curves where insufficient.  It’s the difference between GEN 16 and 25 and its a big difference.
-adam

--
Adam Boulanger
co-founder, Hear For Yourself
t. 617-990-2704
e. adam@hearforyourself.com
w. www.adamboulanger.com

On January 21, 2015 at 3:57:14 PM, Aurelius Prochazka (notifications@github.com) wrote:

Do you need to have "VariableGrowth" in these names at all, why not just AKExponentialCurves?

aure:
AudioKit has human readable class  names not alphabetical ones. Otherwise we'd be naming things AKEnvelopeLinear instead of AKLinearEnvelope, etc.

aure:
Because things are different numbered GENs in Csound don't necessitate that they have to exist in AudioKit that way.  If you're building exponential curves with different methods, that's great, the end result is what we're naming, not the method by which we get there.

adambmedia:
I know what audiokit has.  God dammit Aure.  A Linear Envelope makes sense as a thing because its concise and is readable.  When we’re talking about similar classes with nuanced mathematical differences, that rationale falls apart.  The result is that you’re constantly checking he docs because autocomplete fails you.  For instance, the thought process when designing sound is
1. I know what kind of function I want, it’s an exponential curve where I can control the bend
2. is that, growth, derivative, inflection point, gah, all are equally appropriate, which ones did audiokit choose
3. google or look at the documentation

Better is, I type expon... and tab through the possibilities.  Actually, this way, I may discover alternatives.

Makes sense, no? -adam

--
Adam Boulanger
co-founder, Hear For Yourself
t. 617-990-2704
e. adam@hearforyourself.com
w. www.adamboulanger.com

On January 21, 2015 at 3:57:14 PM, Aurelius Prochazka (notifications@github.com) wrote:

Do you need to have "VariableGrowth" in these names at all, why not just AKExponentialCurves?

adambmedia:
You would still argue that it is human readable. -adam

--
Adam Boulanger
co-founder, Hear For Yourself
t. 617-990-2704
e. adam@hearforyourself.com
w. www.adamboulanger.com

On January 21, 2015 at 4:02:14 PM, Aurelius Prochazka (notifications@github.com) wrote:

AudioKit has human readable class names not alphabetical ones. Otherwise we'd be naming things AKEnvelopeLinear instead of AKLinearEnvelope, etc.

adambmedia:
Totally makes sense. I’m naming to describe what the thing does and how it would be imagined by someone wanting to use it and looking for it. Where do you see a historical reference to GEN routine naming, which is just a number, no? -adam

--
Adam Boulanger
co-founder, Hear For Yourself
t. 617-990-2704
e. adam@hearforyourself.com
w. www.adamboulanger.com

On January 21, 2015 at 3:57:14 PM, Aurelius Prochazka (notifications@github.com) wrote:

Do you need to have "VariableGrowth" in these names at all, why not just AKExponentialCurves?

adambmedia:
Variable growth does make the most sense.

aure:
You introduced AKPhasorControl because AKTableValueControl had not been made rate-agnostic yet.  I fixed that so now your example should work with AKPhasor.   Thanks for finding that.   Please remove AKPhasorControl from this pull request.  I don't think my request that you change the name of the variable growth exponential curves was unreasonable either.  But, now after reviewing it further, this function table closely parallels AKSegmentArray which can be linear or curved.  I propose lumping all of these into one nice AKSegments function table which can handle linear or variable growth curves.

Issue #70, opened 2014-12-31
All OS X examples crash on launch


Nikolozi:
I tried all 4 OS X example apps. When I try to run them they all crash at the following line

int ret = csoundCompile(cs, 5, argv);

It's inside - (void)runCsound:(NSString *)csdFilePath method. Compiler gives EXC_BAD_ACCESS. When I hover over cs and argv they are both non-nil. So, I'm not sure why it's crashing.

I'm running Xcode 6.1.1 on Yosemite 10.10.1.

Up until now I've been playing with iOS example code only. I'll try to create an OS X app following the tutorial on AudioKit.io page to see if I can find out why it's crashing.

Nikolozi:
Hi Aure,

Thanks for the quick response. The funniest thing happened to me just now. As I was writing the response, describing that I don't use audio interface and don't have any drivers installed, it hit me that I have Apogee's MiC connected to my iMac via USB. I thought I'd disconnect it and see if it solves anything. Sure enough, the Mac examples started working. As soon as I connect it back they stop working.

I wonder if the example apps automatically pick the MiC as an audio interface to output sound and fail obviously since it doesn't support audio output? And as a follow up question, is there a facility in AudioKit to choose an output device if multiple audio interfaces are connected to a mac?

Cheers
Niko

aure:
Yeah, that's interesting.  I can try with my USB mic here.  I want to close this issue because the name of the issue might scare people off.  You can open a new issue with the USB mic question if you like, or else I'll just look into it and contact you directly.

Nikolozi:
Good idea regarding closing the issue. It's not something I need resolved right now. I can simply disconnect the MiC when I'm working with AudioKit stuff. If I ever need it resolved I'll re-open it as a separate issue.

Cheers
Niko

gritd:
I think it has something to do with multiple sound cards and cSound selecting the wrong default, I looked up the correct outputs in the csoundQT tool and added -i and -o arguments like this in the line above the error:
char *argv[5] = { "csound", "-+ignore_csopts=0",
            "-+rtaudio=coreaudio -i=adc0 -o=dac0","-b256", (char*)[csdFilePath
                                                    cStringUsingEncoding:NSASCIIStringEncoding]};

(adc0 for microphone input and dac0 for built-in output in my case)
All is now working perfectly.

PR #88, opened 2015-03-12
Specify ADC0 and DAC0 in csound arguments. Resolves GH-70

thisandagain:
This patch was originally suggested by @gritd in #70 but resolves a real set of issues on Mac OS 10.10.2. I was able to reproduce the issue described by either:
- Installing [SoundFlower](https://rogueamoeba.com/freebies/soundflower/)
- Connecting a FocusRite Scarlet 2i2 via USB

Attempting to run either the `HelloWorld` or `AudioKitDemo` applications in either of these states causes an `EXC_BAD_ACCESS` exception on line `450` of `CsoundObj.m`.

Hope this helps! :smile: :cat2: :cake:

### DIFF
--- a/AudioKit/Libraries/OSX/csound-OSX/CsoundObj.m
+++ b/AudioKit/Libraries/OSX/csound-OSX/CsoundObj.m
@@ -445,7 +445,7 @@ OSStatus  Csound_Render(void *inRefCon,
         //    }
         
         char *argv[5] = { "csound", "-+ignore_csopts=0",
-            "-+rtaudio=coreaudio", "-b256", (char*)[csdFilePath
+            "-+rtaudio=coreaudio -i=adc0 -o=dac0", "-b256", (char*)[csdFilePath
                                                     cStringUsingEncoding:NSASCIIStringEncoding]};
         int ret = csoundCompile(cs, 5, argv);
         mCsData.running = true;


aure:
This would overwrite the settings inside of the AudioKit.plist.  Do you think the defaults there should be adc0 and dac0?

thisandagain:
Good catch. What is curious is that removing the change to `argv` within `CsoundObj.m` and updating the `plist` seems to result in the same exception being thrown. In this case, `CsOptions` within the `csd` being generated looks like this:
```xml
<CsOptions>
-o dac0           ; Write sound to the host audio output
--expression-opt  ; Enable expression optimizations
-m0               ; Print raw amplitudes
-i adc0           ; Request sound from the host audio input device
</CsOptions>
```

Options should be recognized as `-+ignore_csopts=0` is set. Any thoughts as to why this option would seemingly be ignored? I'll keep digging this evening (PST) as well.

aure:
The other thing that is weird is that your edits specifiy -i=adc0 and -o=dac0 but according to the docs you're not supposed to use equal signs.  Perhaps they're optional?

aure:
I have confirmed on my end that changing the plist does change the outputs. Is there anything we can do to test this on your system?  Perhaps a screenshare?

aure:
There have been a bunch of changes to AudioKit since this pull request and the next version will be another big internal change, so I am going to close this pull request for now.

PR #94, opened 2015-03-30
Basic operators overloading in Swift

AlesTsurko:
Arithmetic.swift contains overloading for: +, -, *, /. This gives a possibility to use basic operators instead of plus, minus, dividedBy and scaledBy methods. It also contains a useful extension for float, which converts midi note number to ratio (aka speed of AKFileInput aka frequency of AKGranularSynthesisTexture). The last one ported from SuperCollider.


aure:
Great stuff thank you!

PR #95, opened 2015-03-30
Make the state property @dynamic (silences a warning in Xcode 6.3)

megastep:
A simple fix as the Xcode 6.3 beta warns about reusing the `state` property in this class that is already declared in a parent class. I'm assuming here that reusing that property is the desired effect.

### DIFF
--- a/AudioKit/Operations/Signal Modifiers/Volume and Spatialization/AKMix.m
+++ b/AudioKit/Operations/Signal Modifiers/Volume and Spatialization/AKMix.m
@@ -20,8 +20,6 @@
     AKParameter *current;
 }
 
-@dynamic state;
-
 - (instancetype)initWithInput1:(AKParameter *)input1
                         input2:(AKParameter *)input2
                         balance:(AKParameter *)balancePoint;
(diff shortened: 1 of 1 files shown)


aure:
Does this warning appear only with AKMix?  This pattern is repeated throughout AudioKit's operations, so I am wondering if it needs to be a more global fix.

megastep:
It is true that I used to get more of these warnings before upgrading to AudioKit 2.0, so it appears that it has gotten better with that new release. I'm only using the iOS side of things so I haven't tried compiling the OS X side of the library in my project, but this was the only warning I was still getting with the current code.

aure:
Hey so I think that we just have to remove those lines but not add the @dynamic thing.
Would you like to change your pull request to do that?  You deserve credit for finding this! :)

megastep:
Sure, the only reason I used `@dynamic` was because Xcode basically suggested to do that. It's probably not necessary if the property is properly declared in the parent class. I can amend the commits here.

megastep:
OK, new commit should take care of this!

PR #97, opened 2015-04-03
Various improvements for the *Plot views

megastep:
The main focus on this was making the presentation properties (such as line widths and colors) exposed and inspectable in Interface Builder.

At the same time I also modernized these classes quite a bit, and hid some of the internals by moving the Csound protocol adoption outside of the headers.

### DIFF
--- a/AudioKit/Utilities/Plots/AKAudioInputFFTPlot.h
+++ b/AudioKit/Utilities/Plots/AKAudioInputFFTPlot.h
@@ -13,10 +13,12 @@
 #if TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
 /// Plots the FFT of the audio input
-@interface AKAudioInputFFTPlot : UIView <CsoundBinding>
+IB_DESIGNABLE
+@interface AKAudioInputFFTPlot : UIView
 #elif TARGET_OS_MAC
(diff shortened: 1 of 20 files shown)


aure:
You absolutely rock!

megastep:
Thanks! :) By the way, I was wondering whether it would make sense to make a call to `[AKManager addBinding:self]` directly from within the constructor in these classes. This would make it unnecessary to do it manually otherwise... And this way all you'd have to do to get a functional display is to connect it in the storyboard.

aure:
Yes, I had thought of that recently, while I was upgrading the examples.  Totally the right thing to do.  The other thing I want to do is consolidate a lot of the repeated code over similar plots.  But, I have a bunch of other stuff to do if you want to be plot-master for a while, that would be so appreciated.

megastep:
I could do that. Actually just looking at the code I think it would make more sense to do the binding in `didMoveToSuperview:` so that the init is fully done. I don't see an unbind method on `AKManager`, is that intentional? It would make sense to also remove it once the view is deleted or gone.

aure:
That's sounds really great.  I'd probably call it "removeBinding" to parallel the "addBinding" method but it sounds like a very good idea.

PR #98, opened 2015-04-03
Reworked the plot views so that explicit addBinding: calls to AKManager are no longer needed

megastep:
As we discussed in my last pull request, here is what I've done:

* Added a new `removeBinding:` class method to AKManager
* New `AKPlotView` class that is the base of all plot classes, abstracting away a lot of the boilerplate and automatically registering the objects with AKManager when views are added/removed from the hierarchy.
* This also greatly simplified the differences between iOS and OS X.
* Fixed a few memory leaks in several of these classes (particularly calls to `malloc` without a corresponding `free`). There are actually quite a few outstanding ones where pointers get obviously overwritten by NSData, I haven't touched that as I'm not sure what the intent is, but I made some FIXME notes.

### DIFF
--- a/AudioKit/Core Classes/AKManager.h
+++ b/AudioKit/Core Classes/AKManager.h
@@ -98,4 +98,7 @@
 /// @param binding The object that will be added to Csound's binding list
 + (void)addBinding:(id)binding;
 
+/// @param binding The object that will be removed from Csound's binding list
++ (void)removeBinding:(id)binding;
+
 @end
--- a/AudioKit/Core Classes/AKManager.m
+++ b/AudioKit/Core Classes/AKManager.m
@@ -331,4 +331,9 @@ static AKManager *_sharedManager = nil;
     [[[self sharedManager] engine] addBinding:binding];
 }
 
++ (void)removeBinding:(id)binding
+{
+    [[[self sharedManager] engine] removeBinding:binding];
+}
+
 @end
--- a/AudioKit/Utilities/Plots/AKAudioInputFFTPlot.h
+++ b/AudioKit/Utilities/Plots/AKAudioInputFFTPlot.h
@@ -12,7 +12,11 @@
 IB_DESIGNABLE
 @interface AKAudioInputFFTPlot : AKPlotView
 
-@property IBInspectable AKColor *lineColor;
+#if TARGET_OS_IPHONE
+@property IBInspectable UIColor *lineColor;
+#else
+@property IBInspectable NSColor *lineColor;
+#endif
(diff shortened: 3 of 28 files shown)


aure:
Wow so good!

aure:
What I love about this is that its much cleaner now.  In case you are interested in doing more things from the AudioKit to do list, I'll just let you know that we'd like to consolidate the most common drawing routines into a superclass, either AKPLotView or another file.  Also, we don't want to rely on the EZAudio Rolling Plot stuff so we'd like to make our own.  Anyway, you've been so helpful, just thought you might want to know where things are headed.  As for the memory leaks, I'd love to get your expert eyes on those for sure.  I really want this to be robust.

megastep:
Yeah, there's definitely room for improvement still as I noticed a lot of duplicated code between all those classes as I was going through them.

I tried to stick with the UI code as much as possible instead of messing with the processing code, so I'd have to take a closer look at memory management in there. You definitely don't want to use one of these views and have them slowly killing your app by leaking memory. :) It's a good habit when using something like `malloc()` to write the corresponding `free()` at the same time so you don't forget later.

aure:
Thanks for the tip.  I have high hopes for the graphing capabilities here and if there's a solid core, it will easier to build features on top of.  I wanted to avoid require a package like CorePlot but we can at least see what they have that we want - tick marks, interactivity (pinch), etc.

megastep:
Another thing I've been considering doing is making a subproject for a static library. It's kinda painful to basically have to re-add every single source file to your project when the structure of the library changes. It would be better solved by an AudioKit framework ideally, but a static library that can be included as a subproject in other projects would work well. We could even have separate iOS and OS X targets as I've seen in some other libraries.

This is mostly Xcode configuration stuff though, shouldn't really affect the code of the library itself. For a good example of that, I've been using the ObjectAL project as an audio library in my own apps for a while (obviously I want to switch to AudioKit!) and that's how it's structured.

https://github.com/megastep/ObjectAL-for-iPhone

aure:
Great idea. I'd really to have a static library or framework but I wasn't able to get it right and released what I had because it works for now.  The simpler it is to plug in AudioKit to a project, the better.  I'm also willing to update the docs and escort the people already using AudioKit into any changes that are  improvements.  So, definitely, proceed with my fullest blessing and support.

megastep:
I think I'll tackle the `malloc()` abuse first, seems to be mostly in the plot classes with a couple of other classes. I think I have a good solution for that using `NSData` instead. Better to go the Objective-C way and let ARC handle that for you.

PR #100, opened 2015-04-04
More cleanups in plots, fixed a bunch of memory leaks

megastep:
* Moved more of the duplicate code to the parent `AKPlotView` class
* Audited calls to `malloc()` to make sure that memory is not leaked, replaced by direct access to the `NSData` buffer when possible.
* Added convenience `AKPlots.h` for use in Swift bridging headers, etc.

### DIFF
--- a/AudioKit/Core Classes/AKManager.m
+++ b/AudioKit/Core Classes/AKManager.m
@@ -84,8 +84,8 @@ static AKManager *_sharedManager = nil;
         NSString *audioInput  = @"adc";
         
         if (dict) {
-            audioOutput = [dict objectForKey:@"Audio Output"];
-            audioInput  = [dict objectForKey:@"Audio Input"];
+            audioOutput = dict[@"Audio Output"];
+            audioInput  = dict[@"Audio Input"];
         }
         
         csound = [[CsoundObj alloc] init];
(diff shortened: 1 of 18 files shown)


aure:
Once again, awesome work.

aure:
Regarding AKPlots.h, I am thinking at some point before 2.1 when we're happy with the plots and other things that all of this just becomes part of AKFoundation.h.  I think I was reticent of including it because I knew it had some issues, but now that it's being bulletproofed by you, I think it's foundation material.  If you agree, I'll also let you decide when to move the AKPlots.h includes into AKFoundation.h - ie. if you're confident in them, I will be too.

megastep:
That's a great idea to "graduate" these classes to AKFoundation, however I don't think I'm confident enough yet to say that they are ready for this at the moment.

Mainly, several of these classes have not been implemented fully (some not at all) on OS X. Additionally, I've mainly tested the ones that are in the iOS version AudioKitDemo, and several of these classes are not even used in that app. The OS X side of things really needs to be tested more.

I think first we should have a test case (maybe just an extension of the AudioKitDemo) that features every single plot class so we can test them all on both iOS and OS X. I'm sure there are some issues left here and there... Then once we have made sure they are at least functional on those, then we can confidently add them to the foundation.

aure:
I think I'd like to test all the plots from within the AudioKitDemo project.  I'll also add them to the OSX side.  Probably be there by tonight.

PR #101, opened 2015-04-05
More plot cleanups

megastep:
* Remove unused code for rendering views in IB
* Fixed storyboard with non-existing properties
* New methods to get `NSMutableData` samples
* Made waveforms plots using `EZAudioPlot` thread-safe, moved the UIKit code out of Csound's way


aure:
Things are definitely coming together!

PR #102, opened 2015-04-06
Fixed the memory drainage

megastep:
Finally figured out that the memory getting out of control was mostly as a result of the Csound thread not having a proper autorelease pool setup. Added it to the loop inside `CsoundRender()` seems to do the trick and it's no longer going all over the place for me. I'm sure it should also show in your own code being run from within these callbacks.

Also some minor changes to the plots using EZAudioPlot, however I think there is some serious memory corruption going on in `updateBuffer:` from this class. Not sure what the purpose of `_setMaxLength` is there and it seems to be causing problems with triggering memory access out of bounds.

### DIFF
--- a/AudioKit/Platforms/OSX/CsoundObj.h
+++ b/AudioKit/Platforms/OSX/CsoundObj.h
@@ -129,8 +129,14 @@ typedef struct {
 - (MYFLT *)getOutputChannelPtr:(NSString *)channelName
                    channelType:(controlChannelType)channelType;
 
+// Read-only samples
 - (NSData *)getOutSamples;
 - (NSData *)getInSamples;
+
+// Writable alternatives
+- (NSMutableData *)getMutableInSamples;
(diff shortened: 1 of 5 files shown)


megastep:
I also just updated the OS X code of CsoundObj but I'm wondering why we even have that? Shouldn't it be mostly the same? Seems confusing to have a separate file for mostly platform-independent code. Also I noticed that you were using `@autoreleasepool` on that platform in some of the methods. Might want to double check with the iOS side of things.

aure:
I'm blown away.  All the memory problems seem to be gone, even stuff not having to do with plots.  This is huge!  So, I had help, but i mostly wrote the CsoundObj.m by myself.  It was obviously very close to the iOS version but there were a couple of key difference.  I'd certainly love to have one version or at least one base version that can get tweaked slightly for each OS if required.  

And yes, the rolling waveform still seems to be crashing immediately at least half the time.

Issue #93, opened 2015-03-29
Podfile dependency missing


niklassaers:
Hi, to include AudioKit from Cocoapods, I had to add the dependency "pod 'TPCircularBuffer'" in addition to not get linker errors for missing functions in EZAudio.o. Perhaps this dependency should be added to the Cocoapod podspec?

aure:
Hmmm, good catch, I think it has less to do with that podfile dependency and more to do with a .c file that didn't get included in AudioKit for TPCircularBuffer.c.  I have updated the Podspec in the develop branch and pushed the podspec up to cocoapds, hopefully this will fix it.

aure:
Actually it looks like cocoapods is going to require me to update the version number for this process.

niklassaers:
2.0.1? :-) As it stands, Audiokit doesn't compile with Cocoapods. The workaround I mentioned doesn't work for Swift projects that need to enable module support, so in that case it's won't compile at all.

I'd prefer the version bump to not compiling, but if you don't want to bump the version you could make the change in the podspec and then ask the user to point to your repo in the podfile in the README.md :-) For now I'll clone and send you a pull request for a version that includes the dependency, and can be included via a "pod 'AudioKit', :git => 'https://github.com/audiokit/AudioKit.git'"

aure:
Hi, yeah we were just really improving the plots so I wanted to get those done before release a 2.0.1.  Let's get this resolved one way or another today, thanks.

PR #106, opened 2015-04-07
Convert podspec to JSON and add TPCircularBuffer as dependency

niklassaers:
Re issue #93, I added the TPCircularBuffer as dependency so that it will compile

I converted the podspec to JSON because I'm not used to the Ruby syntax and got errors in the changes I attempted there. Although not required, I expect JSON to be more familiar to most developers than the Cocoapods Ruby DSL.

This patch still doesn't fully solve the case where I want to use "use_frameworks!" in my Podfile: when it builds, it complains about iostream not being found in CsoundFile.hpp

I've made a small project that reproduces this error here: https://github.com/niklassaers/AudioKitSwiftFrameworkError (the Podfile points to my repo, but I've included everything so that a download of the project should be enough to compile it and see the error reproduced)


niklassaers:
I've digged a bit more into that missing header, and it seems that while stdlib.h is as expected in /Applications/Xcode62.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/stdlib.h, iostream.h is in /Applications/Xcode62.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/c++/4.2.1/backward/iostream.h

I don't know what this 4.2.1/backward directory is. But I expect it could be added to the podspec file as an include dir? Do you have a suggestion?

niklassaers:
(I should add that the includes in question are iostream, string, vector, map, fstream and sstream. Long list! Has regular C++ been deprecated? ;-) )

aure:
I am a beginner with cocoapods but I will be attending the CocoaPod's hosted Test Jam on the 18th, maybe they can help us get this all resolved?

niklassaers:
Ok, cool that you'll be joining the test jam! :-) I posted a question on Stack Overflow: http://stackoverflow.com/questions/29501154/cocoapods-spec-that-uses-c-and-compiles-as-a-framework

PR #109, opened 2015-04-08
Restructured entire project around a new AudioKit static library

megastep:
I think this is pretty much functional right now, we now have a parent AudioKit project containing everything in the library, with a few caveats:

* It's not currently possible to mix Swift with Objective-C in libraries, so the Swift files still have to be manually added to Swift projects. We may want to consider merging the extensions file within `AudioKit.swift` to keep this simpler on users.
* Resource files are also separate from the code, so these also have to be manually referenced or copied from the projects, particularly `AudioKit.plist`.
* I've also moved all sound files in their own bundle so they can be easily referenced. I've updated the resource loading where appropriate throughout the examples and the library.
* Added a quick primer on how to get setup in `INSTALL.md`.

Hopefully the next step will be building a real framework, but this already significantly simplifies the Xcode setup for all examples.


aure:
I will need to rewrite some of the web pages before merging this changes.  In general the steps before were to just drag and drop the AudioKit folder and then remove the Platforms you don't need so I am not sure this really qualifies as "simpler" but if it can make the upgrade process easier, that would very good, which I think it should.  As for AudioKit.plist, can we avoid having to copy that file over by instead creating including a few definition in AKFoundation.h perhaps?  Is there a canonical way people do this?  Or perhaps one could argue that having your own personal plist file is better since it won't get overwritten when you upgrade?

megastep:
OK, yeah the "simpler" aspect here is to keep up with library changes. There's still the one-time setup of your project which is arguably a little bit more involved. But in the end it is cleaner. And it will probably get easier once we have an actual framework with the headers and library all in one place.

I agree we need to rework AudioKit.plist, I kept it the way it is for now. Ideally, I think it should be entirely optional, and the reasonable defaults (from the file distributed with AudioKit) should be in the code itself if the file can't be found. So people would only need to provide the file if they need to change any of the defaults.

What I would do is create a new class to abstract these preferences, say `AKPreferences`, which would have properties for each of the values in the plist, would be in charge of loading it when needed, and would be shared/references by all the other classes. This would probably be a good change to accompany this.

megastep:
Another thing you may want to consider for v2.1, since OS X 10.10.3 and iOS 8.3 just got released today. I'm sure the final version of Xcode 6.3 will be available at any moment now, with its massive improvements to Swift and new Objective-C qualifiers that can be applied to the API to make it easier to deal with in Swift. Basically you can specify which properties/arguments can't be nil, etc.

aure:
I have just made the "master" branch the default download for AudioKit so that we can work through this period of change without worrying too much about the web site and the code being out of sync.

aure:
And if it isn't obvious by the fact that I merged these changes, I should publicly state how cool this is!  Like 2.0, 2.1 will be a big update to AudioKit.

megastep:
BTW is it ok if I add stuff to the Trello boards?

aure:
Of course!

PR #110, opened 2015-04-09
Centralized all AudioKit.plist in AKSettings class, file no longer required

megastep:
This also prevents lots of duplications of the loading code, and makes sure the file is only parsed once per app start. This in turn cleans up a lot of the classes that used these settings.

Removed `AudioKit.plist` from all projects for which the defaults are fine (i.e. all).


aure:
Awesome, thanks!

PR #112, opened 2015-04-14
Updated CsoundLib OS X framework with custom build, lots of other improvements

megastep:
Quick summary:
* OS X `CsoundLib.framework` replaces the old `CsoundLib64.framework` and now uses `float`s instead of `double`s for samples, for parity with iOS.
* Framework is built from the latest `csound` along with special scripts at http://github.com/catloafsoft/csound; removed headers and sounds from the framework itself to avoid duplication.
* Consolidated `CsoundObj.m` for OS X and iOS into a single module, made it a lot cleaner and more thread-safe.
* Updated the various OS X demos to use the new framework.
* Various other cleanups.


megastep:
Might want to double-check my changes in `AKTable.m`, not sure if these are actually solving (or breaking) anything.

PR #117, opened 2015-04-14
Fix bug with microphone prompt being shown despite plist setting 'Enable...

npahucki:
[AKOrchestra addInstrument]  tries to disable the audio input, but is does not effectively avoid the microphone security prompt because [AKInstrument init] calls [AKOrchestra start] - BEFORE any instruments are ever added. Disabling the audio input in [AKOrchestra addInstrument] is too late to avoid prompting for the microphone, since the AKOrchestra has already been started when any subclass of AKInstrument is created. Thus, moving the code to enable or disable the input in  [AKOrchestra start]'s !isRunning condition solves the issue.

### DIFF
--- a/AudioKit/Core Classes/AKOrchestra.m
+++ b/AudioKit/Core Classes/AKOrchestra.m
@@ -59,6 +59,21 @@
 + (void)start
 {
     if (![[AKManager sharedManager] isRunning]) {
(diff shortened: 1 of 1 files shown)


aure:
Thanks!

npahucki:
Thank you for the great library!

PR #118, opened 2015-04-14
Added OS X versions of missing plots, AKLevelMeter

megastep:
This was actually rather quick to do just by comparing with the other plots which already had working OS X versions.

Also fixed some memory corruption that would likely lead to some crashes in `AKTablePlot`, and cleaned it up further.

I don't have a good test case at the moment for these but I'm pretty confident they should be pretty functional on OS X.


aure:
Yesss!

PR #119, opened 2015-04-15
AKManager and CsoundObj improvements

megastep:
* New utility class method on `AKManager` to simplify loading files from the AKSoundFiles bundle, this simplifies lots of duplicate code in instruments and the demo.
* The temporary `AudioKit.csd` file is now really in a temporary directory, with a temporary name including the process number, and is being removed upon application termination.
* We also wait for the Csound thread to end gracefully when the application terminates, to avoid crashes on OS X. This is now handled through a `NSThread` object instead of the `NSObject` method to launch a selector haphazardly in the background.

### DIFF
--- a/AudioKit/Core Classes/AKManager.h
+++ b/AudioKit/Core Classes/AKManager.h
@@ -19,7 +19,7 @@
  */
 @interface AKManager : NSObject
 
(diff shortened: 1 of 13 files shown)


aure:
The changes here are gorgeous, thank you!

PR #120, opened 2015-04-16
Fixes for AKTable and related classes

megastep:
OK, so `AKTable` was broken as I misunderstood the way it was working (the `malloc()` usage threw me off). Basically `AKTable` is now a simple proxy for the Csound tables, and we leave it to them to manage memory allocation. There's no real need to hold a pointer to the buffer in the class itself.

A couple of other mostly stylistic improvements to related code:
* Use `NSUInteger` instead of `int` for array table sizes, to be consistent with Cocoa `NSArray` semantics.
* Use modern Objective-C syntax to create `NSNumber`

### DIFF
--- a/AudioKit/Tables/AKTable.h
+++ b/AudioKit/Tables/AKTable.h
@@ -16,7 +16,7 @@
 @interface AKTable : NSObject
 
 /// The number of elements in the table.  Often required to be a multiple of 2.
-@property int size;
+@property NSUInteger size;
 
 /// A reference lookup number for the table.
 @property (readonly) int number;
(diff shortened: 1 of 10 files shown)


aure:
Will have to rework AKTablePlot as well I guess.

megastep:
Quite possibly, I hadn't thought about that and it does share some of the same problems. I can take a look at it later. Did this patch fix your earlier problems though? I hear the sine waves in the HelloWorld projects, but not too sure about the output from the AKTable tests.

aure:
Yeah, it definitely solves the problem.  I think AKTablePlot just has some other issue with things not necessarily existing yet.  We'll figure it out, onwards and upwards!

PR #123, opened 2015-04-19
[Travis] misc travis.yml tweaks

jhersh:
- Added some recommended ENV variables for `xcpretty`
- Ensured latest cocoapods and xcpretty are installed before testing
- Cleaned up build command
- Added a podspec linting step


aure:
Thanks so much!

PR #124, opened 2015-04-20
Building iOS Test Projects and Playground as part of TravisCI

aure:


aure:
OSX projects can't be built on Travis due to storyboards which require 10.10 Yosemite

PR #125, opened 2015-04-20
Limit the spinning through a time delay.

aure:
Maybe there's a better way, but this should avoid overspinning for the short amount of time these operations take.  Maybe it would be better through notifications or something, which is why I am creating a pull request instead of just pushing.


megastep:
Well maybe it would be good if `[AKOrchestra start]` itself was blocking, so that you could safely assume that it is fully started by the time the function call ends.

megastep:
Maybe we can just get rid of all these spin-loops; warts and all, `runOrchestra`, which is called from `[AKOrchestra start]`, already does that, so at best they seem already redundant.

PR #129, opened 2015-04-21
Table Tests moved to XCTest

aure:


AudioKit/Tables/Generators/AKExponentialTableGenerator.m
```diff
@@ -46,12 +46,12 @@ - (NSArray *)parametersWithSize:(NSUInteger)size
...
     }];
     
-    int maximumIndex = (int)[[points lastObject] objectAtIndex:0];
+    NSInteger maximumIndex = [[[points lastObject] objectAtIndex:0] integerValue];
     float scalingFactor = (float)size/(float)maximumIndex;
     
     NSMutableArray *scaledPoints = [[NSMutableArray alloc] init];
     for (NSArray *point in points) {
-        int index = (int)[point objectAtIndex:0];
+        NSInteger index = [[point objectAtIndex:0] integerValue];
```
megastep: You could even replace `objectAtIndex:` by a simple subscript, i.e. `point[0]` here.

megastep: Good call by the way, that old code must have been causing all sorts of bad shit, casting a pointer to an `int`. Hope there's not more of this in other places...

aure: Thanks to tests actually, it did not pass at first, making this bug pretty clear.

PR #130, opened 2015-04-21
[Travis] more test tweaks

jhersh:
So many great tests in this project now! Great to see :+1:

I made some minor tweaks:

- Cache cocoapods
- Use Travis' `pod install`
- `CGFloat` consts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,21 +3,22 @@ language: objective-c
 env:
 - LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8
 
-before_install: gem install xcpretty cocoapods -N
+before_install: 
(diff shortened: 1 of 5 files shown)


jhersh:
I thought there was a problem with Travis not installing pods, but [turns out](http://docs.travis-ci.com/user/languages/objective-c/) :+1:

> Also, `pod install` is not run if the Pods directory is vendored and there have been no changes to the Podfile.lock file.

aure:
Cool man, I have ported all of non-sound tests to XCTests, just now looking at clever ways to test that the right sounds were made.  Thinking of creating wavs and hashing them to compare against known out put, or even just taking an FFT of the output and comparing some markers.

PR #131, opened 2015-04-21
Testing some sound generation using Tracked Frequency 

aure:
And various cleanups.


AudioKit/Core Classes/AKManager.h
```diff
@@ -29,6 +29,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// Determines whether or not to log
 @property BOOL isLogging;
 
+/// Place to put lines of testing logs
+@property NSMutableArray *testLog;
+
```
megastep: I'm a bit concerned about keep log lines in memory - there's a real risk of a runaway log just crashing things if it goes out of control.

Also if this is only for unit tests, maybe we can use `#ifdef TRAVIS_CI` to only include this piece then.

aure: This is only added to in tests.  Not only for TravisCI, but locally.  And also, I will be replacing this soon.

megastep: OK, then I suggest we bracket this property and the corresponding code in the message delegate so it doesn't even get included in non-test builds (save a bit of memory). That's what that TRAVIS_CI macro is for. :+1:

aure: Well, like I said, this will be changing and going away, so I hate to put more effort into it, but if you wanna, you can.

PR #132, opened 2015-04-23
Dynamic Frameworks for iOS

megastep:
Well, that's the big one. There may be a few kinks to work out here and there but this is pretty much working. I was able to submit a binary to Apple and have it pass their validation tests, so we should be pretty good.

A few notes on this:
* Even the OS X `CsoundLib.framework` has been updated, and the library name is set to be simpler, so as to make the run script shorter.
* Had to add some code to the run script for iOS to strip x86 slices from the embedded frameworks when building for devices (and thus for submission to Apple), as they really don't like to see any kind of Intel code in there.
* Split up the `INSTALL.md` file into iOS and OS X versions, as it was becoming too confusing.
* Updated all examples to use dynamic frameworks, except for `TableDemo`.


aure:
Not sure why table tests are failing but I can figure that out.  What a huge pull request!

megastep:
Yeah I think something might have gone wrong when I merged your latest changes into this branch, should probably double-check the OS X test project.

PR #133, opened 2015-04-23
Rolling plots improvements

megastep:
I'm trying to improve on the CPU usage for the rolling plots and made a number of improvements, though it's not perfect yet:

* Fixed a bug with closing the paths in EZPlot, and made it a little more memory efficient, and no longer needing write access to the data buffers being passed to it.
* Trigger all drawing directly from the Csound callbacks, which actually makes for a faster scrolling plot and saves having to manage locks.
* Allow to specify an inspectable property to set the refresh rate for these plots, in an effort to gain control over how often the `drawRect:` method gets called.
* Also another little tweak to the iOS run script.

I think it's obvious that the EZPlot drawing code is woefully inefficient as is - it tries to draw a lot more line segments that are really needed (one for each sample in the buffer, two if you include the mirrors). So that takes a lot of CPU to do very little visible to the user. I'll try to improve on this tomorrow...

### DIFF
--- a/AudioKit/Utilities/Plots/AKAudioInputRollingWaveformPlot.m
+++ b/AudioKit/Utilities/Plots/AKAudioInputRollingWaveformPlot.m
@@ -14,10 +14,10 @@
 @interface AKAudioInputRollingWaveformPlot() <CsoundBinding>
 {
     // AudioKit sound data
-    NSMutableData *inSamples;
-    UInt32 sampleSize;
+    UInt32 _sampleSize;
+    NSDate *_lastUpdate;
     
-    CsoundObj *cs;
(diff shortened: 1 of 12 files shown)


aure:
The CPU usage is still high with rolling histrory of 2048 hardcoded maybe we just need to expose this length to be easily changed in IBDesginable, etc.

megastep:
Yeah, we could do that but I'm gonna take a shot at improving the way the drawing is done here. That's the crux of the issue - for every redraw, it draws thousands of line segments, many of which will never be visible. Maybe we can even have some kind of bitmap cache so that only the new samples are drawn.

PR #134, opened 2015-04-28
Various plot improvements, Mac fixes

megastep:
Another big one, highlights following:

* Rewrote a good chunk of `EZAudioPlot` related code to make it render as much as possible in an offscreen `CGLayer`, and only for new samples as they come along. Also uses a bit of the `Accelerate` framework. Arguably it's still not ideal, but it should be better on the CPU than before.
* Consolidated several of the plot classes with duplicated code into parent classes.
* All plots should now be in `AKFoundation.h`
* Added rolling waveform input plot to `AudioKitDemo` project on OS X (not in the Swift version though)
* Fixes for OS X, particularly the orientation of some of these plots as well as a huge memory link in `CsoundObj` (missing an autorelease pool, same thing we had on iOS). I haven't checked the table view plot as the example for it is now gone?
* Slight changes to the run script for iOS to avoid getting errors that should be ignored.

Next I'm gonna make a build of Csound 6.05 for all platforms and will update the binaries straight in the repository when they're ready.


aure:
Bam!

Issue #138, opened 2015-04-29
2.1 podspec


niklassaers:
Hi Aurelius,
has the updated podspec been sent to cocoapods yet? When I search I find version 2.0.1, but not 2.1.

So to test it out, I pointed my podfile directly:
     pod 'AudioKit', :git => 'https://github.com/AudioKit/AudioKit.git'

Having done this, I get "csound.h" not found on line 32 in CsoundObj.h

Any idea what this is about?

Cheers

    Nik

megastep:
Since the library's structure has changed quite a bit since 2.0, especially when it comes to the static library and its assortment of dependencies, it's not a surprise that this would be currently broken. We were actually hoping that CocoaPods users might help us fix this. ;)

megastep:
Actually I see quite a few references in the pod spec to some old files. Let me try to commit a fix for that and see if it helps...

megastep:
Can you try the pod spec on the develop branch and see if it's any better?

niklassaers:
Hi Peter, thanks for the quick reply. (sorry for my slow reply, I'll monitor the thread more closely now :-) )

The error is the same still.

I'm poking a bit at it and have a setup for it, so if you want to do some screen sharing or quick exchange of ideas, you can catch me on Skype as niklassaers

niklassaers:
Hmm, perhaps I had something cached, because now it asks for CsoundObj.h instead. :-) (from AKManager.h)

niklassaers:
Fixed that one, will send you a diff soon

niklassaers:
Pull request sent :-)

niklassaers:
PS, thank you soo much for your quick reply and update to the podspec :-) (and for all the good work you put into AudioKit)

aure:
For me, Xcode can't find AKFoundation.h after I've pod init'ed and opened up the workspace.

PR #139, opened 2015-04-29
End https://github.com/audiokit/AudioKit/issues/138

niklassaers:
Thanks for the updates, Peter. I had to add this to make it build, but then it also builds very nicely :-)


aure:
So, just to follow up, with this podspec you're able to run pod init on a project and things just work?  I just tried and it didn't seem to set things up correctly.  I want to know so that I can push this release up to cocoapods.

PR #145, opened 2015-05-09
Allow audio input to be turned on/off at any moment, expose AKSettings

megastep:
This one is focused on being able to turn on audio input at will during the lifetime of the app, i.e. not just when AudioKit is first initialized.

If only playback is enabled, then we use the more permissive `AVAudioSessionCategoryAmbient` category on iOS, to allow things such as the silent switch to be honored. The old behavior can be restored using the new setting "Playback While Muted", for apps that may require that.

As a side effect, I have changed the input setting to be OFF by default (as opposed to ON before), in the plist and the AKSettings defaults. To make this easier on the user, `AKSettings` is now part of AKFoundation and several of its properties are writable, including `audioInputEnabled`. This one must now be set explicitly in the programs (see what I did in `AudioKitDemo` where this is called only when first going to the analysis tab), or in the plist.

Note that this mostly affects iOS, even though the API is available on Mac too, where input is pretty much always available. The main advantage is avoiding the dreaded permissions dialog on iOS if not necessary.

Other changes:
* More cleanups in `CsoundObj.m`, including getting rid of duplicate or unused properties and methods
* The singleton for `AKSettings` is now accessed through a `shared` method (instead of the old `settings`) as it caused issues with Swift thinking it was actually a convenience constructor.
* Enabling/disabling of audio input is now mostly done through `AKSettings`, so `AKOrchestra` doesn't need to worry about that anymore.

I've updated all flavors of `AudioKitDemo` but any other examples you may have in the Examples repo might need to have audio input explicitly turned on if they need it.

### DIFF
--- a/AudioKit/Core Classes/AKFoundation.h
+++ b/AudioKit/Core Classes/AKFoundation.h
@@ -15,6 +15,7 @@
 #import "AKCompatibility.h"
 
 # pragma mark - Core Classes
+#import "AKSettings.h"
 #import "AKManager.h"
 #import "AKOrchestra.h"
 
--- a/AudioKit/Core Classes/AKManager.m
+++ b/AudioKit/Core Classes/AKManager.m
@@ -239,12 +239,12 @@ static AKManager *_sharedManager = nil;
 
 /// Enable Audio Input
 - (void)enableAudioInput {
-    [self.engine setUseAudioInput:YES];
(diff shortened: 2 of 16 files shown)


aure:
Excellent update!

narner:
This is super useful, Stephane!

PR #146, opened 2015-05-10
Made a Resources folder 

aure:
AudioKit.plist didn't belong in Core Classes and the SoundFile bundle didn't look good in the top directory.


megastep:
Cool, I'm sure we'll revisit some of that as we put the framework together.

PR #155, opened 2015-05-21
Adding AK prefix to a bunch of classes

megastep:
As we start building the framework, a good practice is to make sure that all public classes in the library have unique names, and the best way to do this is to make sure they all use the AK prefix. The main holdouts so far are classes in the Utilities/Instruments group.

This patch changes those I could without introducing clashes with already existing names. This is actually a problem that we need to address, but it might be better to hash it out between @aure  and @narner to figure out the best way to resolve this.

Basically, you'll see that for instance we have a `Mandolin` class in Utilities, but we can't rename it to `AKMandolin` because there's already such a class under Operations. This applies to a bunch of those instruments (pretty much all remaining classes without the prefix).

So we need to agree on how best to handle this if we're going to keep these non-prefixed classes as public and part of the library. Either chuck them out if they're really just examples, or find some better way to rename them that still includes the prefix. One thing is sure is that we can't have a public `Mandolin` class as part of the framework, since it's very likely to conflict with similar classes in apps.

### DIFF
--- a/AudioKit.xcodeproj/project.pbxproj
+++ b/AudioKit.xcodeproj/project.pbxproj
@@ -14,10 +14,10 @@
 		C4C8125B1AE5CA03003A8B73 /* Sekere.h in Headers */ = {isa = PBXBuildFile; fileRef = C4C812591AE5CA03003A8B73 /* Sekere.h */; };
 		C4C8125C1AE5CA03003A8B73 /* Sekere.m in Sources */ = {isa = PBXBuildFile; fileRef = C4C8125A1AE5CA03003A8B73 /* Sekere.m */; };
 		C4C8125D1AE5CA03003A8B73 /* Sekere.m in Sources */ = {isa = PBXBuildFile; fileRef = C4C8125A1AE5CA03003A8B73 /* Sekere.m */; };
-		C4FAE5F11AED6B2600FD71E2 /* AudioFilePlayer.h in Headers */ = {isa = PBXBuildFile; fileRef = C4FAE5EF1AED6B2600FD71E2 /* AudioFilePlayer.h */; };
-		C4FAE5F21AED6B2600FD71E2 /* AudioFilePlayer.m in Sources */ = {isa = PBXBuildFile; fileRef = C4FAE5F01AED6B2600FD71E2 /* AudioFilePlayer.m */; };
-		C4FAE5F31AED6B2600FD71E2 /* AudioFilePlayer.m in Sources */ = {isa = PBXBuildFile; fileRef = C4FAE5F01AED6B2600FD71E2 /* AudioFilePlayer.m */; };
-		C4FAE5F41AED6B2600FD71E2 /* AudioFilePlayer.m in Sources */ = {isa = PBXBuildFile; fileRef = C4FAE5F01AED6B2600FD71E2 /* AudioFilePlayer.m */; };
+		C4FAE5F11AED6B2600FD71E2 /* AKAudioFilePlayer.h in Headers */ = {isa = PBXBuildFile; fileRef = C4FAE5EF1AED6B2600FD71E2 /* AKAudioFilePlayer.h */; };
+		C4FAE5F21AED6B2600FD71E2 /* AKAudioFilePlayer.m in Sources */ = {isa = PBXBuildFile; fileRef = C4FAE5F01AED6B2600FD71E2 /* AKAudioFilePlayer.m */; };
+		C4FAE5F31AED6B2600FD71E2 /* AKAudioFilePlayer.m in Sources */ = {isa = PBXBuildFile; fileRef = C4FAE5F01AED6B2600FD71E2 /* AKAudioFilePlayer.m */; };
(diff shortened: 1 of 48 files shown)


megastep:
I'd personally be in favor of exploring how these classes/files could be merged to avoid spreading functionality around, or otherwise use a longer prefix, like `AKInstrument...` but that would be rather cumbersome.

narner:
@megastep I think this makes a lot of sense, but we may have to consider the fact that Apple reserves two letter prefixes for framework classes. Maybe something like `AKI` as a prefix for `AKInstrument` classes, etc.

aure:
@narner no way are we having different prefixes for different parts of AudioKit.  
@megastep This actually got to this place because instruments were only mean to be examples for people, not included in AudioKit, but then it became clear that they would be so helpful that we included them in 2.1.

So, we have tons of operations and only a handful of instruments.  Most of the instruments will not overlap in names so I didn't feel the need to standardize the name to AKXXXXInstrument.  But in the case of physical models, this becomes an issue.  So, I guess the instruments have to be renamed and I vote for renaming only conflicting instruments by adding Instrument to the *end* of the name  So Mandolin becomes AKMandolinInstrument.  Sure, both Mandolin variations will come up in code completion but only one will suffice in any given context.

aure:
I am leaving this open if you want to finish the process in one request, or else I could merge and rename the rest myself, up to you.

megastep:
Well I am looking into this, and for the instruments, I see that most of them have a separate `...Note` class defined in the same file. I wonder if there's a real justification for exposing these since they seem to only be used within the module for the instrument itself, and there doesn't seem to be a way to use these in the APIs otherwise?

For instance look at the current `Mandolin` and `MandolinNote` classes. It seems like the note class is only used once in the initializer for `Mandolin`. Would it make sense to consolidate the code to the Mandolin class if the note is not really usable by end users? Or am I missing something?

aure:
Having the notion of a note that is specific to the instrument is useful because then you can add notes to phrases and play them back.  It's a pattern that is quite prevalent in the examples and stuff, so I think we shouldn't mess with it for now.  But, I am open to an improvement to the architecture if it makes things easier for users.

megastep:
Sure - but how would a user go about using, say the `MandolinNote` class? I see that it's a subclass of `AKNote` but where are the extra attributes it includes used outside of the `Mandolin` class? I see that it is used to encapsulate some of the attributes created when the `Mandolin` is created, but it doesn't seem to be used anywhere else...

I guess this would make more sense if this was indeed intended to be used as example code that the user can then mess with and extend, but as is to me it looks a little bit weird in the library itself.

megastep:
I guess this would make more sense to me if the `Mandolin` class had a `MandolinNote` property exposed that the user could interact with, something like this:

`@property (readonly) MandolineNote *note;`

Would that make sense?

aure:
Well, no because an instrument can be simultaneously playing any number of notes.

megastep:
OK, then maybe my lack of knowledge of the AudioKit API is shining here, so how would we typically go about making the current `Mandolin` instrument play several `MandolinNote`s ?

aure:
Total pseudocode, but something like this.  mandolin can have instrument properties, and mandolinNotes have different properties, frequency being a common one.

AKMandolinNote *note1 = [[AKMandolinNote alloc] init];
note1.frequency = akp(440);
note1.duration.value = 1.0;
AKMandolinNote *note2 = [[AKMandolinNote alloc] init];
note1.frequency = akp(660);
note1.duration.value = 1.0;
AKMandolinInstrument *mandolin = [AKMandolin presetSomeMandolin];
[AKOrchestra addIntrument:mandolin];
[mandolin playNote:note1];
[mandolin playNote:note2 afterDuration: 1.0];

megastep:
OK, so I guess where I'm getting confused is how this differs from the current `Mandolin` class (before we rename it, not the separate `AKMandolin`). From what I see, this `Mandolin` class (in Utilities/Emulation) creates its own `AKMandolin`, hooks up the properties from the `MandolinNote` it also created there, and just goes from there. Neither note nor instrument are exposed there.

So if I created a `Mandolin` object (not `AKMandolin`) I wouldn't have access to the actual `AKMandolin` it contained to send `playNote:` to. Since it's actually an `AKParameter` - not an `AKInstrument`, I don't think this is possible?

Sorry if I'm being a bit obtuse here, I'm trying to better understand how this all fits together and how we can maybe make it simpler. :)

megastep:
Actually, scratch that last comment, I must have been looking at a different class, `Mandolin` really is an `AKInstrument`...

megastep:
Alright, I think I understand this better now. This just highlights how badly we need to clarify these class names. So there's an important difference between the model (`AKMandolin`) and the instrument (`Mandolin`) that wraps it. So then it will be better if we rename that class to `AKMandolinInstrument` so it will make it more obvious how this fits together.

megastep:
On a related note, while we're talking about renaming classes, how about using a `...Model` suffix for model classes? So we'd have `AKMandolinModel` being used by an `AKMandolinInstrument`. Verbose but probably less confusing.

aure:
The only reason I don't want to add the Model to the name is that Alexander Hodges has already been using the physical models to great effect and I wouldn't want to change the names on him.

megastep:
OK, fair enough. Just appending Instrument to the other classes is already a significant step up in clarity. I just committed to rename the rest of the classes, maybe you can merge after CI finishes checking it.

PR #158, opened 2015-05-24
New testing system

aure:
I've only ported 10 of the tests from the previous Swift based command line system, but the rest will just be a matter of doing a few per day for the next couple of weeks.  The set up seems pretty solid, might be a few too many ifdef TRAVIS_CI's floating around, but we can clean this up over time too.


AudioKit/Core Classes/AKManager.m
```diff
@@ -116,15 +112,20 @@ - (instancetype)init {
...
                     @"-o %@           ; Write sound to the host audio output\n"
                     "--expression-opt ; Enable expression optimizations\n"
                     "-m0              ; Print raw amplitudes\n"
-                    "-i %@            ; Request sound from the host audio input device",
-                    AKSettings.shared.audioOutput, AKSettings.shared.audioInput];
+                    "%@\n",
+                    AKSettings.shared.audioOutput, inputOption];
         
-        _csdFile = [NSString stringWithFormat:@"%@/AudioKit-%@.csd", NSTemporaryDirectory(), @(getpid())];
+        _csdFile = [NSString stringWithFormat:@"%@/AudioKit.csd", NSTemporaryDirectory()];
```
megastep: Why did you remove the `getpid()` here? That actually made the temporary file more secure by making its filename less predictable...

AudioKit/Platforms/Common/CsoundObj.m
```diff
@@ -89,41 +99,61 @@ - (instancetype)init
...
+#else
     if (_cs) {
         csoundCompileOrc(_cs, (const char *)[orchestraString cStringUsingEncoding:NSASCIIStringEncoding]);
     }
+#endif
 }
 
 - (void)stop {
+#ifdef TRAVIS_CI
+#else
```
megastep: I'm not completely convinced you do need to comment out all this code for testing. Even if you're rendering faster than realtime, you should still want to finish cleanly.

AudioKit/Core Classes/AKManager.m
aure: AudioKit produces the same file over and over again, it's never different for different builds, so I got sick of their being a lot of files in that directory.  I have no strong opinion, so if you want it back, I can put it back.

megastep: Well the right way to do this, considering these are temporary files, would be to make sure that they are removed upon program termination. :)

Arguably this is not too much of a problem on iOS where this is all within the app sandbox, but could be a security issue on the Mac.

--
Stéphane Peter
megastep@megastep.org

AudioKit/Platforms/Common/CsoundObj.m
```diff
@@ -142,11 +172,15 @@ - (void)record:(NSString *)csdFilePath toURL:(NSURL *)outputURL
 
 - (void)record:(NSString *)csdFilePath toFile:(NSString *)outputFile
 {
+#ifdef TRAVIS_CI
+    [self performSelectorOnMainThread:@selector(runCsoundToDisk:) withObject:@[csdFilePath, outputFile] waitUntilDone:YES];
```
megastep: Again here, why go back to the old way of performing the selector on a thread instead of what we have with `NSThread`? It's essentially the same but gives us some additional control on the thread to terminate it more cleanly.

aure: The thread version crashes in tests.

megastep: Any idea why? That's pretty odd.

megastep:
Generally, my main objections are with the changes to CsoundObj which tend to use a bit too much `#ifdef TRAVIS_CI` I think - ideally you want to keep these to a minimum so that such a core class isn't too different from what ends up in the main library.

I suppose there are things I can make better in this class, most notably making sure the temporary files get deleted (there are notifications we can intercept for that) and double-checking that threading code. Maybe you exposed some edge case there.

AudioKit/Platforms/Common/CsoundObj.m
aure: No idea, and its strange, it will work some of the time, maybe 10%?  And the failures are a variety of unexplained things.  With performSelector it works every time.

megastep: Weird - there's probably some kind of race condition somewhere. We should get to the bottom of it...

megastep:
I think I'll just merge this now, will make it easier to look at these issues closer myself.

PR #160, opened 2015-05-25
Improvements for unit testing

megastep:
Some of these tests do still fail for me, however at least they don't seem to be crashing.

Mostly I wanted to restore `CsoundObj.m` to use a little ifdef's as possible. I noticed you were not using setup/teardown methods in the tests, and it made more sense to me to do the kind of thing you were doing within those, so I exposed the necessary code through `CsoundObj` and `AKManager`.

I hope I didn't break anything as I'm not too sure of the purpose of some of your new code, but I think this should be an improvement.

As a side effect, this also simplifies a little bit of the boiler plate for individual tests.


megastep:
Hum, lots of odd failures happening in Travis that didn't happen on my system...

narner:
@megastep Try triggering a rebuild on Travis; there have been times where I've seen it fail on Travis but passed on my machine, and a rebuild showed nothing was wrong.

aure:
From looking at the Travis log I see these are not the trivial Travis hanging errors, but instead the files not really being created correctly for the MD5 hash to run on them.  None of the tests are broken on my own system or on Travis running the current branch so this is fixable

megastep:
Actually, I was wondering if the `renderToFile:duration:` method is supposed to be blocking? I think it might have something to do with this. Your performSelector call would wait for completion, whereas my NSThread code doesn't as is.

This could explain how this is getting tested before the file gets a chance to get created properly.

megastep:
These last changes seem to help for me, I no longer get crashes in tests on either iOS or OS X, although some of those tests do fail for me (but that seems to be another issue).

I also fixed the way the OS X test project was setup, apparently there's no need to duplicate the linking configuration for test target.

I really wonder if we should change the `record:..` APIs to either say that they are blocking, and maybe give an option for them to spawn their own threads? This really depends on what the intended usage for these is, you probably don't want to run Csound on the main thread if you have a UI.

megastep:
Woohoo, Travis passed!

aure:
Fantastic!

PR #163, opened 2015-05-26
Added AKEqualizer presets

narner:


aure:
There are a few things I'd like to change on these before moving forward.  Will contact you on Screenhero later to resolve.

PR #179, opened 2015-06-15
Changed CsoundObj license to MIT

megastep:
I noticed that the CsoundObj files still mentioned that they were licensed under the LGPL. To avoid contamination for the rest of the library, I updated the licensing bit to reflect the MIT license. I am assuming this is kosher with everybody involved.

Also I tried to compile the library with Xcode 7 and it suggested a few configuration changes, shouldn't impact anything for now.

### DIFF
--- a/AudioKit.xcodeproj/project.pbxproj
+++ b/AudioKit.xcodeproj/project.pbxproj
@@ -1912,7 +1912,7 @@
 			isa = PBXProject;
 			attributes = {
 				CLASSPREFIX = AK;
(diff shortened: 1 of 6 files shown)


megastep:
Getting some very weird Travis failures here - passes just fine on my branch, with the same code. Seems to be something on their servers.

aure:
You don't think it has to do with Xcode7 testability?

megastep:
That crossed my mind, but then it would fail on my branch too, and it doesn't. The log shows something failing in the simulator.

PR #185, opened 2015-06-19
Travis CI beta support for Xcode 6.3 / OS X 10.10

megastep:
So it turns out that there is currently beta support for the newer platforms on Travis, which in turns allows us to run more OS X tests. Seems to be working better than the regular VMs right now, so this turns it on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: objective-c
+osx_image: beta-xcode6.3
 
 env:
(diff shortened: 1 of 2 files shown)


aure:
This is great, how did you find out about it?

megastep:
I went to their website to see if they add anything about it since it was a common problem, then they had a blog about this new beta. Nice to see they're trying to bring this up to date.

aure:
Good sleuthing!

PR #188, opened 2015-06-30
Sound font progress for Stephane to Review

aure:
This pull request is not intended to necessarily be accepted just yet, but to be shared for progress review and discussion, for instance:

* The message delegate pattern is something I think we should get rid of, meaning changing AKManager, any objections?  I'll do that next
* I will convert the instruments and preset array from arrays of strings to arrays of objects
* I didn't quite grok instruments and presets before, but now it seems like I will need two players, one for instruments and another for presets, no biggie.


megastep:
Weird, we seem to have these Travis failures occurring again... Anyway:

* I don't particularly object to relying on notifications, but if that's something we're gonna do extensively we should probably have some utility class methods on `AKManager` to make that easier and abstract away some of the cruft. Might even be able to have a utility method there too to help getting the pieces of the messages back.
* So are instruments and presets completely independent of each others then? I was under the impression that presets were attached to instruments. but I could be wrong. Maybe we can still stick with one player class that can handle both?

aure:
I'm not worrying about Travis just yet, I didn't test on my system.   Instruments and presets are very distinct types of things.  A preset can contain multiple instruments for instance.

megastep:
Interesting, I thought it was the other way around. Anyway, I'm gonna play around with Soundfonts as I need to create my own here. That will give me a better idea of how all of this is structured.

aure:
The Travis error is weird, referring to "07AcousticGuitar" when I don't think I have that anywhere in my code anymore.  Changing a lot of other stuff so no worries, but that is definitely strange.

aure:
This is useful: http://www.synthfont.com/Viena_HELP.pdf

megastep:
Very interesting, then it would make sense for the player to only use presets since they are the visible part of this? The instrument info would be mostly informational.

Also it might make sense to have a mono player too. Might be less CPU intensive if the font only has mono samples in it.

aure:
It does seem like instruments and presets are necessary to have different players for at the operation level.  Instruments have the drum kits for example.  In any event, both are created now.  To make it even easier to use, I'll make a Utility AKInstrument and an example app.  This is probably okay for merging as is though.

megastep:
Looks good, by the way I was wondering about these new notifications. Is it the only place in AudioKit where we need to parse the output from Csound like this, and is this why you need to have a way to handle the messages? I'll probably make some changes to CsoundObj to make this easier.

aure:
Yes, this is the only place that we're parsing the output.  I don't know of anything else that uses this type of functionality, but it might come up.  I could envision bolstering AKManager to know more about the state of affairs inside the engine, for instance if there is an error, we should probably log regardless of how isLogging is set and give the user some good information back.

megastep:
Yeah, I am concerned about what may happen when you have multiple sound fonts for instance (all objects will get the global notifications), or since we now handle all messages. We need to clean this up significantly I think to avoid corruption.

It would probably be helpful to know more about the codes that Csound uses to send alongside these messages, we might be able to use them to filter.

aure:
The instrument and preset list is preceded by a note saying what file it is from a la:

Instrument list of "/Users/aure/Library/Developer/CoreSimulator/Devices/5C73E22D-A696-4AFF-A2CA-CD1A4B944BBE/data/Containers/Bundle/Application/CC4D8D74-B041-47A2-9D99-800F2174D991/AudioKitPlayground.app/AKSoundFiles.bundle/Sounds/GeneralMidi.sf2"

So I should just check this.

PR #192, opened 2015-07-02
More improvements for AKSoundFont

megastep:
I made another small API change to Csound to mark the end of the lists of presets/instruments, and this takes advantage of that.

Also added a few utility methods on `AKSoundFont` to locate presets and instruments easily.


aure:
Seems to be a minor warning on one of files from Travis:

⚠️  /Users/travis/build/audiokit/AudioKit/AudioKit/Sound Fonts/AKSoundFontInstrumentPlayer.m:133:42: values of type 'NSUInteger' should not be used as format arguments; add an explicit cast to 'unsigned long' instead [-Wformat]
    [inputsString appendFormat:@"%lu, ", _soundFontInstrument.number];

but no biggie, will merge.

megastep:
Funny, I didn't touch that piece of code. That's probably your doing. 😛

PR #193, opened 2015-07-04
Run Phase script improvements

megastep:
* Moved the contents of the scripts in their own `run-script.sh` for iOS and OS X.
* Fixed the iOS script to handle lipo better, should fix the errors showing up when building on device.
* Updated all projects to use these new scripts directly.
* Updated the documentation to reflect these changes.


aure:
Very nice but now I have to update screencasts! :)

PR #194, opened 2015-07-05
Moved AKMidi to a notification center based model

aure:
I've tested this to work on both OSX and the iOS simulator.  For the simulator, you can hook up any device connected to your Mac via MIDI network sessions.  Other than some clean up, I think the only thing that should be added is a setting for whether MIDI is default on or not.  I like it default on, but I suppose if your app doesn't need its a little bit of a waste.  Anyway, Stephane, now that I've seen your midi set up I think you'll like this for your sound font tests as well.


megastep:
Looks good, I'll take a closer look later.

PR #202, opened 2015-07-11
Added AKClip class (csound opcode 'clip')

dclelland:
Hi there, I needed a simple limiter, so I built a limiter based on the csound 'clip' opcode.

I noticed in most of your operation classes there was mention of them being auto-generated - so, wasn't sure how this might fit into your workflow, but I thought I might as well put in a PR anyway :)


dclelland:
Cheers buddy! :)

aure:
Fixing one thing, but looks good.

aure:
We auto-generate because it makes things easier, but you followed the mold so its good.  Will eventually add a test for this operation too.

PR #205, opened 2015-07-14
Ezaudio update

megastep:
Basically imported `EZAudio` 1.0 (at least the bits we use) and updated the rolling waveform plots to use it instead, then merged back some of the changes I had made.

Also updated the examples to use the changed property for the plot color.

Likely more to come as 1.1 was just released we may want to use their FFT code, but this works for now.


aure:
Fantastic!  One thing is that rolling input and output is no longer working in playgrounds, perhaps its a plot vs. plotColor thing.  I'll check it out.

megastep:
Oh yes, that's very likely. I didn't think these plots were used in playgrounds so I didn't check.

PR #212, opened 2015-08-04
MIDI / SoundFont tab in iOS AudioKitDemo

megastep:
This is mostly a port of the current OS X demo app, to add the MIDI and sound font for testing on that platform. It will need more polish, including a tab bar icon.

The good news is that it is working for me. I do not experience crashes when playing the keyboard on iOS either with this code. There seems to be a little bit of occasional audio distortion, but that's another issue.

So whatever is triggering the "many notes" crashes must be somewhere else, maybe in the way that the MIDI playground currently sends the events?


aure:
Even the many note bug persists, it is good to know it is not due to the midi changes.

PR #214, opened 2015-08-09
New Csound binaries with bug fix, interface for Csound debugger

megastep:
The main thing is updated binaries that revert the commit that causes all the problems in Csound. This should allow us to move on for the time being until they find a better way to handle this.

Also, started to implement some of the `csdebug.h` API inside of CsoundObj, only available for debug builds (non-release). This should allow to set breakpoints while developing code interfacing with Csound and do some introspection. I haven't tried to use it yet but I believe it could be useful.


aure:
Hallelujah!

PR #224, opened 2015-08-30
Added default peak scaling to AKAntiresonator

dclelland:
This also avoids this csound bug:
http://csound.1045644.n5.nabble.com/reson-areson-td5736604.html


dclelland:
For the record, this is what I was getting (orange AKResonator, yellow AKAntiresonator):

screen shot 2015-08-30 at 14 43 33

`areson` should be a notch filter, but it is identical to AKResonator when `iscl` is 0.

PS I'll update the tests too...

aure:
Nice!

PR #230, opened 2015-09-11
Changes for Xcode 7

megastep:
Getting ready for the final Xcode 7 release next week, with support for Bitcode, Swift 2 and various Objective-C enhancements. This is by no means comprehensive, but it should make AudioKit fit better in. I'll probably finally get to making a proper framework after this.

I added generics where I felt confident enough about the types being handled, and mostly focused on the public APIs. It didn't always seemed worth the effort to specify these in internal code. This is mostly to improve usage of the API with Swift so it should be good enough for now.

Also now using the new xcode7 Travis VM, which works great now and is available to all.


aure:
You are a rock star!

PR #233, opened 2015-09-19
Created AKMaskedFFT class

dclelland:


dclelland:
I feel like I oughta start writing some tests for all these FFT classes eventually...

aure:
Tests would be nice.  Not too hard to write, just a little boring. :)

PR #239, opened 2015-10-08
Maximum note allocation

dclelland:
This is kind of experimental - added a maximumNoteAllocation attribute to AKInstrument.

I was running into a problem where adding too many notes to an instrument caused the CPU to hit 100% and Csound to stop making any noise. This allows you to set a hard limit on number of notes in order to avoid this scenario.

Not as good as proper polyphony (i.e. with a FIFO stack, where old notes get silenced first), but perhaps better than nothing.

### DIFF
--- a/AudioKit/Core Classes/AKOrchestra.m
+++ b/AudioKit/Core Classes/AKOrchestra.m
@@ -116,6 +116,14 @@
         }
     }
     
+    if (instrument.maximumNoteAllocation) {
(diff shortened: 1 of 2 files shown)


dclelland:
Am interested in setting a CPU limit as well - see http://www.csounds.com/manual/html/cpuprc.html - would perhaps prefer some feedback before going ahead, though.

aure:
Interesting.  I'll accept this change.

dclelland:
Did you mean to merge or close it?

aure:
Forgot to confirm the merge.

PR #240, opened 2015-10-13
CPU improvements, Git LFS support

megastep:
Drastically reduced CPU usage on iOS in AU thread, mostly by avoiding unnecessary critical sections and increasing sleep values.


megastep:
Also adding `.gitattributes` files to set up Git LFS for all the binary files. This way we can do a proper release with Csound 6.06.

PR #244, opened 2015-10-22
tvOS support

megastep:
This is basic support for tvOS 9.0 in AudioKit, along with an example HelloWorld project, based on the iOS one, that works on both the simulator and my developer device with Xcode 7.1.

As I expected, the differences with iOS, while notable, are relatively minor as far as we are concerned. The few changes I had to make to Csound itself don't seem to impede us, and there are mostly a few UI restrictions (such as no UISlider) that impact AudioKit slightly. Even the plots seem to work out of the box with no further changes.

I've been having some issues with setting up LFS properly, so I hope all the files are included in here as they should. The binaries for Csound and libsndfile should be in LFS storage.


aure:
Looks great, I think this justifies a 2.3 release.

narner:
I'd definitely agree with that...this looks awesome @megastep!

PR #251, opened 2015-11-09
Making Csound wait time a parameter for some slower simulators

aure:
This seems to work fine, just making a pull request so I can do travis testing.


megastep:
I wonder if there's a cleaner way to do this, as it does sound like a nasty race condition.

aure:
Well, its no worse than currently, and AKManager is ripe for an overhaul as we know.

PR #260, opened 2015-12-02
Move ```AKBeatClock``` notifications over to the main thread so that …

dylan:
…it doesn't block the CSound thread.


aure:
This looks reasonable and if we don't use it exactly, this is still a step forward.  Thanks.

PR #271, opened 2016-01-27
Carthage support

megastep:
This adds the necessary support for Carthage, in the form of binary archives. As is, Carthage is not able to compile AudioKit properly (because of the location of the Xcode projects mostly), however it is able to fetch precompiled binaries as an alternative, and that works just fine.

Basically as part of the `build_archives.sh` script, a new `AudioKit.framework.zip` file is generated that needs to be uploaded to Github along the other zips for all releases. Then Carthage will automatically pick it up and recognize the architecture of each of the frameworks in there.

Next will be CocoaPods I guess, however this is obviously a better way to handle AudioKit since they use our blessed frameworks directly.

### DIFF
--- a/AudioKit.xcworkspace/contents.xcworkspacedata
+++ b/AudioKit.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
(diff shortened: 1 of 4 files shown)


megastep:
Also we could upload the zip to 3.0 right now to bring Carthage support for 3.0, though it would make sense to just bring this in with 3.0.1.

aure:
Seems like a fine way to do it. Thanks!

PR #273, opened 2016-01-29
Got rid of AKManager and reintroduced AKSettings

megastep:
This is a big one which changes and simplifies the central `AKManager` class, renaming it to the more intuitive `AudioKit` and moving it away from a singleton model. Now all can be done with class methods and properties. Playgrounds and examples have been updated accordingly.

`AKSettings` is back, also providing a namespace to update global settings for AudioKit. Had to patch up a lot of classes that used fixed 44100Hz sample rate for instance. Note that there are a few Soundpipe files that should probably take advantage of this, but C doesn't have access to this Swift class currently.
Also `AKSettings` will see more functionality soon (such as triggering stuff when values are changed), right now this is mostly refactoring the rest of the library to use its values.

Finally, I also took the liberty to bump the version to 3.0.1, although wondering if 3.1 might be more appropriate if we break the API.


aure:
I think it will be 3.0.1.  There haven't been any functional changes, just better code.  This is quite a PR for sure!

loopbum:
Truisms related to why I prefer programming AudioKit in Swift (even if the inner workings need to cooperate with AVFoundation announced at WWDC14).

AudioKit makes it easier to access the power to create — music, synthesis, signal 

“Programs must be written for people to read, and only incidentally for machines to execute.” 
― Harold Abelson <http://www.goodreads.com/author/show/5409448.Harold_Abelson>, Structure and Interpretation of Computer Programs <http://www.goodreads.com/work/quotes/871745>

“The computing scientist’s main challenge is not to get confused by the complexities of his own making.” 
― E. W. Dijkstra

~ Robert D. Thompson

PR #274, opened 2016-01-29
Updated CocoaPods support

megastep:
Small updates to the script that creates the frameworks zip for Carthage, so that the same file can also be used with CocoaPods. The new pod then will install our blessed binaries instead of copying and compiling the entire library. I still need to finish troubleshooting these odd issues with Bitcode, but this will work regardless.

Note that as is it will fail to validate until we push 3.0.1 and upload the binary file, but it worked when I was trying it with my own repo.

Also it would be a good idea to go and update the old pod spec for previous 2.x versions so that they refer to the AudioKitArchive repo now, if that's something we can do.

### DIFF
--- a/AudioKit.podspec.json
+++ b/AudioKit.podspec.json
@@ -9,6 +9,8 @@
 	"file": "LICENSE"
     },
     "homepage": "http://audiokit.io/",
+    "social_media_url": "http://twitter.com/AudioKitMan",
+    "documentation_url": "http://audiokit.io/docs/",
     "source": {
         "http": "https://github.com/audiokit/AudioKit/releases/download/v3.0.1/AudioKit.framework.zip"
     },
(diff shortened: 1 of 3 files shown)


aure:
I tried to change Cocoapods github url for previous versions but Cocoapods rejected them:

[!] Unable to accept duplicate entry for: AudioKit (2.3)

So, the best you could do would be to release a v2.4 and have the github url changed for that one.   However, since this is only slightly less broken, I don't think we should bother and just move forward with Cocoapods 3.0+

megastep:
Mmh, I wonder if it would be possible to do that by sending them a pull request directly and editing the pod specs there.

megastep:
I just did just that, we'll see how that goes. Hopefully somebody just merges it.

PR #279, opened 2016-01-31
Fix frameworks for Bitcode

megastep:
Well that was needlessly complicated to do, but it is now working.

A good way to test this is to try to validate with the Analog Synth X project (you'll have to create a new version on the App Store first). I did this with one of my apps and got it to the point where it didn't produce any errors with the precompiled frameworks.


megastep:
Also did what we talked about and got rid of the AKSettings bridge in Soundpipe.

aure:
Brilliant, so maybe tonight we can tag this as 3.0.1 and get on carthage/cocoapods!

PR #284, opened 2016-02-01
Develop

atomoil:
I've added a 'currentTime' variable to AKAudioPlayer as I need it for the app I'm developing.


aure:
Thanks very much for this!

PR #286, opened 2016-02-03
CoreMidi packet fixes, virtual midi ports, and pitchbend support for AnalogSynthX

essej:
MIDI packet handling needed to be updated to handle multiple events inside of a packet (which is completely legal and common) as well as fixing the handling of multiple packets. Took the opportunity to use some handy borrowed code to iterate over the midi packet list and packet events.

Added virtual midi port creation/destruction functions to AKMIDI so that other apps can see us and send MIDI while we are running in the background. (Requires the AKSettings.playbackWhileMuted to be true in order for background operation to work, incidentally).

Also added pitchbend support to AnalogSynthX, with a default midi pitch bend range of +/-2. I'll leave it to someone else to expose that config parameter in the UI, and possibly add a UI pitchwheel control for setting the globalbend parameter on the instrument.

### DIFF
--- a/AudioKit/Common/Internals/AudioKitHelpers.swift
+++ b/AudioKit/Common/Internals/AudioKitHelpers.swift
@@ -81,6 +81,15 @@ extension Double {
             self = min * exp(log(max / min) * self);
         }
     }
+    
+    /// Calculate frequency from a floating point MIDI Note Number
+    ///
+    /// - returns: Frequency (Double) in Hz
+    ///
+    public func midiNoteToFrequency(aRef : Double = 440.0) -> Double {
+        return pow(2.0, (self - 69.0) / 12.0) * aRef
+    }
+
 }
 
 // MARK: - MIDI Helpers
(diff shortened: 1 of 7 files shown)


aure:
This looks awesome, thanks!  Just going to let the tests run before I merge.

alejandroluengo:
Awesome work @essej 

Could you please add a short snippet showing how to create a (reachable by other apps) MIDI port?

essej:
As the example I added the line to AnalogSynthX's Conductor class which simply calls `midi.createVirtualPorts()`. That's all you need to do! In order for the audio session to stay around in the background you also have to do `AKSettings.playbackWhileMuted = true` before starting the engine (so it uses the right audio session category).

Where do you want a snippet?

aure:
Jesse, for the extension MIDIPacket: SequenceType do you think that it makes sense to call this an AKMIDIPacket (and similarly for MIDIPacketList), to avoid bashing the native CoreMIDI struct MIDIPacket?  Or am I missing the point and we want these to override the CoreMIDI versions?

PR #287, opened 2016-02-03
fixed midi packet parsing to include realtime system commands again

essej:
Just fixing an oversight regarding parsing the realtime commands with the new setup


aure:
Thanks for this fix!

PR #296, opened 2016-02-08
Add AKVariSpeed node

eirikurorri:
I'm using this for my own app and noticed it was missing.

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Time Pitch Stretching/AKVariSpeed.swift
+++ b/AudioKit/Common/Nodes/Playback/Time Pitch Stretching/AKVariSpeed.swift
@@ -1,0 +1,65 @@
+//
+//  AKVariSpeed.swift
+//  AudioKit For iOS
+//
+//  Created by Eiríkur Orri Ólafsson, revision history on GitHub
+//  Copyright © 2016 AudioKit. All rights reserved.
+//
+
+import AVFoundation
+
+/// AudioKit version of Apple's VariSpeed Audio Unit
+///
+/// - parameter input: Input node to process
+/// - parameter rate: Rate (rate) ranges from 0.25 to 4.0 (Default: 1.0)
+///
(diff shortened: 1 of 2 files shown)


aure:
I'm happy to accept this but I remember choosing not to include it because TimePitch seemed to be more capable than varispeed?  Is there a good reason to have this?  Perhaps efficiency?

eirikurorri:
You can indeed get the same effect by using TimePitch, but it will introduce artifacts, making it sound a bit grainy. The VariSpeed effect is a tape-like effect, which sounds very nice and smooth if you're willing to link playback speed with pitch.

For instance, I'm using it to slow down a loop by half, and the difference in sound quality is very noticeable.

There's a nice discussion of TimePitch vs Varispeed [here](https://developer.apple.com/library/prerelease/ios/documentation/AVFoundation/Reference/AVFoundationAudioSettings_Constants/index.html#//apple_ref/doc/constant_group/Time_Pitch_Algorithm_Settings).

And also, congrats on an awesome project! I've been having a blast using it.

aure:
Cool, good to know!

eirikurorri:
Awesome, thanks!

aure:
Are you into making a playground for varispeed?  If yes, I'll hold off, otherwise I'm happy to do it.

aure:
Actually, I think I will do the playground for it because playgrounds are changing and any work you'd do would probably have to get rewritten for Xcode7.3.

eirikurorri:
Right, makes sense. Excited about the upcoming interactive XCode 7.3 playgrounds, though!

loopbum:
Interactive Playgrounds with AudioKit would be incredible as I've known a few people that have been trying since playgrounds have come out. Looking forward to it !!!

aure:
Its already there in the Xcode7.3 branch if you want to try it out. (means also grabbing the latest xcode beta)

PR #297, opened 2016-02-11
Full Carthage support

ikesyo:
Carthage's multi projects support is released in [0.12](https://github.com/Carthage/Carthage/releases/tag/0.12).

You can now build the frameworks by `carthage build --no-skip-current` and the frameworks users can specify `--no-use-binaries` option and build them locally.


megastep:
Actually I had added that workspace to the project to try to get Carthage to work before 0.12, hoping it would pick it up. Since then I thought it was nice to have all 3 architecture projects in the same window to see the changes between them, so I wouldn't mind keeping it around. Does it really get in the way of Carthage?

megastep:
I say that mostly because the individual Xcode projects are further down in the subdirectories.

ikesyo:
> Actually I had added that workspace to the project to try to get Carthage to work before 0.12, hoping it would pick it up.

That does not work, because Carthage before 0.12 would stop searching projects when a project with at least one shared framework scheme is found, so only one of the three projects is used. Adding a workspace holding the projects does not affect the result since the workspace itself does not have any schemes.

> Does it really get in the way of Carthage?

The answer is in the commit log:

> This is noisy for Carthage and Xcode. __Since the projects have the schemes those are all the same name, Xcode can't pick up the correct project to build with for the command__: `xcodebuild -workspace AudioKit.xcworkspace -scheme AudioKit`. Removing the workspace, Carthage would use a command such as: `xcodebuild -project "AudioKit For OSX.xcodeproj" -scheme AudioKit`.

If you'd like to keep the workspace, the schemes should be renamed respectively such as: `AudioKit-OSX`, `AudioKit-iOS`, `AudioKit-tvOS`.

ikesyo:
I've restore the workspace and renamed the schemes: 15f7b31, 8089b87

megastep:
OK, my other main worry was whether this would interfere with our other build scripts, but this seems to be fine.

ikesyo:
Thanks for merging this!

PR #315, opened 2016-03-21
Added indefinite execution

ianrahman:
Allowed for continuous audio playback by adding needsIndefiniteExecution flag


aure:
Thanks!

ianrahman:
NP!

PR #324, opened 2016-03-30
adding support for DefaultToSpeaker

atomoil:
Might want to discuss this one as it restarts the audio engine when the configuration changes, but doesn’t restart any of the nodes. This means that there aren't crashes in code that expects the audio engine to be running (oh yes - the fun I had trying to work out why it was crashing :'(). 

I think that behaviour is ok as if I unplug the headphones from Music or YouTube then playback stops. However I’m not sure what the best behaviour would be so that consumers of AudioKit know that the configuration has changed / the audio engine has restarted.


aure:
In addition to discussing other ways to achieve this, this method doesn't pass travis, (haven't looked, but its likely not OSX or tvOS compatible as is), so we have to address that as well.  I'll look at this a bit more later, but I wanted to respond right away.

atomoil:
Drat!

Does Travis give any useful hints as to what it specifically doesn’t like?

Thanks for the speedy reply and I’m sure we can work it out.
On 30 March 2016 at 18:09:02, Aurelius Prochazka (notifications@github.com) wrote:

In addition to discussing other ways to achieve this, this method doesn't pass travis, (haven't looked, but its likely not OSX or tvOS compatible as is), so we have to address that as well. I'll look at this a bit more later, but I wanted to respond right away.

atomoil:
I've found the issue, and it's exactly what you suspected - DefaultToSpeaker isn't supported on tvOS (as per https://developer.apple.com/library/tvos/releasenotes/General/tvOS90APIDiffs/Swift/AVFoundation.html)

atomoil:
Oops, missed one...

atomoil:
Hurrah! It passes!
Now let's discuss if this is the right way of going about this or if there is a better way :-)

aure:
So, an alternative approach would be to put a gain object automatically on the microphone and turn it down to zero if the output is the speaker.  Is that doable?

atomoil:
So the issue that this is trying to fix is:

If the microphone is active and headphones are unplugged then audio is routed out of the phone earpiece speaker not the main speaker.

Specifying DefaultToSpeaker means it comes out of the main speaker not the phone earpiece speaker.

So as far as I know changing the graph around the microphone wouldn't change where the audio comes out.

Hope that makes sense.

aure:
I am not quite understanding "headphones unplugged then audio is routed of the phone earpiece"... if the headphones are unplugged how could it get there?  I am not sure how "earpiece" is different than "headphone".   What I need to understand is the problem you're trying to solve or the the behavior you want to occur.  I thought you wanted to avoid feedback, in which case if we could capture the even of the headphones being unplugged, turning off the microphone's output would be an okay solution.  But maybe capturing that event is not easy.  I'm also happy to do a skype or screenshare if talking this through over voice would help.

atomoil:
The thing I didn't realise until I added a microphone is that there are 3 places audio can play out of an iPhone:

- speakers (e.g. at the bottom of the device)
- earpiece (e.g. where audio comes out when you are on a call, next to the front camera)
- headphones (e.g. if a stereo jack is plugged into the headphone socket)

I hadn't thought of earpiece as an option, but if code activates the microphone then it appears that the OS routes audio out of the earpiece instead of the speaker.

What am I doing? Well, put simply I have an app when people record themselves with a(n optional) click track playing in the background. 
It's fine for the click track to be picked up on the recording. 
Feedback is not an issue as I record the microphone but don't have it wired up to the audio graph, so it doesn't come back out.

The problem is that as soon as the mic is active the player can't hear the click track anymore as it's coming out of the earpiece and is therefore very quiet. With the updated code it forces the OS to play the click track out of the bottom speaker, not the top / earpiece / next-to-front-camera speaker.

Just done a little googling and the earpiece is also known as `AVAudioSessionPortBuiltInReceiver` (as opposed to the normal speaker which is `AVAudioSessionPortBuiltInSpeaker`. 

Hope that makes a bit more sense. If not, let's Skype or somehow talk in realtime.

aure:
Honestly my inclination is to just accept the pull request because you have given it a lot more thought than anyone else could have.  Its exactly the kind of pull request I like to see- completely motivated by a real use case scenario.

PR #330, opened 2016-04-05
STK files added

PaulBatchelor:
Hopefully this is a good start towards getting STK inside of AK. 

Included here:

- STK source code (duh.)
- STK license
- A makefile that builds the source code (hopefully it works on OSX)
- Sample code using soundpipe with STK. 

To compile the soundpipe example (and STK):

run "make sp_mandol"


aure:
The makefile runs and builds fine on OSX and the test program sp_mandol, produces a good test.wav.  So, all systems go.  However, this doesn't do anything for AudioKit yet.  Question for you is whether you want to get your hands dirty on that Mac Mini and try to make STK work within AudioKit proper or if you want me to accept the pull request and get handed off the code.

PaulBatchelor:
I may want to try to port this into AK myself. This could be a good opportunity for me to understand AudioKit better.

PaulBatchelor:
And by "port this", I mean I'll try to port one of these. Are there any that you'd like to see in AK first?

aure:
I have no preference.  Probably something that uses the rawwaves, because they will have to be encoded into some text format.  Frameworks themselves can't really contain anything except code.  Files in frameworks aren't easily referenced so there would have to be extra instructions to copy a resource bundle, which we did in AK2 and I'd like to avoid.

PaulBatchelor:
Ah. Well in that case, maybe accept this pull request as is? It looks like
there is still stuff on my end that I have to do.

aure:
I think I'd like to see one thing work in AK before committing these files to the final repo.  I'm optimistic, just cautiously so.

aure:
Awaiting v2 of this pr...

PR #337, opened 2016-04-09
add 'replaceFile' to AKAudioPlayer

atomoil:
Means the graph can be setup once and then audio files can be played as needed without having to create new AKAudioPlayer instances or restart the audio engine.


aure:
Sure, why not :)

PR #346, opened 2016-04-12
re-added STK files... 

PaulBatchelor:
I initially was going to do some fancy git footwork to remove the binary files in my PR, but I found
it was just easier to blow up my repo, re-fork, and re-add the files.


aure:
Okay, still kinda hoping to clear out the really weird stuff like the tcp server included in STK, but at least those are just text files.

PR #347, opened 2016-04-13
Fixes an issue with currentTime not being as reliable as it could be.

atomoil:
Depending on the device the previous implementation had some wobble to it. This one is consistent across devices.


aure:
Thanks for your continued vigilance to keep AudioKit up to date with your latest finds!

atomoil:
It's my absolute pleasure. AudioKit is amazingly awesome - how could I possibly not want to help take it to 11 on the awesomes dial??!

PR #351, opened 2016-04-18
Added AKNotifications for detecting output change

aure:


aure:
I think "turnOnAKNotifications" would be better renamed "notificationsEnabled" to match the other Boolean settings.  "turnOn" is a verb/action and so it sounds like a method.

SNIP3R:
Ok, I made the name change and added a deinit so the observer can be removed. Let me know if there are any other issues with this. I get some weird behavior with AKSampler after AudioKit has been restarted, but I think that's unrelated. Everything else I've tested works fine after headphones are plugged in and playback is restarted.

aure:
Well done John!

PR #374, opened 2016-05-10
Update AKAudioPlayer.totalFrameCount upon reloadFile()

jsbean:
**What's in this pull request**

`AKAudioPlayer.totalFrameCount` is updated when the file is replaced.

`AKAudioPlayer.duration` is derived from `totalFrameCount` which was only set during the initialization phase of `AKAudioPlayer`. 

When replacing the file for `AKAudioPlayer` (via `replaceFile(_:)`), the `totalFrameCount`, and subsequently `duration` held the lingering value associated with the initial file given, which was no longer accurate.

See [this repository](https://github.com/jsbean/AKAudioPlayer-UpdateTotalFrameCount) for an iOS application with unit test demonstrating this change.

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
@@ -108,17 +108,6 @@ public class AKAudioPlayer: AKNode, AKToggleable {
         internalPlayer.volume = 1.0
     }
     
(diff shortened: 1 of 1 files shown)


jsbean:
**N.B.:** 

`The job exceeded the maxmimum time limit for jobs, and has been terminated` @ `47 min 3 sec`.

aure:
Thanks!

PR #392, opened 2016-06-09
AKMicrophoneRecorder Fix

roecrew:
Added an AVAudioSession. Without this, our AVAudioplayer's record() will return false, and an empty audio file will be created!


aure:
Thank you!

PR #393, opened 2016-06-10
Minor AKMicrophoneRecorder change

roecrew:
Forgot to add the option 'DefaultToSpeaker'


aure:
THanks, this breaks tvOS, but it shouldn't have microphone recorder anyway, I'll take care of that.

aure:
By the way, I just noticed you're making these pull requests from the master branch.  Development should really be done on the "develop" branch so that it can be tested before going into master.  In the future, I will be more careful about this.  Thanks!

PR #394, opened 2016-06-10
Ak audio file

laurentVeliscek:
Hi Aure,

I don't know if it is the correct way. But I'd like to create a pull request about AKAudioFile...


aure:
Can you make a pull request in to AudioKit's develop branch?  We are trying not to alter the master branch except when making releases.

PR #401, opened 2016-06-12
Minor

roecrew:
Added AKAudioFile build ref


aure:
Every little bit helps!   And... I love getting pull requests!  Thanks.

PR #406, opened 2016-06-14
Added code printing for iOS Playgrounds

narner:
I've added in code printing for the purpose of preset creation to all the iOS playgrounds where I thought appropriate. Let me know if I missed any. I've left the precision of the output to be `"%0.3f"`, let me know if that's ok as well. 

This pull request also has two commits that fix two Playground compile errors.


aure:
Bam!  Nice

PR #408, opened 2016-06-15
Added crest property to AKTriangleOscillator

roecrew:
Allows you to change the offset of the triangle wave's crest. It's sort of similar to a square wave's pulse-width mod.


aure:
You're on fire!

PR #416, opened 2016-06-17
Quick fixed adsr

roecrew:
ADSR was broken so I uncommented the quick fix and it works correctly now.
The problem was if attackDuration in Amplitude Envelope Playground  was greater than 12, decay would never be triggered. I suspect it was because of this conditional "if(*out > 0.9999)".


aure:
Hoping to replace ADSR with something better someday.

PR #420, opened 2016-06-18
Optimized bltriangle crest property

roecrew:
Discovered a bug in the process of this commit, please see issues.


aure:
So great to have you look at these things at such great depth.  Much appreciated!

PR #423, opened 2016-06-20
AKAudioNodeRecorder

laurentVeliscek:
cleaning and fixes of AKAudioFile and AKAudioPlayer before working on AKAudioNode

AKAudioFile :

- Fixed comments to be compliant with XCode contextual help (and Jazzy?)

- Base Directory enums values are now set using CamelCase (Apple’s developper guideline)
so baseDir is now .Resources or .Temp or .Documents
( no change to ExportFormat enum to avoid weird « .Wav » or « .Aif » formats)

- Finally kept most errors checking… (was afraid to break something ;-)

- some init external names have been changed: 
file = try?  AKAudioFile(forReadingWithFileName: "mixloop", andExtension: "wav", 

AKAudioPlayer

- simplified code.
- reloadFile() throws errors (must now use try? reloadFile() )
- replaceAudioFile has been renamed replaceFile and throws errors  (must now use try? replaceFile()

playgrounds: 
updated and tested all playgrounds that used AKAudioPlayer. (70 files ! :-)

I changed some text in the nonAudioPlayground  to load a file using AKAudioFile.
By the way, I don’t understand the « drag » zone in this playground (intriguing…)

Fixed the free sound link (https instead of http)

I discovered that in playgrounds, you could use try without catching errors when you are at the tom level. (!)
Convolution and Newton’s cradle still fail but failed before (it is not my fault !… ;-)


aure:
Great strides, keep them coming!

aure:
BRW, in Swift 3 base enum values will start with lowercase, so we'll be going to that nicer standard soon.

PR #425, opened 2016-06-20
AKNodeRecorder

laurentVeliscek:
With an AKTapeRecorder playground page to demo :-)


aure:
There needs to be some OS checks in there to make it work on OSX but I can handle that.

PR #427, opened 2016-06-21
Added depth parameter to AKTremolo

roecrew:


aure:
Hey, this is great, thanks.  You're really digging deep.  Would you be interested in becoming a core team member?  There aren't many perks, except you get access to our internal Slack and Trello, and show up on the about us page on audiokit.io.  We do get occassional contract work if you're interested on that level as well.

roecrew:
Tremolo clicks at higher frequencies were present before this request. It seems to get worse as the input frequency decreases and the tremolo rate increases.

aure:
choosing a different curve line sine for the faster tremolo is essential.

roecrew:
@aure  Oooooo that's why.
And yes! I'd love to become a core team member. Thanks!

PR #435, opened 2016-06-25
Updated AudioPlayer Playground

laurentVeliscek:
Now, a playhead read-only slider and a textfield reflect playhead’s
position


aure:
Cool!

PR #444, opened 2016-06-30
Initial preset work

narner:
-Added code printing to mandolin playgrounds
-Added in extensions for physical model presets

After looking at the [physical models in the archive](https://github.com/audiokit/AudioKitArchive/tree/d55b01862aaa94a844321805ecba580d9b952eb7/AudioKit/Operations/Signal%20Generators/Physical%20Models), I think it makes sense to just create presets from scratch, since there are inconsistencies between the parameters in physical models in the two versions.


aure:
I don't want empty files in AudioKit so let's just keep this open until there is something there of value.

narner:
@aure I removed those blank files for the time being, and added some presets for some effects. 
It looks like the build is failing at the moment though because the new Table Test Instrument isn't importing `AKFoundation.h`.

aure:
Great start!

PR #450, opened 2016-07-04
Develop

laurentVeliscek:
AKAudioFile Update:

Fixed AKAudioPlayer: Now, the played file is opened « for reading »
It avoids a stupid crash when AKAudioPlayer try to set the
FramePosition if the file was instantiated « forWriting »

Fixed AKAudioFile:
Opening a file from resources worked in playgrounds, but not in an app
(!):  Fixed !
Exporting a file then worked in apps, but not in playground: Fixed !

Changed convenience writeIn init. Removed file extension, settings and
interleaved in init fields
Now, the file is ALWAYS a .caf set with AKSettings format.
As you still can instantiate the old / hard way (using NSURL and
manually set settings and format), user can create an AKAudioFile for
writing with his custom settings if he wishes to.
Then, it is not necessary to provide init fields that could lead to
errors (like creating a .m4a with linearPCM, for example…). We assume
the default AKAudioFile for writing should always be a.caf stereo file
set with AudioKit settings and format.

Added new lazy properties:

.pcmBuffer -> return  file’s samples as an AVAudioPCMBuffer
.arraysOfFloats -> return  file’s samples as [[Float]]
.maxLevel -> return the file’s peak value in dB.

Added a new Inits :

init(createFileFromFloats floatsArrays:[[Float]],
baseDir: BaseDirectory = .Temp,
name: String) throws { …

That lets you create an AKAudioFile from scratch, using array(s) of
floats

added init from AVAudioPCMBuffer.

So now, you can get the samples as floats Arrays, play with them, and
create an AKAudioFile set with the new floats values. (cool :-)

Added 3 new methods that return new AKAudioFiles:

.normalize() with an optional newMaxLevel init field, that let’s you
set the new maxLevel  expressed in dB.
.reverse(): returns an AKAudioFile with reversed samples
.append(file:AKAudioFile): returns a new AKAudioFile with with audio
samples appended from another file.
.extract(from inSample, to outSample): returns a new AKAudioFile with a
range of audio samples

and a static method :
.silent() that lets user generate silent AKAudioFile from any duration.

What is cool now:
- You can now convert convert an mp3 / m4a / mp4 audioFile to
.caf/.wav/.aif PCM file (it was not possible using AVAssetExportSession)

- you can now do some basic audio edits: extract a part of a file,
reverse it, append it to another file, add some silence, normalize,
convert to m4a… So we have a fully programmable audio editor.

\o/

### DIFF
--- a/AudioKit/Common/Internals/AKAudioFile.swift
+++ b/AudioKit/Common/Internals/AKAudioFile.swift
@@ -13,16 +13,16 @@ import AVFoundation
 
 /// Audio file, inherits from AVAudioFile and adds functionality
 public class AKAudioFile: AVAudioFile {
-    
+
     // MARK: - embedded enums
-    
+
     /**
      BaseDirectory enum
 
      - Temp: Temp Directory.
      - Documents: Documents Directory
      - Resources: Resources Directory (Shouldn't be used for writing / recording files).
-     
+
      Ex: let directory = AKAudioFile.BaseDirectory.Temp
 
      */
@@ -49,7 +49,7 @@ public class AKAudioFile: AVAudioFile {
         case mp4
         case m4a
         case caf
-        
+
         // Returns a Uniform Type identifier for each audio file format
         private var UTI: CFString {
             switch self {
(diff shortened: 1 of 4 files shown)


aure:
Wow, this is a big one!  Thank you!

PR #457, opened 2016-07-07
remove syntax-breaking space in func call

zmcartor:
Very small change. Space within variable name breaks playground compilation.


aure:
Thanks, this playground is removed in the upcoming release, so this is not going to be relevant soon.

PR #466, opened 2016-07-11
AVSession and buffer settings

laurentVeliscek:
- Now you can disable any AVAudioSession category setting from AudioKit.
- Added a bufferLength and a recordingBufferLength AKSettings
properties.

Successfully tested in a minimal recorder app with my iPhone4S

Will probably be improved, as AVSession handling is far to be perfect.
But couldn’t wait to share as it works really fine here.


aure:
Great!  I added some os checking compiler directives to get it to compile on OSX.

PR #474, opened 2016-07-13
Adding TOC and header to banks playgrounds

narner:


aure:
Thanks, these probably need footers too?

PR #494, opened 2016-07-23
Develop2

laurentVeliscek:
So at least, we get some clues when nothing works…


aure:
Accepted and immediately deprecated the microphone recorder :)

PR #495, opened 2016-07-23
Minor doc addition

roecrew:


aure:
Every little bit helps and this will be quite useful to those digging in.

PR #498, opened 2016-07-25
Updated UI for RecorderDemo example for iOS

linouk23:


aure:
Really nice update, thanks!

PR #506, opened 2016-08-01
Bit Crusher sample rate range change

roecrew:
SRate val of 0 breaks bitcrusher.


aure:
281 files changed?  I think you wanted to create a pull request off of develop instead of master.

PR #527, opened 2016-08-19
memory free error in EZAudioUtilities

ryanfrancesconi:
I was getting sporadic crashes using the plots on OS X until I changed this code. As I understand EZAudio isn't supported any longer, I'm letting you know here.

// I believe the order should be:
    free(historyInfo->buffer);
    TPCircularBufferCleanup(&historyInfo->circularBuffer);
    free(historyInfo);


aure:
Thanks very much.  We will be keeping our EZAudio up to date so its great that you let us know.

ryanfrancesconi:
Excellent - it seems stable to me now with this change.

PR #531, opened 2016-08-28
Added the zeroes extension method to array 

adamnemecek:
* used this method in AKTable extensively instead of append
* for an array of 4096 elements this  is ~10x faster than appending one-by-one

### DIFF
--- a/AudioKit/Common/Internals/AKTable.swift
+++ b/AudioKit/Common/Internals/AKTable.swift
@@ -109,109 +109,109 @@ public struct AKTable {
 
     /// Instantiate the table as a triangle wave
     mutating func standardTriangleWave() {
-        values = [Float]()
+        values = [Float](zeroes: size)
         let slope = Float(4.0) / Float(size)
         let phaseOffset = Int(phase * size)
         for i in 0 ..< size {
             if (i + phaseOffset) % size < size / 2 {
-                values.append(slope * Float((i + phaseOffset) % size) - 1.0)
+                values[i] = slope * Float((i + phaseOffset) % size) - 1.0
             } else {
-                values.append(slope * Float((-i - phaseOffset) % size) + 3.0)
+                values[i] = slope * Float((-i - phaseOffset) % size) + 3.0
             }
         }
     }
 
     /// Instantiate the table as a square wave
     mutating func standardSquareWave() {
-        values = [Float]()
+        values = [Float](zeroes: size)
         let phaseOffset = Int(phase * size)
         for i in 0 ..< size {
             if (i + phaseOffset) % size < size / 2 {
-                values.append(-1.0)
+                values[i] = -1.0
             } else {
-                values.append(1.0)
+                values[i] = 1.0
             }
         }
     }
 
     /// Instantiate the table as a sawtooth wave
     mutating func standardSawtoothWave() {
-        values = [Float]()
+        values = [Float](zeroes: size)
         let phaseOffset = Int(phase * size)
         for i in 0 ..< size {
-            values.append(-1.0 + 2.0 * Float((i + phaseOffset) % size) / Float(size))
+            values[i] = -1.0 + 2.0 * Float((i + phaseOffset) % size) / Float(size)
         }
     }
 
     /// Instantiate the table as a reverse sawtooth wave
     mutating func standardReverseSawtoothWave() {
-        values = [Float]()
+        values = [Float](zeroes: size)
         let phaseOffset = Int(phase * size)
         for i in 0..<size {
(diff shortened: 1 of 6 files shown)


aure:
Excellent!

bishop800:
Excellence

aure:
The one thing I should have commented but didn't notice is that you should make pull requests from forks of the develop branch back to the develop branch so that changes can get tested thoroughly before releasing onto master.  I've since merged this change back to develop.  I hope to see more awesome pull requests from you so that's why I am mentioning it after the fact.  Thanks again.

adamnemecek:
Alright, thanks for the heads up. Yeah I do plan on committing a bunch actually, I think I'm going to use AudioKit for a project. I'm going to start with some stuff with the MIDI part of the codebase (at least for starters). 

Are you guys doing anything unusual with the build system btw? I tried to install the framework system wide (I do that a bunch so that I can write quick programs to try stuff out) but I keep getting an error that it was build for a target with a different name. I haven't seen this error before and the internet didn't help.

Also where do you guys stand on the "good practices" vs speed spectrum? I'm guessing in the swift part it's more towards legibility, in the c/c++ part it's more towards speed? 

Also are there currently any parts that are yet to be ported to swift?

aure:
Well, the MIDI stuff is working well, but is definitely in the need of some careful inspection.  At least we can probably write more midi tests on 3.3 to ensure any changes don't break anything.  Stephane has a fix-frameworks.sh in the Frameworks folder that might give some clues.  We're eventually going to speed up compilation by using a precompiled soundpipe and sporth, but that will only help with the post-swift compilation which is slow to begin with.  

As for best practices, we try to be as swifty as possible without throwing the baby out with the bathwater, so speed and efficiency are still the most important.   I know we can do better in that department too.  

When you're ready I can invite you to the AudioKit slack group if you want quick access to the core team members.

megastep:
It's possible the current OS X framework is somewhat broken if that's the one you are trying to install globally. That's probably why we don't have it in CocoaPods right now but fortunately I have a fix for it that will be in 3.3.1.

For iOS/tvOS there are requirements when submitting binaries to the App Store (including embedded frameworks) so that we need to strip out the simulator slices from them. This is the main purpose of the fix-frameworks scripts, though it has the added benefit of also making the binaries smaller.

aure:
Also, as far as porting to Swift, there are a few things to do on that front.  What types of things are you interested in?  If you're into physical models a la the AKMandolin, lots of work to be done on that front.

adamnemecek:
@aure: Ok, I would really appreciate the invite to the slack group. You might also want to consider setting up a public one. Google Groups is so 2005 :-).  My email is <github-username>@gmail.com

The midi part works fine but it could be cleaned up. There's quite a lot of duplication in the event iteration (c.f. the isEmpty and clearMetaEvents) that could be wrapped up in some a sequencetype.

What's your migration plan toward swift 3.0? 

Also, I was considering wrapping up all the error handling code in a function using autoclosure. Something to the tune of (I can't think of a good name though):

func tryCall<T>(@autoclosure f: () throws -> T, _ args: Any..., _ msg: String = #function) throws -> T {
  do { 
    return try f()
  }
  catch let error as NSError {
    print("error \(error) \(msg) \(args)")
    throw error
  }
} 

and you'd call it with "try tryCall(throwingFunction())" and "tryCall(throwingFunction())" in swift 3.0.

So I'm actually not super good at synthesizer development, so I'm probably going to start with things other than modeling but I'm planning to learn.

@megastep Yeah, I was. Do you think you could commit it to a new branch or something? Or send me the diff?

megastep:
@adamnemecek This is the commit that fixes it: https://github.com/audiokit/AudioKit/commit/aa1e759630ddbf7f7229d61f3882f2e747b1a65b

You'll still need to rebuild the OS X library yourself with the build_frameworks.sh script there, alternatively you could even use the current OS X binary and simply move out the .dSYM file that's inside the framework. It's the cause of the problems.

aure:
Drop me an email to my audiokit address and I'll send you a slack invite

adamnemecek:
@megastep I can't get this to work, still the same error. I build it with the build_frameworks.sh script. I moved out the dsym file but it seems to be taking the project name from AudioKit.framework/Modules/AudioKit.swiftmodule/x86_64.swiftmodule which is a folder with binary files.

adamnemecek:
I think what would really help me is like the intended development workflow. Like how do you guys develop this?

aure:
I just build using AudioKit for whateverOS.xcodeproj and build playgrounds to test additions to the framework.

aure:
On slack we can also do a screenshare and share setups and such.

adamnemecek:
@aure ok i sent you my email

PR #532, opened 2016-08-29
Modify knobs & sliders custom controls to appear in IB / Thanks to Aure!

swiftcodex:


aure:
Not sure why Travis wasn't passing, but I don't think there's anything wrong with this code.  Release the hounds.

aure:
I do agree with the hound that variable names like knob120_indicator are highly non-Swifty.

swiftcodex:
Super non-swifty. Agreed. It should probably be smallKnobIndicator

swiftcodex:
I wonder who created those property names? ;)

PR #535, opened 2016-08-30
Minor improvements

adamnemecek:
* Consolidated random element generation into one method of CollectionType
* removed some unnecessary constant specifiers


aure:
Perhaps minor to you, but clean ups like this go a long way.  Thanks!

PR #539, opened 2016-09-05
AKCallbackInstrument more usable

laurentVeliscek:
Changed callbacks into callback? :
- an array is unnecessary and less efficient.
- make it optional avoid some tricky things (egg-and-chicken dilemma
with callback definition and instrument init)


laurentVeliscek:
Hi !

please just change line 24:
    public init(callback: AKMIDICallback? = nil) {

(forgot to do it before pull request)

megastep:
Your commit with the fix is already in the PR. ;)

PR #541, opened 2016-09-07
Removed last forward splash from filePath in AKAudioFile init for Temp Directory

afnan-ahmad:
`NSTemporaryDirectory()` returns tmp directory path with forward splash **/** already added at the end.

This was causing problem issue while using `AKAudioFile`'s processing functions when target file was in tmp directory.


aure:
Thanks!

PR #553, opened 2016-09-18
Playground Non-Audio tutorial Swift 3.0

lluisgerard:
I was taking at look at this library for the first time, going through the Playgrounds and there was some changes missing in the Non-Audio Tutorial for **Swift 3.0**.


aure:
Yep, thanks!

PR #561, opened 2016-09-21
Updates to Sequencer example

narner:
-Buttons now show touches on highlight
-Tempo value is now displayed


Examples/iOS/SequencerDemo/SequencerDemo/ViewController.swift
```diff
@@ -108,4 +110,9 @@ class ViewController: UIViewController {
         bassButton?.isEnabled = true
         snareButton?.isEnabled = true
     }
+
```
narner: I think this may be incorrect - this is just a space between methods.

Examples/iOS/SequencerDemo/SequencerDemo/Conductor.swift
```diff
@@ -90,6 +90,7 @@ class Conductor {
         sequence.setTempo(100)
         sequence.play()
     }
+
```
narner: ...same as above comment.

aure: The hounds don't like white space at the end of lines, or lines that just contain some spaces when they could be returns.  Its complaining about horizontal white space, not vertical.  Its a minor thing anyway.

PR #563, opened 2016-09-23
Fix issue with loop running x60 slower on iOS10

ColinEberhardt:
The playground loop code uses a counter that counts up to a trigger. This logic depends on the loop function being called 60 times a second.

Currently this loop runs vvvvveeeerrrryyyyyy ssssslllloooooowwwwllllyyy


aure:
Thanks very much.  I think this happened due to an Xcode Swift 3 suggestion, but still, an embarrassing mistake!

ColinEberhardt:
No worries, I've spotted a few other minor issues so will raise a few more PRs

PR #574, opened 2016-10-10
Ignore DerivedData folder

magicwave:
Ignores local DerivedData folder created by Xcode, where Preferences > Locations > DerivedData may be set to "Relative".


aure:
Thanks and Welcome!

PR #580, opened 2016-10-13
fix DefaultToSpeaker

atomoil:
in the swift 3 update 
`try AKSettings.setSession(category: .playAndRecord, with: .defaultToSpeaker)`

got changed to:
`try AKSettings.setSession(category: .playAndRecord, with: .mixWithOthers)`

this is just a little fix for that - presuming that it wasn't deliberate - if it was then I can think of another approach :-)


aure:
It was deliberate because we're finishing up Audiobus support which requires that setting so I thought I'd make it the default.

atomoil:
Ah, fair play - current code breaks the default to speaker option (which for me is to play a metronome out of the speaker while recording), so sounds like we need extra logic - I'll have a look and submit a new pull request. :-)

Also I think my configuration change notification code is slightly redundant now there is route change notification code, so I should prolly remove it - do u know who added the route change code, as might be worth me having a quick chat with them about it before I break something :-o

Best,

Adam

aure:
Also, just invited you to our slack since you're an advanced user.  It will give you quick access to everyone.

aure:
I'll close this for now.

junhohong:
@atomoil check out #586

atomoil:
So... just picking this one up after the post on the audio kit google group (sorry got distracted elsewhere).

I think my original submission was correct - if someone has specified 'defaultToSpeaker' then it should follow that behaviour, however that obviously conflicts with the audioBus support, but I don't know anything about audioBus support :-s

So, currently we have:

```
if AKSettings.audioInputEnabled {
	if AKSettings.defaultToSpeaker {
		try AKSettings.setSession(category: .playAndRecord,
								  with: .mixWithOthers)
	} else {
		try AKSettings.setSession(category: .playAndRecord)
	}
}
```

I _think_ it should really be:

```
if AKSettings.audioInputEnabled {
	if AKSettings.defaultToSpeaker {
		try AKSettings.setSession(category: .playAndRecord,
								  with: .defaultToSpeaker)
	} else {
		try AKSettings.setSession(category: .playAndRecord,
								  with: .mixWithOthers)
	}
}
```
(note the updated 'mixWithOthers' in the else)

The other option - as per #600 would be to change to:

```
if AKSettings.audioInputEnabled {
	try AKSettings.setSession(category: .playAndRecord,
								  with: .mixWithOthers)
	if AKSettings.defaultToSpeaker {
		try! AKSettings.session.overrideOutputAudioPort(AVAudioSessionPortOverride.speaker)
	}
}
```

But I don't know enough about audioBus to make an informed decision about what would work best - any thoughts?

aure:
Let's go ahead and make that fix and then I'll test with Audiobus, perhaps adding a configuration item for that.

PR #581, opened 2016-10-17
Song processor updates

esimonian:
Increase the volume to 10
Add Bit Crush an Pitch Shifter effects


aure:
I think you mean to create this pull request onto AudioKit's develop branch, please resubmit it!

PR #588, opened 2016-10-26
Develop

ryanfrancesconi:
I've made a few small changes which I need in my application. I don't think they would negatively impact any other project. One thing that you'd probably want to consider is the ability to export the AKAudioFile to a location other than the predefined 3 you had. I have added just the ability to export to the same location as the input file, which is the functionality that I needed. I'm only currently developing for OSX and am not considering iOS specific features.

Thanks!
Ryan Francesconi


aure:
Looks good to me!

ryanfrancesconi:
thank you!
Ryan
 
 
-----Original Message-----

PR #592, opened 2016-10-27
AKAudioPlayer updates

ryanfrancesconi:
The main change here is adding the ability to schedule the buffer in the future. There are a few small notes in comments as well. I don't believe this breaks any existing functionality, but it will need testing in other circumstances than mine to be sure!

The only conceptual difference is that this class would previously schedule the buffer on initialize. This means that the first attempt to schedule in the future would be ignored and the buffer would immedately play. Removed that

The only change to your code is moving the buffer setup to the play event rather than the stop. It's working fine here in my project, but as the changes are significant to an important class. Testing!

Let me know if you see anything weird about the way I added this! I think it's a useful feature and a mandatory one for me.

Thanks!
Ryan

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
@@ -11,10 +11,9 @@ import AVFoundation
 
 /// Not so simple audio playback class
 open class AKAudioPlayer: AKNode, AKToggleable {
-
-
+    
     // MARK: - Private variables
-
+    
     fileprivate var internalAudioFile: AKAudioFile
     fileprivate var internalPlayer = AVAudioPlayerNode()
     fileprivate var audioFileBuffer: AVAudioPCMBuffer?
@@ -25,27 +24,32 @@ open class AKAudioPlayer: AKNode, AKToggleable {
     fileprivate var lastCurrentTime: Double = 0
     fileprivate var paused = false
     fileprivate var playing = false
-
-
+    
     // MARK: - Properties
-
+    
     /// Will be triggered when AKAudioPlayer has finished to play.
     /// (will not as long as loop is on)
     open var completionHandler: AKCallback?
-
+    
     /// Boolean indicating whether or not to loop the playback
     open var looping: Bool = false
-
+    
+    
     /// return the current played AKAudioFile
     open var audioFile: AKAudioFile {
         return internalAudioFile
     }
-
+    
+    // path to the currently loaded AKAudioFile
+    open var path: String {
+        return audioFile.url.path
+    }
+    
     /// Total duration of one loop through of the file
     open var duration: Double {
         return Double(totalFrameCount) / Double(internalAudioFile.sampleRate)
     }
-
+    
     /// Output Volume (Default 1)
     open var volume: Double = 1.0 {
         didSet {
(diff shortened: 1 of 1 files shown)


laurentVeliscek:
HI Ryan,

Interesting. The reason I choosed to set the buffer on stop (and init) was to avoid a laggy play. So playing starts as soon as you ask for. With long files, setting buffer on play can be a problem, especially if you wish to play a few files at the same tile and needs a minimum sync.

But it depends on your needs.

Did you experiment with longer files ?

(With short files, sampler is a better option than akaudioplayer)


What is the reason you wish to change this ?

Best !

Laurent.
Envoyé de mon iPhone

laurentVeliscek:
I didn't check the code. But if it is an option and if you still can use akaudioplayer the way it worked (set buffer on stop), I have no objection.

:-)

Laurent.

Envoyé de mon iPhone

ryanfrancesconi:
Hi Laurent, I'll do more testing with long files. Most of my testing has been with files 2 minutes or under, so you may be right. I'm developing on a 2007 mac - so I like to think of my system as the slowest possible one to run AudioKit!

However, I still don't think stop() is the correct place to set the buffer unless you aren't using startTime and endTime. Have a look at the attached pict of a portion of what I'm working on. I have a fairly typical DAW where the files are being scheduled based on the position in the timeline sync'd to video generally. When the app is in the stopped or play state you can edit the regions, essentially resetting their startTime and endTimes.

So, I think the alternative in my case is to call setPCMBuffer manually when the edit is done. And change it to open rather than fileprivate. Given that - what do you guys suggest?

image

ryanfrancesconi:
I'll see if I can optimize this a bit more. I have a few ideas.

laurentVeliscek:
Very impressing project !

I understand your point. but for video synced DAW, you must be able to
manage 2 hours long files.
(my job is sound engineer for TV programs, and I can tell that it happens
very often)

If you fill the buffer when play is pressed, you may get a very very slow
responsive DAW.
And sync will probably be very hazardous (that's a major problem, phase
inconsistency between tracks on a DAW is unacceptable)
There's no way to ensure that many AKaudioplayers will play in sync.
So I 'm not convinced that AKAudioPlayer is the right engine to achieve
your goal.

Please check your change with huge files.
and check what happen with many tracks...

Filling the buffer can be very long with long files. And If it becomes
laggy, I'd prefer to keep the default behaviour of AKAudioPlayer like it is.


For such a purpose, may be AVComposition classes would be more appropriate
(https://developer.apple.com/reference/avfoundation/avmutablecomposition).
But I've never experimented with it.

Let us know...

Another direction would be to "schedule segment" with an AVAudioPlayerNode.
(But no looping then, that's why I prefered the scheduleBuffer method. And
I remember to have some bugs with completion handler not beeing triggered
correctly)

Good luck !

Laurent.


2016-10-28 18:41 GMT+02:00 Ryan Francesconi <notifications@github.com>:

ryanfrancesconi:
Yes - actually I was using the schedule segment method before trying to merge with AKAudioPlayer. I'll experiment what is the best solution.

Previously you were calling stop() before your play command. So actually you were filling the buffer on both stop and before play previously. The main thing that doesn't work for me is scheduling the buffer on stop and initialize.

I'll work on it some more and see if I can find a good solution. I considered the mutable composition route, but am hoping that I don't need to do that!

PR #595, opened 2016-10-28
AKAudioPlayer: refining the buffer filling

ryanfrancesconi:
I've refined the fill buffer routine more now. Also, I've changed the name of setPCMBuffer() to updatePCMBuffer(), which is more accurate to what it's doing.

The buffer is updated only on these occasions now:
* initialize()
* set startTime
* set endTime
* on immediate playback within a sound file (when the startTime is being set on the fly)

Previously the buffer was being filled with the same data over and over (on stop() which was called too often) even if it didn't need to change. So, this should be more efficient. I've tested with 20 concurrent samples all playing at the same time and it was solid. I do think the current implementation is better suited towards shorter sound files as Laurent mentioned. More on that soon.

One other recommended change: I've commented out the auto-play within the startTime/endTime. I think that this should be in the specific application control class rather than within the API. If you are doing edits while a group of sounds are playing, if one starting playing again immediately when you didn't ask it to, it would be a bug.

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
@@ -24,6 +24,9 @@ open class AKAudioPlayer: AKNode, AKToggleable {
     fileprivate var lastCurrentTime: Double = 0
     fileprivate var paused = false
     fileprivate var playing = false
+    fileprivate var internalStartTime: Double = 0
+    fileprivate var internalEndTime: Double = 0
+    
     
     // MARK: - Properties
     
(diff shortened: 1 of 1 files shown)


aure:
This all seems reasonable to me.  Probably don't need to comment out the stuff, just remove it with good git commits, but I see why you're doing it to be extra clear and careful.

laurentVeliscek:
Excellent! 

It seems that you fixed some stupid cpu wasting. 

I thought you wanted to set buffer at every "play" action but that's not the case and thank you for fixing unnecessary buffer setting.

You're absolutely right about renaming set buffer into update buffer.

Thanks !

:-)

Laurent.

Envoyé de mon iPhone

ryanfrancesconi:
cool thanks guys - It felt a little bold of me to remove your code, that's why I left it in commented out in case you didn't agree. But, I'll clean it up now!

PR #596, opened 2016-10-29
EZAudioFile updates

ryanfrancesconi:
I've added support for embedded markers for WAVE/AIFF files into EZAudioFile. Since you have to deal with the AudioToolbox C stuff to dig these out, I thought this'd be the best place to put this. I've added a new class EZAudioFileMarker to facilitate simpler Swift access over the native C structs that are returned.

I'm not positive this is the best way to import to Swift, but it is working fine like so:

```
            if let markerList = _file.markers {

                Swift.print("Marker List: \(markerList)")
                
                for afm in markerList {
                    let m = afm as! EZAudioFileMarker
                    Swift.print( "name:\(m.name) markerID: \(m.markerID) framePosition: \(m.framePosition) type: \(m.type) ")
                }
```

### DIFF
--- a/AudioKit/Common/Internals/EZAudio/EZAudioFile.h
+++ b/AudioKit/Common/Internals/EZAudio/EZAudioFile.h
@@ -307,7 +307,7 @@ typedef void (^EZAudioWaveformDataCompletionBlock)(float **waveformData, int len
 
 /**
  A list of markers associated with the audio file. These are typically embedded within WAVE or AIFF files only.
- @return A NSArray containing the markers in the file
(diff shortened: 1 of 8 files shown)


aure:
This looks good but the test script is failing on build, I think because you haven't added the new files to the non-Mac frameworks.  I can either attend to that when I get back (traveling today) or if you want to append some commits to this pull request, that of course would be great too.

ryanfrancesconi:
Let me know if I understood you correctly. Added EZAudioFileMarker.* to iOS and tvOS. Sorry, I haven't ever even opened those projects...

aure:
Yep, that was all you need to pass the tests. Thanks!

PR #612, opened 2016-11-11
Renamed AKToggleable to AKPlayable

adamnemecek:
I hope I didn't miss any.


aure:
What's the benefit of this name change?

PR #613, opened 2016-11-11
* added a the clamp method in an extension to closedrange 

adamnemecek:
...and use it dynamicsprocessor


aure:
Awesome, love stuff like this, keep them coming!

PR #615, opened 2016-11-12
Added extensions to AudioComponentDescription to simplify construction

adamnemecek:


aure:
Merging this even though its on master instead of develop, its that good. :) 👍

adamnemecek:
Sorry again about committing to master. I still have one outstanding to master where I just removed a file from the three projects but from then on I'm going to start committing to develop.

aure:
No worries, these were all great improvements I'm happy to make on all branches.

PR #622, opened 2016-11-13
Added Clarinet STK physical model

jorgegarcia:
Updated project, added STK Clarinet internal dependencies, added AKClarinet node.


aure:
Cool, has some issues but I will accept and fix.

PR #630, opened 2016-11-16
removed unnecessary AUParameterTree variable declarations

yyshin:


aure:
Another person committed to clean code, I'm loving it!

PR #636, opened 2016-11-27
arrayOfFloats -> floatChannelData?

dylan:
- Change signature to match pcmBuffer for consistency and descriptiveness.
- Encourage nil checking rather than passing around empty arrays.
- Pre allocate our arrays before we use them to reduce memory thrashing and increase perf.


aure:
Much better, not even sure "float" is even needed in the variable name, but this is definitely an improvement.

PR #649, opened 2016-12-09
Fixed bug where A5 (880 Hz) was used as temperament reference instead…

magnus80a:
… of A4 (440Hz)

Previously playing midi note 57 would result in a pitch of 220 Hz which is A3 and not A4 which it should be unless I've missed something. This corrects that.


aure:
This fix is incorrect.  The midi note to frequency conversion is straight from this:

https://en.wikipedia.org/wiki/MIDI_Tuning_Standard

I think in all likelihood the bug you're experiencing has to do with the generator itself, probably not treating stereo properly and doubling the number of calculations.  Is this the case for all AudioKit generators?

magnus80a:
Thank you, that link made me realize why I'm constantly off by an octave. The midi chart I've been using as a reference starts at C0 and not C1 like the one AudioKit uses.

PR #653, opened 2016-12-12
Added support for multi-stream interfaces

Joebayld:
On interfaces that contain more then 1 stream, this method would fail. Reason being because you needed to get the size of the buffer, but originally it was defaulted to 0 and thus failed.

Also renamed a couple variables to make it more readable.

Let me know what you think.


Joebayld:
Should this have gone to the Develop branch? My apologies if it's incorrect. It's my first pull request!

aure:
yeah, it probably should have gone to the develop branch, can you resubmit?

PR #654, opened 2016-12-12
Added support for multi-stream interfaces

Joebayld:
On interfaces that contain more then 1 stream, this method would fail. Reason being because you needed to get the size of the buffer, but originally it was defaulted to 0 and thus failed.

Also renamed a couple variables to make it more readable.


aure:
Thanks!

PR #657, opened 2016-12-12
* small refactoring of the init callback

adamnemecek:


aure:
You really like method names starting with _?  Its looking a bit weird like that.

adamnemecek:
I weighted a couple of options and I think that the underscore is nice because 

a.) it's a symbol used for the synthesized variables names so _ can mean "private-ish".
b.) It indicates that you should watch out (which you should)

That being said, I'm all ears if you have some suggestions.

aure:
Yeah, I thought about alternatives, and couldn't come up with one so I accepted the PR.

PR #658, opened 2016-12-14
Fix delay issue with AKFFTTap

gubikmic:
AKFFTTap had long delays of more than 1000ms.
Since AKNodeFFTPlot worked well and has a larger buffer (1024 instead of 512), I tried that value in AKFFTTap and it resolved the issue.

A clue also came from AVAudionode.installTap where the documentation says "Supported range is [100, 400] ms." That doesn't seem to be true since 1024 works fine but 1024 / 44100/s = 0.02322..s < 100ms, i.e. outside the documented range, but it served as clue that the buffer might be too small.

The other thing I noticed is that `fftData` should be half the size. Since the input for the FFT is real the first half of `fftData` was always zero anyways.

So it's two changes in one (which might be confusing), so I've split it in two commits.

Lastly, I wonder if `fftData` shouldn't just be Float. (?)

### DIFF
--- a/AudioKit/Common/Taps/AKFFTTap.swift
+++ b/AudioKit/Common/Taps/AKFFTTap.swift
@@ -11,11 +11,11 @@ import Foundation
 /// FFT Calculation for any node
 @objc open class AKFFTTap: NSObject, EZAudioFFTDelegate {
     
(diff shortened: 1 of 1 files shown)


aure:
Thanks, probably changing to float would also be a good idea.

PR #659, opened 2016-12-14
Pevent the script emit errors when all is ok

carlosypunto:
Actual script error

### DIFF
--- a/Frameworks/fix-framework.sh
+++ b/Frameworks/fix-framework.sh
@@ -10,8 +10,19 @@ binary="$framework/AudioKit"
 
 if [ "$ACTION" == "install" ]; then
 	find "$framework/BCSymbolMaps" -name \*.bcsymbolmap -type f -exec mv {} "$CONFIGURATION_BUILD_DIR" \;
-	lipo -remove i386 -output "$binary" "$binary"
-	lipo -remove x86_64 -output "$binary" "$binary"
+
+	BYNARY_INFO=`lipo -info AudioKit`;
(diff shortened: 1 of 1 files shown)


aure:
Looks good to me but this is Stephane's territory.  You want to review and merge this @megastep ?

megastep:
OK, I guess this is a safer way to do it, though I'm wondering why this is happening to you.

carlosypunto:
Hi @megastep, I think I'm done everything in a normal way, but I can not guarantee either, because it has been doing a thousand tests that I did not go into production on the AppStore. If the issue happens again I will give you more information about it.

At the moment I can only tell you that I have this Podfile:

```ruby
platform :ios, '9.0'

target 'Dytective' do
  use_frameworks!

  pod 'Alamofire', '~> 4.2.0'
  pod 'AudioKit', '~> 3.5'
  pod 'CryptoSwift', '~> 0.6.6'
  pod 'R.swift', '~> 3.1.0'
  pod 'RealmSwift', '~> 2.1.1'
  pod 'RxCocoa', '~> 3.0.1'
  pod 'RxSwift', '~> 3.0.1'
  pod 'SwiftyJSON', '~> 3.1.3'
  pod 'RappleProgressHUD', '~> 2.1.2'
end

post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings['SWIFT_VERSION'] = '3.0'
    end
  end
end
```

PR #692, opened 2017-01-15
fun times. note that some of this is work in progress (e.g. the avaudionode extension + AKKernelUnit) but i need to keep track of it somewhere

adamnemecek:


aure:
Looks like good stuff - the start to some very good updates.

Issue #682, opened 2017-01-12
AUV3 Effects memory leak


kowongh:
I have come across some memory leaks concerning the C code in Soundpipe when instantiating some AudioKit AVAudioUnit/AUAudioUnits. I've checked the leak issue with multiple AudioKit effects units and they all seem to point to sp_create malloc which is called from the effect's DSPKernels, and SoundPipe>Modules>base.c

It seems to when we start the engine (after they've been connected and attached) and sometimes when we are instantiating the units... so it leads me to believe something is happening when we start up the audio hardware.

### > Bytes Used	                             # Leaks		  Symbol Name
    320 Bytes   90.9%	3	 	                              +[AVAudioUnit instantiateWithComponentDescription:options:completionHandler:]
    320 Bytes   90.9%	3	 	                               APComponent::NewInstance(unsigned int, bool, void (OpaqueAudioComponentInstance*, int) block_pointer)
    320 Bytes   90.9%	3	 	                                AUIB_OpenAsync(void*, OpaqueAudioComponentInstance*, unsigned int, void (OpaqueAudioComponentInstance*, int) block_pointer)
    320 Bytes   90.9%	3	 	                                 AUAudioUnitSubclassInstance::OpenAsync(OpaqueAudioComponentInstance*, unsigned int, void (OpaqueAudioComponentInstance*, int) block_pointer)
    320 Bytes   90.9%	3	 	                                  AUAudioUnitSubclassInstance::Open(OpaqueAudioComponentInstance*)
    320 Bytes   90.9%	3	 	                                   -[AUAudioUnit initWithComponentDescription:error:]
    320 Bytes   90.9%	3	 	                                    -[AKMoogLadderAudioUnit initWithComponentDescription:options:error:]
    256 Bytes   72.7%	2	 	                                     sp_create
    256 Bytes   72.7%	2	 	                                      malloc
    256 Bytes   72.7%	2	 	                                       malloc_zone_malloc
>

aure:
I'm going to need a sample project or something to be able to act on this issue.

kowongh:
Hey Aure, I just ran the SamplerDemo. Thanks!

aure:
Was this analyzing the simulator or the app running on a device?  I have had some spuriously reported memory leaks on the simulator.

ryanfrancesconi:
[...]

PR #697, opened 2017-01-19
Fixes for Issue #682

ryanfrancesconi:
I've gone through all the AUV3 classes that I found and fixed the closures that had strong references to self in them. This should eliminate the retain cycles in these classes.

I've ONLY tested this with AKAmplitudeTracker -- its retain cycle is now fixed and deallocating immediately! As the code is almost identical with all 60 classes altered, I think they should all work. Still, that was a lot of busywork and I might of missed something.

Should definitely look for this self in closures in other classes too. **[weak self]** !


aure:
Thanks dudebro

PR #700, opened 2017-01-20
Fix crash with AKMIDISampler and AKSequencer

allanweir:
It looks a rewrite of how the MIDIPacketList iterator worked between version 3.5 and 3.5.1 led to crashes in an app which used MP3's or WAV's via AKMIDISampler.

The guard statement in MIDIPacketList+SequenceType which deallocated packets was triggering a object freed before being allocated error. I assume there was a reason it was rewritten so if there's a way to reincorporate those changes whilst making it safe for use in the situations I've found let me know and hopefully we can solve it.

There's also a small addition to AKAudioPlayer which adds a resume() method which doesn't jump to the start of a track after starting again after a pause.


aure:
One problem is that we work off of the develop branch.  Any chance you could make the pull request against the develop branch instead?

aure:
I merged this back into develop myself, thanks very much for the PR!

PR #701, opened 2017-01-20
* refactored SessionCategory

adamnemecek:


aure:
I don't know why this is failing, but I'll find out by accepting and then debugging, because it all looks good to me.

Issue #686, opened 2017-01-13
Pause / Play broken on AKAudioPlayer


allanweir:
Hey,
I've been working on a project with AudioKit since a couple of releases ago and it seems in at least the latest version that calling .pause() and then .play() isn't working as expected.

I have a single AKAudioPlayer with an MP3, connected to an AKMixer.
Calling .pause() will pause as expected
Calling .play() the first time will end the track and trigger .completionHandler
Calling .play() a second time will play the track from the start of the file, since it now thinks it's back at the beginning

I've tried calling .reloadFile() and play(from: pausedTime) but that has the same effect.

Has anyone else found this?

aure:
Can you demonstrate this happening in a playground or a shared project?  I just tested the audio player playgrounds and they seem okay, but I don't doubt what you are experiencing or that there could easily be something wrong with AKAudioPlayer.

allanweir:
Hi Aure,

Sure. I've just tried in the Audio Player playground and I get a broken sound stuck on repeat after I try to resume playback.

If you add the following code to the Playback > Audio Player playground you should hear the same thing when you press the toggle pause button

```
addSubview(AKButton(title: "Toggle Pause") {
            if player.isPlaying {
                player.pause()
                return "Paused"
            } else {
                player.play()
                return "Resumed"
        }})
```

allanweir:
I've just had a play around some more. So when AKAudioPlayer has looping set to true, I get a glitchy sound on repeat. If looping is set to false then the track immediately finishes when I try to resume

ryanfrancesconi:
Hi Allan,
I've replicated your issue. I will see if I can fix it today.
Thanks, Ryan

ryanfrancesconi:
Should be good now pending Aure's merge. Thanks for the report!

aure:
Fixed by https://github.com/audiokit/AudioKit/pull/702

PR #702, opened 2017-01-20
Fix for issue: Pause / Play broken on AKAudioPlayer #686

ryanfrancesconi:
Buffer was being overwritten in the start() method despite pause being true.


aure:
Thanks very much!

ryanfrancesconi:
I think it was my fault actually! :/

aure:
Welcome to the club!

PR #705, opened 2017-01-23
AKAudioPlayer reverse playback - pull#2

ryanfrancesconi:
Sorry about the last weird git pull blast. Here's a concise one with only the correct changes. I'll add an example of the usage in the playground.


aure:
Thanks for the redo of the PR and for the playground too!

ryanfrancesconi:
I haven't been able to use the playgrounds until I got a new mac this week - they didn't run at all for me on my old computer. So, will be better now about making examples when I add something!

PR #708, opened 2017-01-24
Added deinit methods to cleanup framework references

allanweir:
I create and recreate AudioKit setups for a few different situations in my app and realised some underlying audiokit references weren't being cleared up. This seems to remove the ones I've experienced without any issues.

### DIFF
--- a/AudioKit/Common/MIDI/AKMIDISampler.swift
+++ b/AudioKit/Common/MIDI/AKMIDISampler.swift
@@ -114,5 +114,13 @@ open class AKMIDISampler: AKSampler {
     open override func stop(noteNumber: MIDINoteNumber, channel: MIDIChannel) {
         samplerUnit.stopNote(UInt8(noteNumber), onChannel: UInt8(channel))
     }
+    
+    // Discard all virtual ports
+    open func destroyEndpoint() {
+        if midiIn != 0 {
+            MIDIEndpointDispose(midiIn)
+            midiIn = 0
+        }
+    }
 
 }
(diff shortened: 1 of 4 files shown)


aure:
Cool, thanks!

ryanfrancesconi:
should this of been merged into develop?

allanweir:
Hey, yeah sorry that was my fault for doing it initially to the wrong branch.

ryanfrancesconi:
yeah, should undo that merge, aure and maybe allan do a new pull request to develop? I know for me, this will break my code as I'm already doing what you added elsewhere - but I'm happy to alter my stuff if this looks good.

ryanfrancesconi:
AKAudioPlayer on the dev branch is also ahead of the master - so be sure to merge the changes

aure:
Sorry guys I will clean this up today.  Probably just by merging master back into develop and seeing what happens

aure:
Its on develop now.  Probably will leave it in master instead of worrying about reversing the commit and just plan to have a 3.5.2 version out soon.  @ryanfrancesconi  Ryan, if there are changes you want, just make a pull request, thanks!

ryanfrancesconi:
I'm all good with this change. One question though - I didn't have the issue that Allan says in terms of the AKAudioPlayer hanging around after deallocation -- I only had that problem with the AUv3 stuff. Also, I don't see this issue in other similar nodes. For example an AKMixer is deallocated despite having made many attach calls. Just a question here - but is this deinit call necessary? Or, should we add this to all nodes that access the static engine attach method, not just AKAudioPlayer ?

Is it possible Allan, that your memory leak was elsewhere?

allanweir:
Hey, so my memory leak wasn't with the AK classes but the AV ones that they created. Instruments would have no references to any AudioKit classes any more but it was still holding on to AVAudio ones, and the AKSequencer was deallocated but the MusicSequence and MusicTrack instances weren't. Are you seeing the same thing from AKAudioPlayer?

ryanfrancesconi:
Ahh, You're right! Thanks so much for this catch. One more leak plugged. We should look for this in other places.

PR #712, opened 2017-01-26
Let AKSampler propagate exceptions

megastep:
I think it would be a good idea to be able to get more information when the various loading functions for `AKSampler` fail. This changes the APIs to let the methods throw any raised exceptions.

Also cleaned up a little bit what was being sent to `AKLog()` for consistency.


aure:
Other than forcing the tries in the example I'm all for this kind of change.

PR #715, opened 2017-01-27
Memory error in DSPKernel

ryanfrancesconi:
There was a double freed error happening in DSPKernel.hpp 
(*** error for object 0xbaddc0dedeadbead: pointer being freed was not allocated) 

This was because the destroy() method of AKAmplitudeTrackerDSPKernel was being explicitly called but the destructor was also being triggered, the same pointer was being freed twice in both the destructor and in the destroy(). Please check this as I'm seeing some weird behavior about when the destroy() methods are being called. I'm seeing destroy() being called on objects that I never instantiated personally! 

This fix solves the crash I was seeing, but perhaps there are other situations? 

This crash is related to adding deinit { engine.detach() } to AKNode as well as the previous memory fix I did for the self in closures for AUv3. Reason being, I believe beforehand because the memory was being leaked and the objects held onto - these destroy methods were never called. This is my theory anyway! Once things are properly released, this bug was revealed. BOOM!

### DIFF
--- a/AudioKit/Common/Nodes/AKNode.swift
+++ b/AudioKit/Common/Nodes/AKNode.swift
@@ -46,6 +46,7 @@ extension AVAudioConnectionPoint {
     }
     
     deinit {
+        //AKLog("* AKNode")
         AudioKit.engine.detach(self.avAudioNode)
     }
 }
--- a/AudioKit/Common/Nodes/Analysis/Amplitude Tracker/AKAmplitudeTracker.swift
+++ b/AudioKit/Common/Nodes/Analysis/Amplitude Tracker/AKAmplitudeTracker.swift
@@ -82,6 +82,10 @@ open class AKAmplitudeTracker: AKNode, AKToggleable, AKComponent {
         halfPowerPointParameter?.setValue(Float(halfPowerPoint), originator: token!)
     }
 
+    deinit {
+        AKLog("* AKAmplitudeTracker")
+    }
+    
     // MARK: - Control
 
     /// Function to start, play, or activate the node, all do the same thing
@@ -94,7 +98,4 @@ open class AKAmplitudeTracker: AKNode, AKToggleable, AKComponent {
         internalAU!.stop()
     }
     
-    deinit {
-        AKLog("* AKAmplitudeTracker")
-    }
 }
--- a/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
@@ -236,6 +236,7 @@ open class AKAudioPlayer: AKNode, AKToggleable {
     }
     
     deinit {
+        AKLog("* AKAudioPlayer")
         AudioKit.engine.detach(internalPlayer)
     }
     
--- a/AudioKit/Common/Internals/CoreAudio/DSPKernel.hpp
+++ b/AudioKit/Common/Internals/CoreAudio/DSPKernel.hpp
@@ -120,10 +120,13 @@ public:
     }
 
     ~AKSporthKernel() {
+        //printf("~AKSporthKernel(), &sp is %p\n", (void *)sp);
+        // releasing the memory in the destructor only
         sp_destroy(&sp);
     }
+    
     void destroy() {
-        sp_destroy(&sp);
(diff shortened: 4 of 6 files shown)


aure:
@ryanfrancesconi This is passing the test and it looks fine but I just wanted to also ask @adamnemecek to look at since he introduce the AKSporthKernel.  Also, you've changed AKAmplitudeTracker but wouldn't a lot of the nodes be affected in exactly the same way?  I've got a couple of other comments which I'll add to the lines directly.

AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
```diff
@@ -236,6 +236,7 @@ open class AKAudioPlayer: AKNode, AKToggleable {
     }
     
     deinit {
+        AKLog("* AKAudioPlayer")
```
aure: Is this descriptive enough?  I for one wouldn't know what this meant when it appeared in the console.

AudioKit/Common/Nodes/AKNode.swift
```diff
@@ -46,6 +46,7 @@ extension AVAudioConnectionPoint {
     }
     
     deinit {
+        //AKLog("* AKNode")
```
aure: I'm see a lot of these added commented out prints/logs.  Can't they just be comments?  Were they just around for debug?  Not a big deal, just thought I'd bring it up.  (also, lets always use AKLog and not print so it can be suppressed more easily).

AudioKit/Common/Nodes/Analysis/Amplitude Tracker/AKAmplitudeTracker.swift
```diff
@@ -82,6 +82,10 @@ open class AKAmplitudeTracker: AKNode, AKToggleable, AKComponent {
         halfPowerPointParameter?.setValue(Float(halfPowerPoint), originator: token!)
     }
 
+    deinit {
+        AKLog("* AKAmplitudeTracker")
```
aure: Again, is this descriptive enough?  / Do we need it only for debug?

AudioKit/Common/Nodes/Playback/Player/AKAudioPlayer.swift
ryanfrancesconi: Interestingly when AKLog actually outputs that it shows up as:
deinit * AKAudioPlayer

That's pretty clear.

AudioKit/Common/Nodes/AKNode.swift
ryanfrancesconi: Because AKNode is in a bunch of spots, I didn't leave that in - however in debugging the AmplitudeTracker deinit it was helpful to see when it was called. Technically, we don't need it in there! Could just remove it.

AudioKit/Common/Nodes/Analysis/Amplitude Tracker/AKAmplitudeTracker.swift
ryanfrancesconi: same deal, shows up as:
deinit * AKAmplitudeTracker

ryanfrancesconi:
The memory release is in the DSPKernel class, so all the subclasses would inherit this change. I didn't change AKAmplitudeTracker in this pull request in any code way other than comments. The previous changes I made I had to go into each AUv3 class and make changes.

ryanfrancesconi:
One other comment that I forgot to mention in the pull request. I am seeing the ~destructor being triggered everytime, but I don't see the destroy() method called everytime. I only see destroy() called if I connect up a bunch of units. If not as many connections were made, it just does the ~. I didn't trace this too deep as it seemed enough for me to simply deallocate &sp in the destructor only.

ryanfrancesconi:
But yes, please whoever wrote that stuff, please check it as maybe I'm not understanding the sources of the crashes properly.

AudioKit/Common/Nodes/Analysis/Amplitude Tracker/AKAmplitudeTracker.swift
ryanfrancesconi: yes, only need for debugging

aure:
Thanks Ryan!

ryanfrancesconi:
You got it! Hope I am understanding everything - but it seems good to me!

PR #716, opened 2017-01-28
midipacketlist sequencing 2: the unsuckification

adamnemecek:


aure:
@adamnemecek Well, this passes our tests, and it definitely makes sense, but I bet some things will break. :) I'll accept it and see what happens.

PR #718, opened 2017-01-30
Develop3

adamnemecek:


aure:
I'm sorry, but Apple is inconsistent in the way they use both dry wet mix and wet dry mix, but in AudioKit we always use dry wet mix in that order because dry is 0 and wet is 1.  I like the other changes but you'll have to make everything else you changed back to dry wet.

aure:
Also, you committed a change to the playgrounds making the synthesis ones "macOS" and they should start off as "iOS" not a big deal, but worth mentioning.

aure:
Apple has in AUReverb2: kReverb2Param_DryWetMix and even though they might be moving to a wetDryMix standard in their Swift interface, I think that's fundamentally wrong and a direction I don't want AudioKit to take.

adamnemecek:
Fair enough lol. You make solid points.

aure:
Lastly, its such a commonly used node that changing the name is a bigger deal than most name changes (even the examples are breaking).

aure:
Can you please remake the good changes in this pull request?

Issue #742, opened 2017-02-11
Choosing the bank in AKSampler.swift


rrc2soft:
Hello,

I am making a simple DAW, and one of the things I need is the ability to choose the bank when loading a SF2 sound. I was checking the AKSampler class, and it seems that is not possible right now.

Is it possible to add something like this (code is probably wrong, as I write in Objective C):

```
    fileprivate func loadSoundFontBank(_ file: String, preset: Int, bank: Int) throws {
        guard let url = Bundle.main.url(forResource: file, withExtension: "sf2") else {
            fatalError("file not found.")
        }
        do {
            var bMSB: Uint8
            if (bank <= 127) {
                bMSB  = kAUSampler_DefaultMelodicBankMSB;
            } else {
                bMSB  = kAUSampler_DefaultPercussionBankMSB;
            }
            var bLSB: Uint8 = bank % 128;
            try samplerUnit.loadSoundBankInstrument(
                at: url,
                program: MIDIByte(preset),
                bankMSB: MIDIByte(bMSB),
                bankLSB: MIDIByte(bLSB))
        } catch let error as NSError {
            AKLog("Error loading SoundFont \(file)")
            throw error
        }
    }

    open func loadSoundFontBank(_ file: String, preset: Int, bank: Int) throws {
        try loadSoundFont(file, preset: preset, bank: bank)
    }
```

aure:
That seems like a good idea. Can you try it to make sure it works and then create a pull request?

rrc2soft:
I will integrate the library first (I am currently using the amazing audio engine), and once it is integrated and tested I will create the pull request :-)

rrc2soft:
Added pull request. Bear with me, as this is my first one :-).
Unfortunately I might not be able to use this library for my music composition app, as both the instruments and the signal chain must be changed dynamically :-( (cf. https://groups.google.com/forum/#!topic/audiokit/ySeznsj-c80). At least, I am happy I could contribute with something.

aure:
Oh well, too bad, but thanks for the contribution!

PR #744, opened 2017-02-12
Added loadBankSoundFont

rrc2soft:
Added and tested a new function that allows users to choose a certain bank when loading a preset in a soundfont. Check issue #742 .


aure:
So, this looks great, but normally you'd want to make the pull request to our develop branch.  But since this is new functionality and master is not that out of date, I'll accept it here and merge it to develop.  Thanks!

PR #751, opened 2017-02-16
Change to project settings

ryanfrancesconi:
EZAudioFileMarker wasn't included in the public headers group therefore wasn't visible outside the framework. Let me know if I have fixed this correctly! It is working for me now. I didn't realize this error till I started a new project and the class wasn't visible. (I had a hardcoded bridging include in my original project)
Ryan


aure:
Happy to add this, thank you!

Issue #709, opened 2017-01-24
RecorderDemo example error: exportAsynchronously(ios)


masaldana2:
Hello,
I run the original example of RecorderDemo and It compiles with only one error for exportAsynchronously:

`Export Failed Optional(Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSLocalizedFailureReason=An unknown error occurred (-12936), NSLocalizedDescription=The operation could not be completed, NSURL=/Users/miguelsaldana/Library/Developer/CoreSimulator/Devices/79E361A8-8DA9-4E2B-8264-0ED34E1EFCB7/data/Containers/Data/Application/BD8F35B0-2A75-4046-95CC-BF30709C52D3/tmp/DBAE1838-D059-422D-B66B-23454350F109.caf, NSUnderlyingError=0x60000024ebe0 {Error Domain=NSOSStatusErrorDomain Code=-12936 "(null)"}})`

anyone got a solution?
It does work with .wav to m4a but I want to convert .caf to .m4a :)
Thank you

aure:
I don't have a solution, but thanks for making the issue.  Will keep it open until it can be addressed.

tomw:
I have this too, in my own project and in that example. I can't seem to record and export, which is a pretty fundamental problem. Any workarounds until this is fixed?

masaldana2:
I used AVAssetExportSession and make my own converter.
its easy couple lines.

aure:
@masaldana2 Care to share your code?

masaldana2:
yes, im sorry im still new to github
here


`    func convertAudiotoM4A(name:String, cafURL: URL) -> URL{
        let url         = NSURL.fileURL(withPath: cafURL.absoluteString)
        
        let audioURL = AVURLAsset(url: url)
        
        
        let exported = AVAssetExportSession(asset: audioURL, presetName: AVAssetExportPresetAppleM4A)
        exported?.outputFileType = AVFileTypeAppleM4A
        
        let cacheDir = NSTemporaryDirectory()
        let filePath:String = cacheDir + name + ".m4a"
        print(filePath)
        
        //delete prev recording m4a if there's already one
        let fileManager = FileManager.default
        do{
            try? fileManager.removeItem(atPath: filePath)
        }catch{
            print("couldnt delete old caf")
        }
        
        
        exported?.outputURL = NSURL(fileURLWithPath: filePath) as URL
        
[...]

PR #756, opened 2017-02-21
Create URL for export session with fileURLWithPath

tomw:
This is a proposed fix for #709. It re-creates the URL for the `AVAssetExportSession` using `fileURLWithPath`. There might be a nicer way of doing this elsewhere, but I think this is the safest fix for me to make without more knowledge of the framework.


aure:
Thanks!

terenzeyuen:
@aure this is still not fixed, should use .path. Made a PR 
https://github.com/audiokit/AudioKit/pull/867

PR #759, opened 2017-02-24
added touch logic polyphonic and mono keyboard modes

crashingbooth:
(first pull request, so apologies if I'm a little off on github practices)

This implementation seems to behave touch-wise as I would expect a touch keyboard to behave.  In mono mode, if there are multiple keys being pressed down and the key which is actually being sounded is removed, it will replace it with the key from the highest remaining touch.

One limitation of this implementation is that if a touch originates outside of the AKKeyboardView frame and is dragged onto a key, that touch won't register.  I'm sure that there are ways around that in a project, but the workaround would involve code in the superview.  I don't think that limitation is a big deal anyway.
I will keep playing around with other improvements here.

### DIFF
--- a/AudioKit/iOS/AudioKit/User Interface/AKKeyboardView.swift
+++ b/AudioKit/iOS/AudioKit/User Interface/AKKeyboardView.swift
@@ -5,8 +5,8 @@
 //  Created by Aurelius Prochazka, revision history on Github.
 //  Copyright (c) 2017 Aurelius Prochazka. All rights reserved.
 //
-
 import UIKit
+
 
 /// Delegate for keyboard events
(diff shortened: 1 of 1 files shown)


aure:
Cool! Don't worry about these hound notifications, they're nothing important.  I'll look at the code later today.

aure:
Just FYI for next time, we like to develop on the "develop" branch, so you should make the pull request to that branch instead of master.  In this case, there's very little chance of mission critical things being broken in master, so I'll accept it and then merge onto develop, where we can make further updates if needed. Thanks!

aure:
For the record this is working great!  THANK YOU!

PR #772, opened 2017-03-03
fixing swiftlint issues

russellbstephens:


aure:
Hi Russell, but I think these might already be taken care of on the develop branch, which is where updates should be pull requested to.

russellbstephens:
thanks!

PR #788, opened 2017-03-10
Proposed fix to EZAudioPlot continuous redraw of non-rolling waveform

geoffsdavis:
Hey Aure,

I'm back from the dead.

Hey, in EZAudioKit's EZAudioPlot.m file, I've discovered that the **displayLinkNeedsDisplay** method is getting continuously fired from EZAudioDisplayLink's CADisplayLink target callback, of which EZAudioPlot is a delegate. It appears this was intended for rolling audio waveform buffers, so that buffers would be continuously rendered by EZAudioPlot. The delegate method triggers a redraw every time the display link fires, which is basically all the time.

However, with static non-rolling buffers, there's no need for the expensive redraw to keep getting fired. I have three separate instances of EZAudioPlot on my app, and on an old iPad 3, this resulted in a continuous chewing up of 30% more CPU, about 10% per audio plot.

I think I've figured out a good solution which seems to work well so far. Check out the diff below. The delegate method only performs a redraw if it's a rolling waveform, and the updateBuffer method does a manual redraw if it's non-rolling. EZAudioPlot is ultimately a UIView, and if the user changes the plotType to or from rolling, this code should handle the change on the fly.

### DIFF
--- a/AudioKit/Common/Internals/EZAudio/EZAudioPlot.m
+++ b/AudioKit/Common/Internals/EZAudio/EZAudioPlot.m
@@ -353,6 +353,7 @@ UInt32 const EZAudioPlotDefaultMaxHistoryBufferLength = 8192;
         case EZPlotTypeBuffer:
             [self setSampleData:buffer
                          length:bufferSize];
+            [self redraw];
             break;
         case EZPlotTypeRolling:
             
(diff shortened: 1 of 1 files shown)


aure:
Looks good to me!

aure:
So, actually we had to revert this change, so if you need the waveform to be static, we'll have to add a static boolean or something.  Waveforms are not necessarily static.

aure:
I'm thinking something like this would do the trick for you:
        plot.shouldOptimizeForRealtimePlot = false

davidjaystrauss:
Hi there @aure I just [asked a question](https://stackoverflow.com/questions/61259771/audiokit-shouldoptimizeforrealtimeplot-causing-plot-to-go-blank) relating to this functionality in SO, having a bit of trouble if you've got some spare time and wouldn't mind taking a quick look.

PR #794, opened 2017-03-14
Fix memory leak when init() is called more than once

pcbeard:
Noticed this while debugging. The question is, should all init() methods be safe to call more than once?

### DIFF
--- a/AudioKit/Common/Internals/Soundpipe/modules/ptrack.c
+++ b/AudioKit/Common/Internals/Soundpipe/modules/ptrack.c
@@ -92,7 +92,7 @@ typedef struct peak
 
 int sp_ptrack_create(sp_ptrack **p)
 {
-    *p = malloc(sizeof(sp_ptrack));
+    *p = calloc(1, sizeof(sp_ptrack));
     return SP_OK;
 }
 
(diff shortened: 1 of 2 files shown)


aure:
Ideally you'd make your pull requests against the develop branch, but since I don't think these files have any difference I'll accept it and merge it into master myself.  Thank you!

aure:
And, to answer your question, I guess we should cover that potential problem, just seems pretty rare.  But, if you want to make the changes in a pull request, that would be cool.

pcbeard:
Thanks. I'll make future pull requests relative to develop instead of master.

aure:
@PaulBatchelor One of the changes here is a change to soundpipe within AudioKit. Can you look at the change and see if it is something you want for Soundpipe proper?

PaulBatchelor:
@aure just checked. I can see the usefulness in AK, but I'm keeping it out of Soundpipe. 

Reasons:
- I don't want init functions to call create functions. These stages should absolutely be separate. 
- I want to minimize the practice of comparing NULL in my code. I find it to be bad practice.
- calloc zeroes out the memory too. nothings really added to it. Not worth the mass code changes.

PaulBatchelor:
> The question is, should all init() methods be safe to call more than once?

@pcbeard for soundpipe-based modules, unfortunately not. Some init functions, in addition to initializing variables, allocate buffers. Poor planning on my part, I know. I've been thinking about possible fixes for a long time, but most of them are messy and require sweeping changes.

pcbeard:
So the real bug here was to call init() again without first calling destroy().

PaulBatchelor:
yeah. that works too :)

PR #800, opened 2017-03-16
Silence string interpolation warnings

pcbeard:
I see these warnings with Xcode 8.3 beta 4

### DIFF
--- a/AudioKit/Common/Internals/AKNodeRecorder.swift
+++ b/AudioKit/Common/Internals/AKNodeRecorder.swift
@@ -186,7 +186,7 @@
                 try fileManager.removeItem(atPath: path)
             }
         } catch let error as NSError {
-            AKLog("AKNodeRecorder Error: cannot delete Recording file: \(audioFile?.fileNamePlusExtension)")
+            AKLog("AKNodeRecorder Error: cannot delete Recording file: \(String(describing: audioFile?.fileNamePlusExtension))")
             throw error
         }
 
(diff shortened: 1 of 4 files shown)


aure:
What warnings are suppressed by this?  Its seems pretty unnecessarily verbose to have to convert to String explicitly when using string interpolation.

aure:
Maybe it is giving warnings because these are optionals?  I actually don't mind it printing out optional.

pcbeard:
Yes, doing string interpolation with an optional is now generating a warning. I assume this change (suggested by the compiler) tells the compiler we wanted that behavior.

pcbeard:
Here's the warning:

AudioKit/AudioKit/Common/Internals/AKNodeRecorder.swift:189:73: String interpolation produces a debug description for an optional value; did you mean to make this explicit?
AudioKit/Common/Internals/AKNodeRecorder.swift:189:85: Use 'String(describing:)' to silence this warning
AudioKit/AudioKit/Common/Internals/AKNodeRecorder.swift:189:85: Provide a default value to avoid this warning

aure:
While I love getting pull requests, I don't like this solution.  First of all, its possible that something changes in Xcode before it comes out of beta which allows us to leave the code unchanged, and secondly its a really inelegant solution.  I like Xcode's second suggested option better of providing a default value.  Perhaps a nil coalescing operator would do the trick.  At least then we have an opportunity to provide a bit more info than just relying on String(describing) to do whatever it will do.

pcbeard:
I have no problem with your assessment. I'd just like to see these warnings handled. Perhaps there's a compiler switch to suppress them.

Issue #675, opened 2017-01-04
Bluetooth iOS 10


themichael123:
Hi, 

since iOS 10 it is no longer possible to use a bluetooth headset in combination with AudioKit (input, processing, output). 

Can this be fixed ?

aure:
Can you provide more information?  Are you saying this worked with AudioKit in iOS9?  Do you have an example project you can share that demonstrates this so we can test?  Closing until more information.

themichael123:
Hi Aure. 

Yes, this worked in iOS 9. Please find the Xcode Project attached. As you can see, the microphone input is not taken from a connected bluetooth headset, but rather from the internal mic. 

Thanks.
[MicrophoneAnalysis.zip](https://github.com/audiokit/AudioKit/files/686985/MicrophoneAnalysis.zip)

aure:
Hi, we're working on this, its not forgotten.

themichael123:
Thanks Aure.

aure:
I'm not sure if its taken care of by this, but here's some updated bluetooth stuff https://github.com/audiokit/AudioKit/pull/770

themichael123:
Thanks. Will test and let you know...

pcbeard:
I'm assuming the problem boils down to this:

if AKSettings.audioInputEnabled and AKSettings.useBluetooth are both true, then on iOS 10 the opts parameter is set to [.allowBluetoothA2DP, .mixWithOthers]. According to the documentation,

> If this option and the allowBluetooth option are both set, then when a single device supports both the Hands-Free Profile (HFP) and the Advanced Audio Distribution Profile (A2DP), the Hands-Free ports will be given a higher priority for routing.

Even if this isn't the exact cause, it seems like more than a bool is needed to control how Bluetooth is configured. Some kind of enum, or perhaps a value of type AVAudioSessionCategoryOptions that lets the application configure which Bluetooth options should be set. These options would be union'd along with the options passed along to AKSettings.setSession().

aure:
The state of affairs for me currently is that I can get AudioKit outputing to Bluetooth no problem, but the input is still not working.   When I ask AVAudioSession for the list of available inputs, it doesn't even show the bluetooth microphone.  Apple's music memos app also doesn't use the bluetooth input, but facetime does appear to.  Anyway have any ideas?

aure:
[...]

PR #801, opened 2017-03-17
Allow more precise control of audio session bluetooth options

pcbeard:
This addresses the comment I made in issue #675.


aure:
I think you're right about the manner of this solution, but I am not a big fan of the name "useBluetoothOptions" because that implies that the var is a Boolean.  I'd just change it to "bluetoothOptions". I can do that after accepting the pull request or you can make the edit if you agree.  Thanks!

pcbeard:
OK, I made the change.

pcbeard:
This should fix the travis build, since AVAudioSessionCategoryOptions isn't available on macOS.

aure:
Cool!

PR #802, opened 2017-03-17
Enable adding custom Sporth ugens written in Swift

jconst:
This PR adds a wrapper around Sporth's `plumber_ftmap_add_function()` functionality that calls through to a Swift class called `AKCustomUgen`. This class contains a compute function that the AudioKit user writes, which pops and pushes data to/from the Sporth stack, allowing the user to write arbitrary Sporth ugens in Swift. See the "Sporth Custom Effect" playground page I added as an example.

This functionality is particularly useful to me, as I'm working on an audio effects processing app, and I have all of my effects written in Sporth, but I would like to add more functions in addition to what Sporth currently ships with.

The `customUgens` init parameter works correctly on both AKOperationEffect and AKOperationGenerator. But as a side note, it feels to me like these classes should be combined, because they're like 90% the same code and I don't see a big reason for them to be separate.


aure:
Wow, this looks really great.  Just give me a little bit to fully grok this and then I'll pull it in, even if its not quite passing Travis, which I can probably fix easily.

PaulBatchelor:
Oh snap!

aure:
I'm having some trouble getting the playground running:

Playground execution failed: error: /var/folders/5d/zkr3lw7x71vd60mckbvrnm_w0000gn/T/./lldb/99429/playground72.swift:3:8: error: no such module 'Sporth_Custom_Effect_PageSources'
import Sporth_Custom_Effect_PageSources

aure:
I never was able to import AudioKit from within a playground page Sources swift file, and that seems to be an issue here again.  I wonder why it is working for you.  Are you on Xcode beta?  Maybe they've finally fixed this.

jconst:
Ah, that happened to me a couple times and I assumed it was just some quirk with Xcode for me. I'm on 8.2.1, not on beta, but I could always get past it by (as dumb as this sounds), opening `throttleUgen.swift`, cutting out the `import AudioKit` line, trying to run the playground, then pasting it back in the same spot and running it again. (I read this somewhere on Stack Overflow.)

As for a permanent fix, I'll see if I can come up with something – but if the worst case scenario is just leaving out the playground page to merge in later when they finally fix Xcode, that would be fine with me.

jconst:
Oh, and you know what, I forgot to actually organize the new files the same way they're organized in the Xcode project, so I'll do that (right now a lot of them are just in the AKOperationEffect folder). I'm not sure why Travis is unable to find AKSporthStack.h though.

PaulBatchelor:
Apologies, this is a bit of a tangent... 

It's really awesome and surprising to see AudioKit people using the Sporth API. I'm very happy about it. I don't have *any* documentation for ``plumber_ftmap_add_function()`` besides using it twice in polysporth. @jconst how did you figure this one out?

jconst:
Glad you're interested in these changes! It wasn't actually that hard, I was just poking around in your examples and found [this](https://github.com/PaulBatchelor/Sporth/blob/master/examples/api/user_function.c), which I followed to get most of this set up. The only thing that I'm surprised I found (almost accidentally) was `plumber_recompile_string_v2`, which turned out to be crucial for getting this to work with AKOperationGenerator.

PaulBatchelor:
Amazing. I'm glad you could make the code work for you. Let me know if have any other sporth-related questions or problems :)

aure:
I just got the playground working, but yeah, we'll have to think of something more consistent than that StackOverflow hack, clever as it is. I'm going to accept the pull request because the playground sounds really awesome and I'm eager to play with this code. I may have some small changes like capitalization to UGen or even UnitGenerator, depending on how we all feel about the name.

aure:
I don't mind including a UGens section in AudioKit with the throttle.swift file and others that we come up with.  They'll be part of the framework, so the playgrounds will see them.  I think they can be really useful and great examples anyway.

aure:
Just FYI the only thing preventing Travis from building was including the new files in the projects for macOS and tvOS.  Easy thing to forget to do or never do if you're just doing iOS development.

PR #805, opened 2017-03-18
Update ViewController.swift

AdrianBinDC:
Updated `setupUI()` method to make it more compact.


aure:
That is better, but doesn't compile.  Can you make the change on the develop branch and make the pull request off of that?  If it were on the develop branch I would have accepted the PR and made the changes myself.

PR #811, opened 2017-03-22
Move notification handling code to main thread/queue

pcbeard:
While debugging I noticed these notifications weren't arriving on the main thread. This probably prevented the audio engine from being restarted correctly.


aure:
Thanks!

PR #818, opened 2017-03-27
Fix #817 

### LINKED ISSUE #817: AKAudioPlayer completionHandler gets called before the audio is played
Hi

**Background/Setup**

We are recording audio from a device's microphone and storing it in a file (via AVAudioRecorder) while playing the realtime buffered waveform through AudioKit's EZAudioPlot and AKNodeOutputPlot. For playback, we are playing back a recorded m4a from a file using AKAudioPlayer with the same visualization of whatever is being output from the device's speaker for visual feedback of the audio.

**Issue**

AKAudioPlayer's completionHandler gets called as soon as the play() function is called without any audio output. This happens only after AVAudioRecorder is used to record an audio. Until AVAudioRecorder is used AKAudioPlayer's completionHandler works fine as expected. Issue persists for all the audio files until the app is relaunched.

I am unable to figure out reason why this is happening. Please help me if I am missing anything here or this is a known issue. I am happy to provide the sample code if needed.

PS: This is only happening on devices iOS 10.2 and above.

pavan309:
audioInputEnabled is always true if AKMicrophone or AKStereoInput is initialized.

Now setting it back to false in deinit

### DIFF
--- a/AudioKit/Common/Nodes/Input/AKMicrophone.swift
+++ b/AudioKit/Common/Nodes/Input/AKMicrophone.swift
@@ -47,6 +47,10 @@ open class AKMicrophone: AKNode, AKToggleable {
             }
         #endif
     }
+    
+    deinit {
+        AKSettings.audioInputEnabled = false
+    }
 
     /// Function to start, play, or activate the node, all do the same thing
     open func start() {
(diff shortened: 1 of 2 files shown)


aure:
Nice!

PR #832, opened 2017-04-01
Set aftertouch msg length

SebWeb3:
MIDI aftertouch msg is 2 bytes long


aure:
What's the point of this pull request?  It doesn't have any new code in it.  Is it just a request to make an AudioKit release soon?

Issue #807, opened 2017-03-20
Audio sync / timing / looping is wrong


trevor-sonic:

trevor-sonic:
(accidentally pressed enter with empty post. so I closed the issue)

trevor-sonic:
I'm trying to do a metronome with audio file for  accurate timing.
in the code below I have a metronome sound perfectly sliced in DAW for 122BPM and song is 122BPM and it is perfectly start from 1st beat.

When I play them as the code below it lose sync after 15 sec and later it is more noticeable.
If I use the 1 beat of the 1 bar metronome by .endTime properties the timing issue is drastically noticeable.

As metronome  was slower/behind the song; I tweak the bpm to 123.03 but it always lose the sync.

I could achieve to perfect sync without AudioKit. By using audio node + pcm buffer etc. and it's super accurate till the end of the song 4 minutes! 

I want to use AudioKit for improve my code but how to improve this? I can test and improve AK.
I can provide metronome + song samples if necessary.

    let file1 = try AKAudioFile(readFileName: "122bpm.wav", baseDir: .resources)
    let file2 = try AKAudioFile(readFileName: "one_life.mp3", baseDir: .resources)

    var player1 = try AKAudioPlayer(file: file1)
    var player2 = try AKAudioPlayer(file: file2)

    // uncomment line below for dividing 1 bar metronome sound to 1 beat click
    //player1.endTime = Double (60.0 / 123.03)
    player1.looping = true

    var mixer = AKMixer(player1, player2)

    AudioKit.output = mixer
    AudioKit.start()

aure:
Yeah, if you can share the project, I'll be able to look at this some more.  It seems like there's some time lost in the looping of AVAudioPlayer (under AKAudioPlayer).

trevor-sonic:
I simplified for you to test. I work on playground. It's enough to create a new play ground page and paste this code. In the zip file there are 2 files metronome and song . Drag them into resources folder. I hope that helps to find the calculation issue.   

[song122bpm_metronome122bpm.zip](https://github.com/audiokit/AudioKit/files/856584/song122bpm_metronome122bpm.zip)



     // Metronome sync test
    import PlaygroundSupport
    import AudioKit

    let file1 = try AKAudioFile(readFileName: "122bpm.wav", baseDir: .resources)
    let file2 = try AKAudioFile(readFileName: "one_life.mp3", baseDir: .resources)

    var player1 = try AKAudioPlayer(file: file1)
[...]

PR #833, opened 2017-04-01
Fix AKAudioPlayer Looping Drift

sweetmandm:
## Why?
With the previous looping logic in `AKAudioPlayer`, the playback time would drift because of overhead incurred by the interface between AKAudioPlayer and AVAudioPlayerNode. This can be resolved by deferring to `AVAudioPlayerNode`'s internal looping mechanism.

This is in response to https://github.com/audiokit/AudioKit/issues/807 and incorporates @aure's suggested fix while maintaining the completion handler behavior.

## What Changed?
* Use AVAudioPlayerNode's internal looping mechanism instead of applying the buffer looping in AKAudioPlayer.
* `internalCompletionHandler` is only used for when the audio track actually completes playback.
* When the user toggles the `looping` boolean and the track is playing, the player schedules its next buffer with the updated looping logic at the next loop time.
* When `looping` is set to `false`, the loop will stop and the callback will fire when the current loop completes.

## Caveats
* I'm not quite sure what will happen if the user rapidly/frequently changes the value of `looping`, since each time they do it will trigger a call to `scheduleBuffer`. Though presumably the _previous_ call would have its buffer 'interrupted' and so only the last would win.
* This seems like it adds too much code (looking at the `AKScheduledAction` class). I think it may be better to disallow the user to toggle looping on/off after the player is created. The problem of course is that would be an API-breaking change.

## Tests
I didn't write any :P if the general idea represented here is OK'd I'll try to fill in some tests for this behavior.

## Playground
This playground can be used to demonstrate the fix, and demonstrate the `looping` toggle.
[fix-akaudioplayer-loop.playground.zip](https://github.com/audiokit/AudioKit/files/887737/fix-akaudioplayer-loop.playground.zip)

### DIFF
--- a/AudioKit/Common/Internals/AKScheduledAction.swift
+++ b/AudioKit/Common/Internals/AKScheduledAction.swift
@@ -8,32 +8,34 @@
 
 import Foundation
 
-class AKScheduledAction
-{
+class AKScheduledAction {
+
     private var interval: TimeInterval
     private var block: (() -> Void)
     private var timer: Timer?
-    
+
     init(interval: TimeInterval, block: @escaping () -> Void) {
         self.interval = interval
         self.block = block
         start()
     }
-    
+
     dynamic func start() {
         timer?.invalidate()
         timer = nil
-        timer = Timer.scheduledTimer(withTimeInterval: interval, repeats: false) { [weak self] (timer) in
-            guard timer.isValid else { return }
-            self?.block()
-        }
+        timer = Timer.scheduledTimer(withTimeInterval: interval,
+                                     repeats: false,
+                                     block: { [weak self] (timer) in
+                                        guard timer.isValid else { return }
+                                        self?.block()
+        })
     }
-    
+
     dynamic func stop() {
         timer?.invalidate()
         timer = nil
     }
-    
+
     deinit {
         stop()
     }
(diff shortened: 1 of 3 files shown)


sweetmandm:
thanks hound :D sorry, wish i'd seen the lint rules before submitting this, I'll clean this up

aure:
You don't have to worry about the hounds unless its easy.  This looks great.

aure:
So this doesn't pass Travis tests because one of the commands you use is 10.0 or higher whereas AudioKit needs to work iOS9+.  It looks like a fairly minor thing.

sweetmandm:
Ah somehow I totally missed that, should be fixed in 22c264d

aure:
Okay, so the iOS side of things passes, so everything is probably fine.  What someone (doesn't have to be you) has to do is to add your new files to the macOS and tvOS project files.   I'll do it tonight if you're uncomfortable with it.  It's not a big deal.

sweetmandm:
Cool I can get to it tomorrow as well, sorry to submit a busted build, been cramming this into those spare moments :P

aure:
I propagated the change to all the project files and Travis is passing again.  Thanks so much for this awesome pull request.  Got your eyes set on any other outstanding issues?  I appreciate the help!

sweetmandm:
Thanks for the merge! And for a really great piece of oss. I haven't looked into any of the other issues too deeply yet but I'm happy to keep looking for issues to pick off as I hack around with this project and get more acquainted with it.

PR #837, opened 2017-04-04
Fixed memory leaks in AKSequencer and AKMusicTrack

JSMilton:
Found a couple of memory leaks in AKSequencer and AKMusicTrack, so thought I'd do a PR for it. Never done anything for open source before so sorry if this is the wrong way to go about it! Thanks.


aure:
Thank you very much!

PR #840, opened 2017-04-05
Adding rate functionality for MusicPlayer based sequencers

JSMilton:


aure:
This change breaks the SamplerDemo project.  Looks like the SamplerDemo example could be updated relatively easily to use setRate instead, but I kind of like use properties if possible.

JSMilton:
Ah ok, I'll update to use the property setter. The reason I didn't was because there are other properties, such as tempo and length, that have a separate method for the setter so I thought that was the style you guys were going for.

aure:
If that's the style, you can leave it as you did it and update the example project.  I guess we have evolved a bit since then but haven't updated this code.

aure:
If you wanted to make everything property setters, I'm not going to complain, but that would be a separate pull request I guess.

JSMilton:
I'll probably just update the Demo for now, then I'll make a separate PR if I get round to refactoring the property setters. Cheers!

Issue #856, opened 2017-04-20
AKNodeOutputPlot plotType buffer is not working as expected from 3.5.4


pavan309:
Hi

Background/Setup

We are recording audio from a device's microphone and storing it in a file (via AVAudioRecorder) while playing the realtime buffered waveform through AudioKit's EZAudioPlot and AKNodeOutputPlot. For playback, we are playing back a recorded m4a from a file using AKAudioPlayer with the same visualization of whatever is being output from the device's speaker for visual feedback of the audio.

Issue

AKNodeOutputPlot graph is not being plotted either when recording or when playing back. Looks like the issue is only with plotType buffer, rolling plotType seems working fine. 

I am unable to figure out the way to debug this issue. Is there a way where I can run AudioKit code instead of from framework. Please let me know if I am missing anything.

Thanks

pavan309:
Do you think the reason I am seeing this issue because of the condition that was added in this [commit](https://github.com/audiokit/AudioKit/commit/86e760038528ca86b4978c8e0961e2f2759b17a6) in the method `displayLinkNeedsDisplay`?

aure:
Yeah,  I think you're right that that is the issue and should be reversed.

pavan309:
Hi Aure,

I have raised PR which I think it will fix. Do let me know if that works.
Is there a way where I can debug AudioKit code along with the my code?

Thanks.

aure:
Merged your fix.

PR #858, opened 2017-04-21
Possible fix for issue #856 

pavan309:
AKNodeOutputPlot graph is not being plotred either when recording or when playing back for buffer plot


aure:
Hi there, I will be accepting this pull request shortly, but at the same time I'll add a fix for playgrounds.

PR #867, opened 2017-04-28
Fix export error

terenzeyuen:


aure:
It is fixed on the develop branch, which is where pull requests should be based off.  It's also fixed in a better way there.

terenzeyuen:
Thanks @aure, any idea when it will be merged to master as recorder seems to need that fix to work.

PR #870, opened 2017-04-30
Split AKTuningTable into extensions, move into Microtonality subdirectory

marcussatellite:
Sorry for the ugly pull request...accidentally included some UI files and it took some reverts to redo correctly.

Note that this will break the playground because I renamed the indian scales.  I need some help setting up the playgrounds so I can easily test while modifying AudioKit (so I don't have to build_frameworks all the time).


aure:
This looks good.  I think I'll accept the pull request and then make Travis-CI happy.  Need to add three slash comments eventually too.

PR #879, opened 2017-05-06
AKMIDI+SendingMIDI.swift func openOutput() correction

trevor-sonic:
Destination name can be empty ( e.g. "") So optional string is better for here. 
To get all endpoints; we should map them without condition

`

    public func openOutput(_ namedOutput: String? = nil) {
                guard let tempPort = MIDIOutputPort(client: client, name: outputPortName) else {
                    return
                }
                outputPort = tempPort

        // Corrected version
        // Destination name can be empty ( like "") So optional string is better for here. 
        // To get all endpoints; we should map them without condition
        if namedOutput == nil {
            _ = zip(destinationNames, MIDIDestinations()).map {
                    endpoints[$0] = $1
            }
        }else{
            _ = zip(destinationNames, MIDIDestinations()).first { name, _ in
                namedOutput! == name
                }.map {
                    endpoints[$0] = $1
            }
        }

       //      Original version was below
       //        _ = zip(destinationNames, MIDIDestinations()).first { name, _ in
       //            namedOutput.isEmpty || namedOutput == name
       //        }.map {
       //          endpoints[$0] = $1
       //        }
        
      }
`


aure:
You're making a pull request off of the wrong branch.  Can you submit it off of develop?  It should only be one file changed, not 121!

trevor-sonic:
I'm sorry Aure, I messed up with my pull request, I didn't use this before 🤗 
Now, I did it correctly with another pull request.

PR #880, opened 2017-05-06
openOutput() correction: Select one with given name or all

trevor-sonic:
// Destination name can be empty ( like "") So optional string is better for here. 
    public func openOutput(_ namedOutput: String? = nil) {
            guard let tempPort = MIDIOutputPort(client: client, name: outputPortName) else {
                return
            }
            outputPort = tempPort

    // To get all endpoints; and set in endpoints array (mapping without condition)
    if namedOutput == nil {
        _ = zip(destinationNames, MIDIDestinations()).map {
                endpoints[$0] = $1
        }
    }else{
    // To get only  endpoint with name provided in namedOutput (conditional mapping )
        _ = zip(destinationNames, MIDIDestinations()).first { name, _ in
            namedOutput! == name
            }.map {
                endpoints[$0] = $1
        }
    }    
    }


trevor-sonic:
I guess this is a wrong branch, I'll change...

PR #883, opened 2017-05-08
Add AKFMOscillatorBankMonophonic and AKOscillatorBankMonophonic

warpling:
* Add `AKOscillatorBankMonophonic`/`AKFMOscillatorBankMonophonic` classes: All the great note-playing benefits of AKOscillatorBanks without the polyphony; plus methods to play a note with a velocity and duration.
* Add ability to connect multiple inputs at a time to `AKMixer` after initialization

There's definitely a bit of code reuse here I'd like to DRY if anyone has suggestions! I'm new to Swift but a protocol or extension didn't seem quite right? Probably best to add this in such a way that the other kinds of oscillator banks can easily inherit the behavior…

giphy-1

### DIFF
--- a/AudioKit/Common/Internals/AudioKit.h
+++ b/AudioKit/Common/Internals/AudioKit.h
@@ -79,6 +79,7 @@ FOUNDATION_EXPORT const unsigned char AudioKitVersionString[];
 
 // Generators
 #import "AKOperationGeneratorAudioUnit.h"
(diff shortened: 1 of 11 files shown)


aure:
This has some bad stuff in it, all the AKInputDevice crap that I was experimenting with, so I can't accept the PR, sorry!

PR #887, opened 2017-05-09
EndpointInfo structure created for having more information about the EndPoint

trevor-sonic:
Hi,

The new file I've added is for having more information about Endpoint.
Existing base structure was giving only `names` of the endpoints. 
Like;

    for n in midi.inputNames{
        print("Name: \(n)")
    }

I've added a similar structure and keep the exist (for backward compatibility)
Now we have;

    for info in midi.inputInfos{
            print("Display name: \(info.displayName)")
            print("Manufacturer: \(info.manufacturer)")
            // name, driverOwner, model, image
        }

More info properties can be added. I've added 6 most important ones.

### DIFF
--- a/AudioKit/Common/MIDI/AKMIDIEndpointInfo.swift
+++ b/AudioKit/Common/MIDI/AKMIDIEndpointInfo.swift
@@ -1,0 +1,56 @@
+//
+//  AKMIDIEndpointInfo.swift
+//  AudioKit For iOS
+//
+//  Created by dejaWorks on 06/05/2017.
+//  Copyright © 2017 AudioKit. All rights reserved.
+//
+
+import UIKit
+
+public struct EndpointInfo {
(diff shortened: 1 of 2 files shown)


aure:
This is good.  I am not so worried about backwards compatibility as I am about getting it right, but I'll accept the PR and see what else can be done.  The only thing is you haven't added the file to any .xcodeproj files (for iOS and macOS) so the file wouldn't show up as part of AudioKit.  You can either do that and add to this PR or I'll just accept and fix it myself, later tonight.

trevor-sonic:
Hi Aure,

I've added the project file in my repository and tried to add existing PR but I don't know if it is successful - anything I can do, just tell me I'll check again

aure:
Yeah, good enough, I'll add it to macOS now.  thanks!

PR #888, opened 2017-05-09
Fix tiny bug in new AKMicrophoneTrackerEngine

warpling:
Also removes implicit start call in initialization.


warpling:
Wrong branch. 😬

PR #889, opened 2017-05-09
Fix AKMicrophoneTrackerEngine stop method and remove implicit start on init

warpling:
@aure is god 😭

### DIFF
--- a/AudioKit/Common/Internals/Microphone Tracker/AKMicrophoneTrackerEngine.m
+++ b/AudioKit/Common/Internals/Microphone Tracker/AKMicrophoneTrackerEngine.m
@@ -30,7 +30,6 @@
         sp_ptrack_init(sp, ptrack, 512, 20);
         
         ezmic = [EZMicrophone microphoneWithDelegate:self];
-        [self start];
     }
     return self;
 }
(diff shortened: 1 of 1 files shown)


aure:
I am a mere summoner of gods like @syedhali and @PaulBatchelor

aure:
What was so bad about the implicit start on init?  I had mixed emotions but I put there because I figured that there wasn't much reason to have a mic tracker that is stopped.  99% of the time the next line would be to start the thing.  But perhaps you see it form a race condition or something.  Like I said, I had mixed emotions anyway, so I'm fine with it as you've changed it just curious.

warpling:
I think the thought was that it's a convenience that can also be dangerous. I'm not sure about this particular case but an implicit call to start could, trigger something like a mic permission prompt before the developer prefers to ask simply as unintended the side-effect of object initialized.

If we're moving towards hot-disconnecting/swapping where multiple graph objects may be kept on hand but not active I think it makes sense to make enabling/disabling them more explicit. It's definitely worth a larger discussion if we want to move things towards being more declarative which I think would prevent some pitfalls and not affect the learning curve of the library if we kept the great plethora of examples current :)

syedhali:
@warpling your logic makes sense for not explicitly starting the mic. Optionally, you could provide a constructor that automatically starts the mic (if people end up wanting that behavior).

Issue #886, opened 2017-05-09
MIDI Pitch Wheel handler is not catching the event. (bug)


trevor-sonic:
Hello,
I'm testing the MIDI example code given in MidiMonitor project with my hardware midi connection.
All listener are working well (note on, off, controls, after touch) but `receivedMIDIPitchWheel` I see in the debug panel that framework detect pitch bend changes but it doesn't trigger the handler in the test code. (I tested break point also, no luck)

    ...
    var midi = AKMIDI()
    override func viewDidLoad() {
        super.viewDidLoad()
        midi.openInput()
        midi.addListener(self)
    }
    ...
    // **Only this handler doesn't work in the same context that other handler works**
    func receivedMIDIPitchWheel(_ pitchWheelValue: MIDIByte, channel: MIDIChannel) {
        updateText("Channel: \(channel + 1)  midiPitchWheel: \(pitchWheelValue)")
    }
    ...

aure:
Judging from the quality of you work and your Pull Requests I bet you can track this bug down and fix it yourself!  But, I will look for it to when I get the chance

aure:
Trevor, pitch wheel works just fine.  It took me a while to figure out what was wrong with your code.  You have the wrong type, so the method signature is wrong, it should be:     
`func receivedMIDIPitchWheel(_ pitchWheelValue: MIDIWord, channel: MIDIChannel) {`

trevor-sonic:
Hi Aure, I'ms sorry I couldn't see this different `MIDIByte` - > `MIDIWord` I was secretly feeling that pith bender has to have bigger value than a byte but in the example (MidiMonitor) it was `MIDIByte` so I thought something else is occurring. Thank you for your time, I'll return the favour by adding some codes in AK.

PR #892, opened 2017-05-11
Minor Correction MidiMonitor Example to handle Pitch Wheel (issue #886)

trevor-sonic:
(issue #886) In the example handler `func pitchWheelValue` wasn't calling by handler because of type mismatch. 
I've just updated the code according to  Aurelius'  solution.


aure:
Thanks for finding this!

Issue #902, opened 2017-05-18
Cannot create a mono file with AKAudioFile.init(createFileFromFloats ...


MathieuLescure:
In AKAudioFile+ConvenienceInitializers.swift:

With the following constructor, an error is thrown (by `try self.write(from: buffer)`) if floatsArrays contains only one array.

```
    public convenience init(createFileFromFloats floatsArrays: [[Float]],
                            baseDir: BaseDirectory = .temp,
                            name: String = "") throws {
        
        let channelCount = floatsArrays.count
        var fixedSettings = AKSettings.audioFormat.settings
        
        fixedSettings[AVNumberOfChannelsKey] = channelCount
        
        try self.init(writeIn: baseDir, name: name)
```

Replacing 
`try self.init(writeIn: baseDir, name: name)` 
by 
`try self.init(writeIn: baseDir, name: name, settings: fixedSettings)` 
fix the problem.

aure:
Can you make a pull request for this fix?

MathieuLescure:
No problem, I will.

PR #904, opened 2017-05-19
Fix an error when creating a mono AKAudioFile from floats

MathieuLescure:
Related issue: https://github.com/audiokit/AudioKit/issues/902


aure:
Thanks!

PR #915, opened 2017-05-24
Fix: cannot create an AKAudioFile with space in name via init(writeIn…

MathieuLescure:
Creating an AKAudioFile with `AKAudioFile(writeIn: , name: , settings: )` fail if the name contains characters which are not allowed in URLs (like spaces).

Use the URL initializer `URL(fileURLWithPath: )` instead of `URL(string: )` to fix the issue.


aure:
Thanks Matheiu!

PR #916, opened 2017-05-25
Fix: AKNodeRecorder.reset() fail to delete the recorder file

MathieuLescure:
From what I understand, all URL objects used in AKAudioFile are now built with URL(fileURLWithPath:

So URL.path should be used instead of URL.absoluteString to get a string without the "file://" prefix.

This pull request fix only the reset() function in AKNodeRecorder. But the same issue seems to occur in AKAudioFile.exportAsynchronously. As I'm not using this function, I'm not confident enough to make the changes.


aure:
Thanks!

PR #920, opened 2017-05-26
Fix: AKTimePitch CPU usage stay high when bypassed

MathieuLescure:
AVAudioUnitTimePitch still use the CPU when the pitch is set to 0 and the rate to 1. But the CPU usage is 0 when the audio unit is bypassed by setting its bypass property to true.

CPU Usage values on iPad Air with (overlap: 32, pitch: 2400, rate: 32):
- started: ~29%
- bypassed by setting rate and pitch to default value: ~10%
- bypassed with AVAudioUnitTimePitch.bypass = true: ~1%

The same is true for other nodes based on Apple audio units. But AVAudioUnitTimePitch is one of the more CPU intensive and so its more visible.


aure:
Great find! I guess we should make this change for the other units as well.

PR #928, opened 2017-06-07
Offline render feature

davebang:
Implement offline render node and custom file player to support offline render node. There are still a couple of issues that need fixing.


aure:
I'm going to close this for now as it looks like a new a pull request is imminent and this one now is riddled with conflicts anyway.  I think it jumpstarted the process though.

PR #932, opened 2017-06-14
fix playground text ‘makensure’ to ‘make sure’

swiftcodex:


aure:
makesnsense to me!  thanks!

PR #936, opened 2017-06-25
add setting allowing developer to handle AVAudioSessionRouteChange no…

Greenwell:
Adds a setting allowing developers to handle the AVAudioSessionRouteChange notification independently, instead of having AudioKit handle this and restart the audio engine any time a route change occurs.

This is critical for apps that re-build the audio chain during runtime, because they will need to re-build the audio chain in case of a route change as well.


aure:
Thanks!  I wonder if we shouldn't have "enableRouteChangeHandling" or something like that.  I don't like the double negative of !disabled.

Greenwell:
Good point! I've submitted another pull request addressing that.

PR #943, opened 2017-07-03
Make AKMicrophoneTrackerEngine SoundPipe parameters on par with AKFrequencyTracker

warpling:
`AKMicrophoneTrackerEngine` was having difficulty tracking low notes (E2 on a guitar is what lead to finding this bug). This change brings `AKMicrophoneTrackerEngine`'s values on par with `AKFrequencyTracker` to give it a larger window to detect lower frequencies ("Smaller windows will give better time precision, but worse frequency accuracy (esp. in low fundamentals)." [according to psound documentation](https://groups.google.com/d/msg/audiokit/3ZPNv1vY9ZY/TTHL1MRLBwAJ)).


aure:
Thanks!

PR #945, opened 2017-07-05
Fix iOS 11 crash when AKAudioPlayer stops internal

aure:
player from its own completion handler.


aure:
I'm going to accept this because I think it doesn't do any harm, but we'll need to have to give iOS11 features a full implementation.

PR #952, opened 2017-07-10
AKMIDITransformer and playgrounds

EricGeorge:
Added AKMIDITransformer, which is a class that takes an array of AKMIDIEvents and can transform them.  Multiple AKMIDITransformers can be chained together to create complex MIDI transformations.  Also added 2 playgrounds that demonstrate use of AKMIDITransformer:  MIDI Scale Quantizer and MIDI Chord Generator.


aure:
Looks good to me!

PR #955, opened 2017-07-14
Refactor of start() for readability and bug fixes.

warpling:
Refactoring AudioKit's start method for readability and stability:
* `AVAudioEngineConfigurationChange` is now monitored regardless of AKSettings. Previously it was only monitored if `defaultToSpeaker` was set to `true`.
* Added new `allowAirPlay` option to `AKSettings` (defaults to false).
* Added AKLog warning when `allowAirplay` or any bluetooth options are set to `true` on iOS 9 where they are not supported
* Added an explicit check for tvOS (it previous was an`else` case)
* Fixed a bug where setting `bluetoothOptions` caused `.mixWithOthers` to not be set.
* Renamed `audioEngineConfigurationChange` to `restartEngineAfterConfigurationChange` to be in parallel with the existing and similar `restartEngineAfterRouteChange` method name.

### DIFF
--- a/AudioKit/Common/Internals/AKSettings.swift
+++ b/AudioKit/Common/Internals/AKSettings.swift
@@ -76,6 +76,9 @@
     /// Additional control over the options to use for bluetooth
     open static var bluetoothOptions: AVAudioSessionCategoryOptions = []
 #endif
+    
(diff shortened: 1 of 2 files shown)


aure:
Nice job taming the hounds too!

PR #956, opened 2017-07-16
Fixing crash from route/category changes when app is not active 

warpling:
Apps that do not support [background audio](https://developer.apple.com/library/content/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html) (UIBackgroundModes array in info.plist) and attempt to start AVAudioEngine while the app is in, or entering, the background crash (error code `561015905`, more humanly readable as `AVAudioSessionErrorCodeCannotStartPlaying`).

These changes add a convenience constant to `AKSettings` indicating if the app supports background audio; if it does then route/category changes that happen while the app is not active will allow AudioKit to attempt to (re)start the AVAudioEngine. Apps that do not will do nothing. This does not prevent explicit calls to `AudioKit.start()` when the app is not active.

This may not be a perfect or all inclusive solution. Open to suggestions, ideas, and concerns! 
It'd be worth having a larger discussion around whether these same protections should be added to `start` and other places.

### DIFF
--- a/AudioKit/Common/Internals/AKNotifications.swift
+++ b/AudioKit/Common/Internals/AKNotifications.swift
@@ -17,4 +17,14 @@ extension NSNotification.Name {
     ///
     public static let AKEngineRestartedAfterRouteChange =
       NSNotification.Name(rawValue: "io.audiokit.enginerestartedafterroutechange")
+
+    /// After the audio engine configuration is changed, (change in input or output hardware's channel count or
+    /// sample rate, for example) AudioKit restarts, and engineRestartAfterCategoryChange is sent.
+    ///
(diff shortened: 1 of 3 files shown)


aure:
The only issue with this pull request is that you're a bit to iOS-centric.  Need to do some OS checking around some of the additions, basically opening the for macOS project up and seeing what UI references it is seeing.  I can fix this too if you're done for the day.

warpling:
Very good point. I can add the iOS checks but do you think we should investigate a similar problem exists for macOS? I don't want to introduce something that makes the platforms too uneven.

PR #966, opened 2017-07-30
AKMixer fix

ryanfrancesconi:
There was a bug in AKMixer which was incorrectly specifying which bus to use for incoming inputs. I think the comments in the updated code are clear, but essentially before it was always connecting on mixerAU.numberOfInputs. The problem I found is that after around 40ish connections the audio going through this mixer would start to break up in a very unsettling way. Since the AVAudioMixerNode starts its life with 8 inputs, the starting point should be mixerAU.nextAvailableInputBus, which starts with 0. I've also changed the AKMixer to allow specifying which bus exactly to use. Kind of a power user feature, but it's something that I need.

I've tested now with over 100 changed connections to the mixer and it seems alright now.
Will keep thinking about optimizations for this class!


aure:
This looks cool, and I will definitely merge once it passes Travis.

PR #968, opened 2017-08-01
Additions for AKAudioUnitManager + Example macOS project

ryanfrancesconi:
Should be self-explanatory. Let me know if I screwed something up!
Ryan


aure:
Awesome!

PR #972, opened 2017-08-03
Feature/audioplayer seek fix

juusokos:
Improves AKAudioPlayers seeking abilities:
- Player is now stopped before updating buffer to remove glitchy behaviour of sound playing for some seconds when seeking an already playing track.
- Playhead no longer shifts by X amount of seconds when using play play() or play(from startTime: Double, to endTime: Double) methods
- internalCompletionHandler handler is now actually only called when the track plays to end and not when seeking
- Remove a check in startTime setter to allow a fairly normal use case of seeking from track -> replacing file -> seeking from same spot


aure:
Thanks!

PR #973, opened 2017-08-03
Reset engine connections sample rate and channel count when replacing…

juusokos:
Reset the engine connections sample rate and channel count when replacing file in AkAudioPlayer (in case these are not the same).


aure:
Seems good to me!

PR #984, opened 2017-08-10
Made AKSettings.BufferLength samplesCount and duration public.

dave234:
Seems like they should be accessible outside of the framework.


aure:
Since Xcode9 beta 5 breaks AVFoundation, Our integration testing is broken, but this looks like a fine change.

PR #988, opened 2017-08-10
Song processor update

dave234:
Playing all files in separate players to demonstrate AKAudioFile sync with play(at:)
Gave iTunes exporter a completion callback to make loading asynchronous and handle failures easier.
Moved song loading logic into song {didSet{}}
Various UI fixes to handle asynchronous loading.


aure:
I rarely get updates to example projects, so this is a real treat, thank you so much!

PR #989, opened 2017-08-11
Offlinerender

dave234:
OfflineRenderNode is an inline renderer, it can render without stopping the audioEngine.  Some nodes might have issues, AKVariableDelay is one.  AVAudioPlayerNode can schedule buffers, in offline mode, but not files.  Scheduling files may work sometimes, but it's totally unreliable.  I'm going to try to make a function that will traverse the upstream nodes, looking for known offenders so I can print a warning or even throw an error.

Updated the songProcessor example with offline render export, should use a device to test.


aure:
Thank. You.!

PR #993, opened 2017-08-12
Ak node io connections

dave234:
README - Significant change  :)

This is the riskiest bit in preparation for clean disconnect and making connections post initialization. 

Added an extension to AudioKit with static connect functions mirroring AVAudioEngine's with the intent of convenience and safety. 
 An AVAudioNode will be automatically attached if connecting using these.

In AKNode, connectionPoints are now public and calculated from AVAudioEngine rather than storing an instance variable that could get out of sync with AVAudioEngine.
The biggest difference implementation-wise is that when appending to connectionPoints, an output connection is made automatically with AudioKit.format
Now that AKMixer is using nextAvailableInput instead of connectionPoints.count for input bus this hopefully won't be an issue for anyone.
Refactored AKBalancer, AK3DPanner, and AKMixer around use of connectionPoints.

Added disconnect convenience wrappers to AKNode.
    
Deprecated AKNode.disconnect() as the naming is misleading, renamed it to detach.
Deprecated AKNode.disconnect(nodes:) as the naming is misleading and function is unrelated to self, warning suggests using new AudioKit.detach(nodes: [AVAudioNode]) convenience function instead.

I ran the Examples using AKBalancer and AK3DPanner without issue.

### DIFF
--- a/AudioKit/Common/Internals/AudioKit.swift
+++ b/AudioKit/Common/Internals/AudioKit.swift
@@ -485,3 +485,31 @@ extension AVAudioEngine {
         #endif
     }
 }
+
+
+//This extension makes connect calls shorter, and safer by attaching nodes if not already attached.
+extension AudioKit {
+    private static func safeAttach(_ nodes: [AVAudioNode]){
+        _ = nodes.filter{ $0.engine == nil }.map { engine.attach($0) }
+    }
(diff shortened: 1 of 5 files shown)


aure:
Thanks very much,  keep 'em coming.

PR #997, opened 2017-08-13
change button color when highlighted

Greenwell:
Changes the color of AKButton when it is highlighted (=tapped down). If highlightedColor is set, this color will be used, otherwise the regular color will be used with an alpha value of 0.6.


aure:
Ask and you shall receive, sometimes! Thanks!

PR #999, opened 2017-08-13
AKConnection changes

dave234:
Changed default connection behavior from SET to ADD.
    
Started refactoring nodes using new connections (with SET behavior) and things were breaking.  Looked into to it and found that throughout the codebase connections were generally added, not set.  This makes sense.  I've refactored the connection behavior so that connect(to...) now ADDS to follow the existing convention.  In order to support SET behavior I created a group of setOutput functions where you can explicitly set the output (will break existing connections).


dave234:
Sorry about all the formatting stuff!  How do I get the warnings before making a PR?

PR #1001, opened 2017-08-14
Implement AKInput

dave234:
Implemented AKInput for all nodes.

Most of the nodes just needed the protocol added.  The commit with 53 files doesn't need much review.  For the rest (12-13 nodes), it would be good took give them a once over.
For all of the nodes, I gave init a default value for the input node parameter.  Now the nodes can be initialized with init() instead of init(nill).
Nodes with more than one internal node require AKProtocol.inputNode to be implemented within the class, I added it where necessary, returning the avAudioNode of the first node in the chain.  
[Some changes](https://github.com/dave234/AudioKit/commit/4621a705210d5f570a66dd64a56ebd2480245a36) were necessary to properly expose the input node with a non-optional value on some of the nodes, or where the input was split internally.  In short, I added an inputMixer in front of everything to handle this.


aure:
Phenomenal update!

Issue #969, opened 2017-08-02
Crash receiving a phone call while using AKMicrophone


stonehouse:
I currently use AudioKit for listening to the microphone and applying FFT analysis. I've discovered that the AVAudioEngine causes a crash if it is started while a phone call is active. If you receive a phone call and then go to an app that is using AudioKit/microphone then the app will crash in the AudioKit.restartEngineAfterRouteChange(_) method when it tries to call self.engine.start().

The only solution I'm aware of for this problem is to check whether a phone call is active before attempting to prepare or start the engine.

I've submitted a radar for this, but I think AudioKit should check if the microphone is in use, and if so, check for active phone calls.

aure:
There have been updates that @warpling implemented for Blackbox, but perhaps those are still in the develop branch.

warpling:
I'm pretty sure this had to do with the phone call triggering the app resigning and an engine interruption which AK hears and then attempts to start the `AVAudioSession` again; it fails when the microphone is active because it's trying to start `AVAudioSession` in a category that it's not allowed to be in (`PlayAndRecord` or `Record`) since the app is not in the foreground and likely doesn't have background audio entitlements. 

This PR stops AK from attempting to restart if it doesn't have the correct entitlement to restart in the category it wants: https://github.com/AudioKit/AudioKit/pull/956

Still in `develop` branch.

stonehouse:
I was able to reproduce the crash with AVFoundation in a separate sample app (which I submitted to Apple in the radar).

```
let engine = AVAudioEngine()
guard let inputNode = engine.inputNode else {
    return
}
			
let bus = 0
inputNode.installTap(onBus: bus, bufferSize: 2048, format: inputNode.inputFormat(forBus: bus), block: { (_,_) in		
})
engine.prepare()
```

If you start a phone call, and run an app with this code, the app will crash immediately.

warpling:
Thank you, personally, for filling that!
Let me know if you get a chance to test the develop branch to see if it fixes the issue too.

aure:
I don't see any actions I can really take on this issue, so I will close it, but it would be great if you still posted updates here about your feedback from Apple on the radar you filed.

stonehouse:
[...]

PR #1005, opened 2017-08-16
Added lock on microphone while a phone call is in progress

stonehouse:
I've primarily created this PR as a proof of concept for a fix to the crash documented in #969 where AudioKit causes a crash if a phone call is in progress.

### DIFF
--- a/AudioKit/Common/Internals/AudioKit.swift
+++ b/AudioKit/Common/Internals/AudioKit.swift
@@ -9,6 +9,9 @@
 #if !os(tvOS)
 import CoreAudioKit
 #endif
+#if os(iOS)
+import CoreTelephony
+#endif
 
 import Dispatch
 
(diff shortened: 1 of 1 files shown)


aure:
Thanks for the follow through

aure:
Oops, you did wan this merged correct?

stonehouse:
I'm happy for it to be merged. I wasn't totally sure if you'd think it was a good idea for this kind of logic to be there. I thought you might want to have a bit of a discussion, but if you think it looks good then no worries.

aure:
Yeah, it looks like a good idea to me.

aure:
@warpling Might be interested in this too.

stonehouse:
I'm also happy to push check style fixes if you want.

warpling:
While checking `CTCallCenter` as a condition for locking might work for the mentioned use case I'm uncomfortable with this solution as it seems to overlook a more general solutions that could fix related crashes. It's unlikely the conflict is related purely to phone calls; have you looked into other audio routing callbacks to see if there's a more general way to cover this case?

stonehouse:
I'm definitely still looking for a better solution. Any suggestions? As far as I can tell, it seems like it relates specifically to trying to access the microphone while a phone call is in progress. But I'll try to test a bit more with the routing changes.

warpling:
Yes that's what I've seen as well 🤔 I won't have a chance to mess with it for a bit but have you tried polling for available input devices after receiving a restart or route change notification?

PR #1006, opened 2017-08-16
AKMusicTrack music track init bug fix.

trevor-sonic:
This bug could be available when I tried to use 2 sequencer and setup the tracks. My app was working on simulator but not on device. Device was breaking with BAD ACCESS code=1. To able to produce the bug, and find the problem I've activated Zombi objects and Address sanitizer. There are still some work to do for this function (AKMusicTrack->init) but the fix that I've done has improved and corrected the bug.

screen shot 2017-08-16 at 23 27 17


aure:
Looks like you committed some files that you didn't want to, referencing a change to the xcodeproj file for the sequencer demo.  I can fix this.

trevor-sonic:
I don't know what to do against "travis-ci"

aure:
Travis CI is broken as of Xcode beta 5 I suppose we could get off the beta for our testing, but I thought I'd try to wait it out and hope for a better beta 6.

megastep:
We could also just turn off macOS tests until Beta 6 comes out.

PR #1011, opened 2017-08-24
Pulled render notify out of AKLazyTap and into a base class.

dave234:
AKRenderTap is a base class that will set a render notify on an audio unit.
Instead of other classes having to set a render notify using the C interface, they can just subclass AKRenderTap and implement  renderNotifyBlock.
Implementing renderNotifyBlock is similar to how AUAudioUnit subclasses implement internalRenderBlock.

AKLazyTap inherits from AKRenderTap.
Added AKRenderTap and AKLazyTap to AudioKit.h.


aure:
Love getting pull requests from you. Definitely looking forward to examples/playgrounds that show these new things off.

PR #1013, opened 2017-08-24
AKRenderTapAKLazyTap3rdTry

dave234:
Third time is the charm!  Sorry about all the churn here, I think this one will stick.
AKRenderTap is now initialized with it's render block, which simplified the class greatly.
AKLazyTap now has an AKRenderTap as an ivar rather than inheriting from it.


aure:
I will want to see this new stuff in action when I am back from my vacation!

PR #1019, opened 2017-08-29
AKTimeline

dave234:
Converts AudioTimeStamp from system audioTime to a timestamp that represents
 the progression of a time in the context of a timeline.  Can be synchronized with great accuracy.


aure:
No idea what this does but you're the man. :)

larryseyer:
Yay!!!!!

PR #1026, opened 2017-09-01
AKTimelineTap

dave234:
AKTimelineTap is an audio tap on the render thread, based on AKRenderTap.  AKSamplerMetronome is a class that utilizes it, It triggers audio with sample accuracy and can be synchronized perfectly with other audio sources like AKAudioPlayer.  There's a new example project, MetronomeSamplerSync, that shows how to use AKSamplerMetronome, but AKTimelineTap is the real advancement.

Also, there's a bug fix for AKButton in there, totally related.


aure:
Exciting!

PR #1034, opened 2017-09-05
Add option to lazilly buffer AKAudioPlayer's AKAudioFile.

warpling:
`AKAudioPlayer` can take a significant amount of time creating buffers for it's internal player when initialized. If immediate playback isn't necessary for your application optionally deferring buffering may save significant time when creating large graphs.

Looking for feedback on naming and any problems you all might see with this! Considering adding an public function for optionally ensuring the player is buffered before `start` but after `init` if more convenient.


larryseyer:
I, for one, need this today!  Thank you for suggesting this!

aure:
If two people need it, then it must be good.

PR #1037, opened 2017-09-08
AkClipPlayer

dave234:
A group of classes that utilize AVAudioPlayer to read multiple clips of audio from one or many audio files in a sequence.  Streams from file so memory footprint remains low for large files.  Sample accurate scheduling.


aure:
Another awesome addition to AudioKit, thank you!

PR #1039, opened 2017-09-09
Alternate to ak unit defines

jandyman:
Proof of concept of AU audio base class. Not urgent, for review. Has been tested with separate project, but using the same code in a Playground causes errors:

```
        let oscillator = AKOscillator()
        let gainer = AKBooster2(oscillator)
        AudioKit.output = gainer
        AudioKit.start()
        oscillator.start()
        sleep(4)
        oscillator.stop()
```


aure:
@jandyman On Xcode9 the playground makes a sound but the gainer.gain has no effect

aure:
I've made the necessary fixes so this should pass tests, which will come right after I accept this pull request.

PR #1042, opened 2017-09-12
Fixes for AKNodeOutputPlot and HelloWorld examples

megastep:
I took a closer look at why the plots were no longer working in the HelloWorld examples, and basically this was because the `AKNodeOutputPlot` class didn't really connect itself to a tap if a node wasn't given to it explicitly, which never happened when loading the plot from a storyboard.

Instead a reasonable default is to set it up with the default global output since that's what this plot seems to be here for.

The other step to make this work was to move the AK init stuff in the examples out of the `viewDidLoad` methods to the basic init, so they get called earlier in the lifecycle before the plot got instantiated - allowing the default to work as expected. It just makes more sense to move non-UI code to this kind of init anyway.

I'm not sure at what point this plot class started deviating from the others as they seem to have different types of setup, but in any case that fixes it for now. I haven't checked macOS either yet.

### DIFF
--- a/AudioKit/Common/User Interface/AKNodeOutputPlot.swift
+++ b/AudioKit/Common/User Interface/AKNodeOutputPlot.swift
@@ -53,7 +53,7 @@ open class AKNodeOutputPlot: EZAudioPlot {
     ///
     required public init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
-        setupNode(nil)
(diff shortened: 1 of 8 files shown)


megastep:
Well I did check the macOS example and it was indeed still working, however I made a simple change so that the code is closer to the other examples.

aure:
Good changes!

PR #1043, opened 2017-09-12
Swift 4 fixes to example projects for AudioUnitManager

ryanfrancesconi:
I think they're working right now.


aure:
Weird that Travis says it failed, seems to have actually passed!

ryanfrancesconi:
it definitely compiled and ran ok for me!

PR #1048, opened 2017-09-15
syntax fix for dataSources in inputDevices

dorjeozer:
dataSources can be nil, so it needs to be checked before access.


aure:
Thanks very much!  Always nicer to have a pull request than an issue! :)

PR #1058, opened 2017-09-22
Documentation Returns -> Return :/

dave234:
Documentatio, correction, Returns -> Return :/


aure:
I would have thought it was returns too.  I rarely use that field.

PR #1062, opened 2017-09-24
Added some missing @objc 

jldunk:
Added some missing @objc


aure:
It looks like you added versions of the files at the top level instead of changing the files in place. Would you like me to me move the files into the correct place?

jldunk:
Sure aure, first time doing this!

PR #1069, opened 2017-09-28
Built in a workaround for empty mixer on running engine bug.

dave234:
If an AVAudioMixerNode's output connection is made while engine is running, and
there are no input connections on the mixer, subsequent connections made to
the mixer will silently fail.  A workaround is to connect a dummy node to the mixer
prior to making a connection, then removing the dummy node after the connection
has been made.


ryanfrancesconi:
did you have time to file a bug with apple?

dave234:
Not yet.

dave234:
They might say that the empty mixer not rendering is a feature, but I think the mixer not accommodating new connections when one method is used vs another can’t be called anything but a bug.

Issue #1065, opened 2017-09-26
ObjectiveC bridge problems


emrahgunduz:
We have an old objectivec project that is using AudioKit. After the 4.0 update, we started having problems when calling Swift methods from objc part of the project.

`No visible @interface for 'AKMixer' declares the selector 'detach'`
AKMixer does not override this method and AKNode version is not accessible. Adding another `@objc` for the method at AKNode level solved the problem.

`No visible @interface for 'AKMixer' declares the selector 'connect:'`
Old connect is deprecated in favor of `connect:to:bus:`, yet Xcode does not mark connect as deprecated. Also, new method is unreachable from objc. Added two methods at AKMixer level which seems to solve the problem for me right now.
```swift
  @objc open func connect(input: AKNode?, bus: Int) {
    input?.connect(to: self, bus: bus)
  }
  
  @objc open func connect(input: AKNode?) {
    input?.connect(to: self, bus: nextInput.bus)
  }
```

`No visible @interface for 'AKAudioPlayer' declares the selector 'replaceWithFile:error:'`
Method is not reachable.

`No visible @interface for 'AKAudioPlayer' declares the selector 'playFrom:to:'`
Method is not reachable.

`Error:(338, 37) property 'duration' not found on object of type 'AKAudioFile *'`
I know this is defined at AVAudioFile extension. Code completion can see duration, but on compile this error appears. Adding same duration definition to AKAudioFile class solved the problem.

More errors appear as soon as I solve another...
`Property 'avAudioNode' not found on object of type 'AKMixer *'`

`No visible @interface for 'AKAudioPlayer' declares the selector 'initWithFile:looping:error:completionHandler:'`
It looks like a new init is written which includes a block alias `AKCallback`, which is not accessible from objc. No deprecated error for old method. New method does not like void blocks and is not reachable.

I had to return back to last version 3 of AudioKit. Looks like new Swift4 version is not really compatible with objectivec.

I tried these also: Deleting derived data folder does not solve any of these. Errors start to appear again on first build. Tried deleting all framework files and re-compiling, did not work. Same errors are appearing either using via Cocoapods or adding AudioKit project + framework by hand.

*PS: These are the ones we are currently seeing. Do not know if there are more in other classes as we are only using just a small part of the whole package.*

aure:
Did you get the 4.0.1 update? It addressed a lot of the objecitve-C issues.

emrahgunduz:
[...]

PR #1071, opened 2017-09-29
Solving some objc problems

emrahgunduz:
I'm sending three commits to your way.

Commit f644871 solves the problem for AKAudioFile's most variables not being accessible from ObjectiveC calls.

Commit 9e4b340 adds two connect methods on Swift side for AKMixer which makes them methods accessible from ObjectiveC. As it is not a class, extension does not accept @objc addition. Another method is for detach, which is not overriden before. It simply calls super's detach method.

Commit aa470ac adds @objc to AKAudioPlayer's replace:file and play:from:to methods. It looks like, some other methods too are missing @objc, but I did not want to mess with them.

I've run the travis tests, seems no errors were found. Framework is also building without any errors.

These additions solved [my problems](https://github.com/AudioKit/AudioKit/issues/1065).


aure:
Thanks!

PR #1077, opened 2017-10-02
Updates to Audio Unit Manager

ryanfrancesconi:
I've updated the Audio Unit manager to correctly load AU's that are midi controlled. 
kAudioUnitType_MusicEffect
vs
kAudioUnitType_Effect

These were missing before. I've unfortunately had to make a code breaking change and change the format of the internal AU array to be of a more generic type "AVAudioUnit" from "AVAudioUnitEffect". The docs say the midi AU's are "AVAudioUnitEffect" but it doesn't work.

I've also made the init that allows for custom insert lengths.


aure:
Awesome work Ryan!

PR #1091, opened 2017-10-17
Add support for a clip playback completion handler.

hhanesand:
Add an optional `AKCallback` onto the `FileClip` protocol, which allows users to perform an action when a clip finishes playing.

This was almost already supported, since the private `scheduleFile` takes an optional completion handler, but there was no way for the client to pass a completion block in.

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
@@ -44,7 +44,7 @@ extension AKClip {
 /// A file based AKClip, with a completion handler.
 @objc public protocol FileClip: AKClip {
     var audioFile: AKAudioFile { get }
-    var completionCallback: AKCallback?
+    var completionCallback: AKCallback? { get }
 }
 
 /// A FileClip implementation, used by AKClipPlayer.
--- a/AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
@@ -41,7 +41,7 @@ extension AKClip {
     }
 }
 
-/// A file based AKClip
+/// A file based AKClip, with a completion handler.
 @objc public protocol FileClip: AKClip {
     var audioFile: AKAudioFile { get }
     var completionCallback: AKCallback?
--- a/AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
+++ b/AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
@@ -44,6 +44,7 @@ extension AKClip {
 /// A file based AKClip
 @objc public protocol FileClip: AKClip {
     var audioFile: AKAudioFile { get }
+    var completionCallback: AKCallback?
 }
 
 /// A FileClip implementation, used by AKClipPlayer.
@@ -61,6 +62,9 @@ open class AKFileClip: NSObject, FileClip {
     /// The duration of playback.
     open var duration: Double
 
+    // The callback executed when the end of the clip is reached.
+    open var completionCallback: AKCallback?
+
     /// Create a new file clip.
     ///
     /// - Parameters:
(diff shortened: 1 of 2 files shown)


dave234:
NIce!  You could just name it completion instead of completionCallback though.

AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
```diff
@@ -41,9 +41,10 @@ extension AKClip {
     }
 }
 
-/// A file based AKClip
+/// A file based AKClip, with a completion handler.
 @objc public protocol FileClip: AKClip {
     var audioFile: AKAudioFile { get }
+    var completionCallback: AKCallback? { get }
```
dave234: Since this is already an objc protocol, I think it would be better to make this addition optional.
```
@objc optional 
var completion: AKCallback? { get }
```

AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClipPlayer.swift
```diff
@@ -105,15 +105,15 @@ open class AKClipPlayer: AKNode, AKTiming {
                                  time: 0,
                                  offset: clip.offset + diff,
                                  duration: clip.duration - diff,
-                                 completion: nil)
+                                 completion: clip.completionCallback)
```
dave234: If made optional just coalesce ```clip.completionCallback ?? nil```

AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClipPlayer.swift
```diff
@@ -105,15 +105,15 @@ open class AKClipPlayer: AKNode, AKTiming {
...
                 }
 
             } else {
                 scheduleFile(audioFile: clip.audioFile,
                              time: clip.time - offset,
                              offset: clip.offset,
                              duration: clip.duration,
-                             completion: nil)
+                             completion: clip.completionCallback)
```
dave234: If made optional just coalesce ```clip.completionCallback ?? nil```

AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClip.swift
```diff
@@ -61,23 +62,29 @@ open class AKFileClip: NSObject, FileClip {
     /// The duration of playback.
     open var duration: Double
 
+    // The callback executed when the end of the clip is reached.
+    open var completionCallback: AKCallback?
```
dave234: I think it should be called completion rather than completionCallback, this name along with the type makes it clear what it's purpose is.

hhanesand:
This isn't working as expected yet, I'm getting a callback before the clip finishes. Gotta go read docs...

> did recieve time playback 3.377000
> track with length 4.71793650793651 finished. this IS the longest track. looping true
> looping!

dave234:
Are you working on looping it?  If so, awesome!  If it helps, here's a AVAudioPlayerNode wrapper that loops 1 entire file by scheduling at least 2 loops out using the completion handler.    

https://github.com/dave234/AKFilePlayer

dave234:
It would be really cool to be able to set loopStart and loopEnd though.

hhanesand:
As it is right now, this is only useful to me on iOS11, because of the new callback APIs provided. This lets us do something like this :

```swift
let completion = completion.flatMap { (completion) in
    return { (_ type: AVAudioPlayerNodeCompletionCallbackType) in
        completion()
    }
}

playerNode.scheduleSegment(audioFile, startingFrame: offsetFrame, frameCount: frameCount, at: startTime, completionCallbackType: .dataPlayedBack, completionHandler: completion)
```

That leaves this PR kind of in limbo until iOS11 comes out... The callback was getting called early [because of this](https://stackoverflow.com/questions/29427253/completionhandler-of-avaudioplayernode-schedulefile-is-called-too-early).
@dave234 I don't see this addition being useful to anyone until we support only iOS11.  We could add an #available check, but that would result in the callback being called earlier on pre-ios11 versions....

dave234:
@hhanesand I think this change would still be useful.  Another user made a similar PR, but closed it after seeing this one.  We can always extend the function to match the iOS 11 implementation later.  Could you make the requested changes?  I'll merge it.

Issue #1025, opened 2017-08-31
AudioKit pauses background audio when started.


warpling:
Tested on `/develop` on 8/30/17.

To reproduce take any AudioKit example and run it while background audio is playing.

I've confirmed the session category is `.ambient` and the option `.mixWithOthers` is being used so something else must be afoot… suggestions welcome!

EssamSoft:
seem here 👍

aure:
I don't know if there's anything that can do be done with this issue.

warpling:
It's possible to fix. Based on AudioKit's default category, mixing, and routing settings it appears it should not be mixing and not pausing music when started, but it is. I'd like to explore this more when I can because it probably will surface something else that is not as it appears configuration wise and because my users are mentioning it.

aure:
You know where to find me.

aure:
I think one of the things I'd like to verify with issues like this is whether or not AudioKit is the culprit, or just AVAudioEngine. Does this happen with a simple AVAudioEngine project with those session categories? If so, then its more of an Apple bug. We'd still want whatever work around that is found to be integrated into AK. Really I have no idea where to look and I hate having issues that I can't act on.

dave234:
Do you have a control that you're using?  If another background app uses
mixwithothers, it works.

warpling:
What do you mean by "a control"?
Which Audio Kit sample app are you testing with? I'll try to use the same.

dave234:
By control I meant two apps where you can produce the desired result.  So
app a is in the background playing, then app b starts playback in the
foreground, result is - a keeps playing.  Then swap out app b for an
AudioKit app and test.

warpling:
Ah right! I'll work on this. When I first tested it I used a few AK based apps on the store and an AK sample app. I know when I used to use TAAE it wasn't an issue so I assume it's not AV unless it's a recent bug but it seems unlikely.

PR #1103, opened 2017-10-28
🎶 🛑 Let background audio continue to play when AudioKit is initialized

warpling:
As documented in #1025 AudioKit's initialization will pause background audio even when the `AKSettings` seem to suggest AudioKit will use a category like `.ambient` that should support mixing.

As far as I can tell the problem is that `AVAudioSession`'s category/options are not configured until AudioKit is started but setting `AVAudioEngine`'s output node causes the `AVAudioSession` to activate with its default category of `AVAudioSessionCategorySoloAmbient` [which pauses background/mixable audio](https://developer.apple.com/library/content/documentation/Audio/Conceptual/AudioSessionProgrammingGuide/AudioSessionCategoriesandModes/AudioSessionCategoriesandModes.html). This PR ensures `AVAudioSession`'s category/options are configured based on current `AKSettings` before the `AudioEngine`'s output is touched.

⚠️ This may be a naive fix as there may be other potential code paths that cause `AVAudioSession` to be activated. It might be prudent to move this new `updateSessionCategoryAndOptions()` method call to the `AVAudioEngine` extension method `connect()` but I'm not sure if this is overkill or problematic especially while the session is already active.

### DIFF
--- a/AudioKit/Common/Internals/AKSettings.swift
+++ b/AudioKit/Common/Internals/AKSettings.swift
@@ -230,6 +230,60 @@ extension AKSettings {
         }
     }
 
+    @objc open static func computedSessionCategory() -> SessionCategory {
+        if AKSettings.audioInputEnabled {
+            return .playAndRecord
+        } else if AKSettings.playbackWhileMuted {
+            return .playback
+        } else {
+            return .ambient
+        }
+    }
+
+    @objc open static func computedSessionOptions() -> AVAudioSessionCategoryOptions {
+
+        var options: AVAudioSessionCategoryOptions = [.mixWithOthers]
+
+        if AKSettings.audioInputEnabled {
+
+            options = options.union(.mixWithOthers)
+
+            if #available(iOS 10.0, *) {
+                // Blueooth Options
+                // .allowBluetooth can only be set with the categories .playAndRecord and .record
+                // .allowBluetoothA2DP comes for free if the category is .ambient, .soloAmbient, or
+                // .playback. This option is cleared if the category is .record, or .multiRoute. If this
+                // option and .allowBluetooth are set and a device supports Hands-Free Profile (HFP) and the
+                // Advanced Audio Distribution Profile (A2DP), the Hands-Free ports will be given a higher
+                // priority for routing.
+                if AKSettings.bluetoothOptions.isNotEmpty {
+                    options = options.union(AKSettings.bluetoothOptions)
+                } else if AKSettings.useBluetooth {
+                    // If bluetoothOptions aren't specified
+                    // but useBluetooth is then we will use these defaults
+                    options = options.union([.allowBluetooth,
+                                             .allowBluetoothA2DP])
+                }
+
+                // AirPlay
+                if AKSettings.allowAirPlay {
+                    options = options.union(.allowAirPlay)
+                }
+            } else if AKSettings.bluetoothOptions.isNotEmpty ||
+                AKSettings.useBluetooth ||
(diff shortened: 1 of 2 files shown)


aure:
Looks cool but I'll let some more eyeballs on it... basically if @dave234 and @megastep are cool with this I think its a go.

SwiftyFlow:
I noticed it happens because of the declaration of:
    let mic = AKMicrophone()

If not done the music stays, AudioKit lowers it a bit but it remains. 

However I loose my wave form because of the impossibility to call AKMicrophone() :( I can't afford that... Gotta display a message to the user telling that if the music is restarted after app starts it will stay. Not graceful...

Any idea?

warpling:
@SwiftyFlow, could you elaborate? Are you declaring an `AKMicrophone` in your code before starting AudioKit and saying that when you then start AudioKit, currently playing music pauses? Are you setting `AKSettings` to enable `audioInputEnabled`?

aure:
I'm not exactly sure why this pull request is where this discussion is being had.  Seems to me if its a current thing then it should be an issue.

warpling:
I think @SwiftyFlow possibly thought it was related to the changes, but you're right that it'd be worth making a new issue if it's unsolved.

SwiftyFlow:
Ok guys sorry if it wasn't clear.

I have an app with the main screen having in the top declarations:
//MARK: - Players and Recorder
    var wavFileName = ""
    var audioRecorder: AVAudioRecorder!
    var audioSettings = [String : Int]()
    var classicPlayer: AVAudioPlayer?
    var timerForClassicPlayer: Timer!
    var fullRecordingDuration: Double = 0.0
    
    //MARK: - Declarations Audio.io
    @IBOutlet private var waveForm: AKNodeOutputPlot!
    var micBooster: AKBooster!
    var moogLadder: AKMoogLadder!
    var mainMixer: AKMixer!
    let mic = AKMicrophone()     <------- that's the guy...
...

If I leave the let mic = AKMicrophone() it pauses the music (ie: Spotify) when the app starts

If I remove it, it keep playing thanks to later in the code:

let quality = LowSpecDevice ? AVAudioQuality.low.rawValue : AVAudioQuality.medium.rawValue
        self.audioSettings = [AVFormatIDKey: Int(kAudioFormatMPEG4AAC), AVSampleRateKey: 6000, AVNumberOfChannelsKey: 1, AVEncoderAudioQualityKey: quality]
AKSettings.bufferLength = .medium
        do {
            try AKSettings.setSession(category: .playAndRecord, with: .mixWithOthers)
        }catch {
            AKLog("Could not set session category.")
        }

AKSettings.defaultToSpeaker = true
        self.micBooster = AKBooster(AKMixer(self.mic)) ; self.micBooster.gain = 0
        self.mainMixer = AKMixer(self.moogLadder, self.micBooster)
        
        AudioKit.output = self.mainMixer
        AudioKit.start()
        
        self.waveForm.node = self.mic
...

But like this, as the self.mic has never been initialised, I loose my waveForm (AKNodeOutputPlot) visual

Arf, I'll leave the message for now as like I was saying if the user play again spotify after opening it remains so it's good enough for now. 

However, you guys are right it might not have it's place on this discussion. Tell me when you've read and I'll delete the comments to clean...

warpling:
@SwiftyFlow Please start this as an issue. Feel free to leave the comments for posterity unless @aure wants them removed.

aure:
Nah, no reason to delete what's here already, but repeat what you need to on the issue so people don't have to dig for info.

PR #1104, opened 2017-10-28
fix hard coded key for snare drum sequence track

WAJLemos:
replace 2 with Sequence.snareDrum.rawValue

looks like i picked up some of your commits on master... sorry


aure:
Thanks!

WAJLemos:
This is probably the wrong place to ask, but... the Drums example crashes for me. You?

PR #1114, opened 2017-11-09
Deprecated AKOfflineRenderNode in favor of AudioKit.renderToFile.

dave234:
AKOfflineRenderNode was not working post iOS 11, so it's been deprecated.  

Extended AVAudioEngine and AudioKit with renderToFile method to support offline rendering moving forward.


aure:
Cool. would be great to develop a playground for this.

sacred0x01:
So it's unable to render offline on devices prior iOS 11 now? Because I initially started to digging into you framework only because of offline render for iOS <= 10

PR #1146, opened 2017-12-07
AKPlayer 

ryanfrancesconi:
Mostly AKPlayer!
a few other small changes elsewhere too. If this checks out I would suggest deprecating AKAudioPlayer. AKPlayer will be much easier to maintain and has some fixes that people have requested in the past.


AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift
```diff
@@ -38,63 +38,10 @@ extension AKAudioFile {
...
-                    framePosition = position
-                    lastPeak = peak
-                }
-                position += block.count
-            }
-        }
-
-        let time = Double(framePosition / pcmBuffer.format.sampleRate)
-        return time
+    @available(*, deprecated, renamed: "AKPlayer.findPeak(pcmBuffer:)")
```
dave234: This seems backwards to me.  The peak finding should be at the file or buffer level, then the player should call into that.

ryanfrancesconi: I don't to use AKAudioFile any longer, but if you think it should stay around I'll move this back. This is my function btw.

ryanfrancesconi: Basically I was wondering if it would be better to have a utility class for these kinds of useful but static things.

dave234: If I were to implement it myself, I would probably do an extension on AVAudioPCMBuffer because that is common denominator.
```let floatData = pcmBuffer.floatChannelData```

Comment (dave234) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: It would look like this ```let peak = myBuffer.peak()```

Comment (dave234) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: Or let peak = myBuffer.floatPeak(), since buffers can contain non-float data.

Comment (ryanfrancesconi) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: yeah, i considered that. i haven't entirely liked the blurring of lines between extensions and foundation classes. I suppose it happens a lot in other places in AudioKit, so I think I'll take your advice. Where are these type of extensions going?

Comment (ryanfrancesconi) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: AudioKitHelpers?

Comment (dave234) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: AudioKit/AudioKit/Common/Internals/Utilities/AVAudioBufferConvenience.swift

Comment (ryanfrancesconi) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: AVAudioBufferConvenience.swift ?

Comment (ryanfrancesconi) [reply AudioKit/Common/Internals/Audio File/AKAudioFile+Utilities.swift]: went with peakTime as a property.

File: AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift [hunk 2]
```diff
@@ -0,0 +1,613 @@
...
+        }
+    }
+    
+    //MARK:- Private Parts
+    
+    // The underlying player node
+    private let playerNode = AVAudioPlayerNode()
+    private var mixer = AVAudioMixerNode()
+    private var buffer: AVAudioPCMBuffer?
+    private var startingFrame: AVAudioFramePosition = -1
```
dave234: Why are these -1?

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,613 @@
...
+    open var pan: Float {
+        get { return playerNode.pan }
+        set { playerNode.pan = newValue }
+    }
+    
+    /// get or set the current time of the player. If playing it will scrub to that point.
+    /// If buffered it will preroll.
+    public var startTime: Double = 0 {
+        didSet {
+            if startTime < 0 {
```
dave234: startTime = max(0, startTime)

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,613 @@
...
+            if isPlaying {
+                stop()
+                play()
+            }
+        }
+    }
+    
+    public var endTime: Double = 0 {
+        didSet {
+            if endTime > duration {
```
dave234: endTime = min(endTime,duration)

dave234:
If AKTiming is adopted, we can start making extensions for players and sequencers and anything else that we might want to synchronize.  Extensions like myPlayer.synchronizeWith(other: AKTiming) become possible.

AudioKit/Common/Internals/Utilities/AVAudioBufferConvenience.swift
```diff
@@ -69,4 +69,62 @@ extension AVAudioPCMBuffer {
...
+    // return the highest level in the given array
+    private func getPeak(from buffer: [Float]) -> Float {
+        // create variable with very small value to hold the peak value
+        var peak: Float = -10_000.0
+        
+        for i in 0 ..< buffer.count {
+            // store the absolute value of the sample
+            let absSample = abs(buffer[i])
+            
+            if absSample > peak {
```
dave234: peak = max(peak, absSample)

Any time you can remove an "if" and maintain readability.

ryanfrancesconi:
Ok i'll switch to that. I started with it, as you saw I borrowed some of your code, but then I wasn't sure if I wanted to make the time() audioTime() function visible.

AudioKit/Common/Internals/Utilities/AVAudioBufferConvenience.swift
ryanfrancesconi: will do

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
ryanfrancesconi: let updateNeeded = (buffer == nil || startFrame != startingFrame || endFrame != endingFrame || fade.needsUpdate)

dave234: This seems a little fragile.  I'm sure it works great now, but maintainers need to understand why it's -1.  While clunkier, a notSet Bool would be clearer, hopefully there's a more elegant solution.  In general, I try not to use special values in the place of flags.  An optional might work here too.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -162,21 +160,14 @@ public class AKPlayer: AKNode {
...
-            }
+            endTime = min(endTime, duration)
         }
     }
     
-    /// Current time of the player in seconds.
+    /// - Returns: Current time of the player in seconds.
+    /// Use setTime() or startTime to set the time
     open var currentTime: Double {
-        get {
```
dave234: Why remove the setter?

ryanfrancesconi: There's a lot of redundancy for startTime once AKTiming is added. Seems kind of weird to have setTime() startTime AND currentTime. What do you suggest?

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
ryanfrancesconi: optional sounds good to me.

AudioKit/Common/Internals/Utilities/AVAudioBufferConvenience.swift
```diff
@@ -69,4 +69,62 @@ extension AVAudioPCMBuffer {
         }
         return buffer
     }
+    
+    /// - Returns: The time in seconds of the peak of the buffer or 0 if it failed
+    open var peakTime: Double {
```
dave234: A function, rather than a property makes more sense here.  Calling a function is a hint to the user that there is work being done.  If it were a catchable value on an immutable object, a property would make sense.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
ryanfrancesconi: although it doesn't really matter what the value is as long as it doesn't start with a valid one. The goal is to prevent filling the buffer needlessly

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
dave234: I would just remove setTime.  What's startTime for?

AudioKit/Common/Internals/Utilities/AVAudioBufferConvenience.swift
ryanfrancesconi: yeehaw. keep that whip a crackin

ryanfrancesconi: hee yaw!

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
ryanfrancesconi: setTime is part of the AKTiming protocol. One of the reasons I chose not to originally use the protocol.
startTime and endTime allow you to set the in point / out point independent of playback. It's also used to store the last in/out point specified by the play(from: to:)

dave234: So it is.  That _is_ confusing.

dave234: So startTime is the same as setTime, correct?

ryanfrancesconi:
image

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+    }
+
+    deinit {
+        AKLog("* deinit AKPlayer")
+    }
+
+    // MARK: - Static Methods
+
+    /// convert an AVAudioTime object to seconds with a hostTime reference
+    open class func audioTimeToSeconds(hostTime: UInt64, audioTime: AVAudioTime) -> Double {
```
dave234: This should return an optional Double and return nil if !isHostTimeValid.
```if !audioTime.isHostTimeValid { return nil }```

Comment (dave234) [reply AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift]: Also, it should be moved from this class as it's unrelated to the player.  Add it as an extension to AVAudioTime in AudioKit/Common/Internals/Utilities/AVAudioTimeShim.swift.  Or make it private.  Another option, which you may prefer, is to create an AKUtility class and add it as a static function there.

File: AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift [hunk 9]
```diff
@@ -0,0 +1,724 @@
...
+        self.audioFile = audioFile
+        initialize()
+    }
+
+    public override init() {
+        super.init(avAudioNode: mixer, attach: false)
+    }
+
+    private func initialize() {
+        guard audioFile != nil else { return }
```
dave234: guard let audioFile = audioFile else { return }
Then get rid of the the forced unwrapping ```audioFile!```.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+    /// with a hostTime reference
+    public func play(when scheduledTime: Double, hostTime: UInt64?) {
+        play(from: startTime, to: endTime, when: scheduledTime, hostTime: hostTime)
+    }
+
+    public func play(from startingTime: Double,
+                     to endingTime: Double,
+                     when scheduledTime: Double,
+                     hostTime: UInt64? = nil) {
+        let refTime = hostTime != nil ? hostTime! : mach_absolute_time()
```
dave234: let refTime = hostTime ?? mach_absolute_time()
Removed forced unwrapping.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+    public func play(from startingTime: Double, to endingTime: Double, at audioTime: AVAudioTime?, hostTime: UInt64?) {
+        preroll(from: startingTime, to: endingTime)
+        schedule(at: audioTime)
+        playerNode.play()
+
+        completionTimer?.invalidate()
+        prerollTimer?.invalidate()
+
+        let prerollTime = audioTime != nil ?
+            AKPlayer.audioTimeToSeconds(hostTime: hostTime!, audioTime: audioTime!) : 0
```
dave234: Here's a crasher.
There's nothing here to prevent optional hostTime from being force unwrapped while nil.  
Make hostTime non-optional, or guard where it'll be used.
Forced unwrapping should be avoided as a rule.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+            let revStartTime = endTime > 0 ? duration - endTime : duration
+
+            startFrame = AVAudioFramePosition(revStartTime * fileFormat.sampleRate)
+            endFrame = AVAudioFramePosition(revEndTime * fileFormat.sampleRate)
+        }
+
+        let updateNeeded = (buffer == nil || startFrame != startingFrame || endFrame != endingFrame || fade.needsUpdate)
+
+        if !updateNeeded {
+            Swift.print("updateBuffer() no update needed.")
```
dave234: Is this expected?  If so remove print statement.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+            return
+        }
+        let frameCount = AVAudioFrameCount(endFrame - startFrame)
+        //let totalFrames = AVAudioFrameCount(audioFile.length)
+        buffer = AVAudioPCMBuffer(pcmFormat: processingFormat, frameCapacity: frameCount )
+
+        do {
+            audioFile.framePosition = startFrame
+            // read the requested frame count from the file
+            try audioFile.read(into: buffer!, frameCount: frameCount)
```
dave234: Optional bad!

guard let buffer = AVAudioPCMBuffer(pcmFormat: processingFormat, frameCapacity: frameCount ) else { return }

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+        if isFaded {
+            fadeBuffer()
+            fade.needsUpdate = false
+        }
+
+        // these are only stored to check if the buffer needs to be updated in subsequent fills
+        self.startingFrame = startFrame
+        self.endingFrame = endFrame
+
+        Swift.print("updateBuffer() read from \(startTime) to \(endTime)")
```
dave234: Remove print statements for expected behavior.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+    // Apply sample level fades to the internal buffer.
+    // TODO: add other fade curves or ditch this method in favor of Audio Unit based fading.
+    // That is appealing as it will work with file playback as well as buffer
+    private func fadeBuffer() {
+        if fade.inTime == 0 && fade.outTime == 0 {
+            return
+        }
+        AKLog("fadeBuffer() inTime: \(fade.inTime) outTime: \(fade.outTime)")
+
+        guard isBuffered, let buffer = self.buffer, let audioFile = self.audioFile else { return }
```
dave234: When there's more than two conditions, format them like this:
```
guard isBuffered,
    let buffer = self.buffer,
    let audioFile = self.audioFile else {
        return
}
```

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
```diff
@@ -0,0 +1,724 @@
...
+                } else {
+                    gain = 1.0
+                }
+
+                //sanity check
+                if gain > 1 {
+                    gain = 1
+                }
+
+                let sample = buffer.floatChannelData![n][i] * Float(gain)
```
dave234: Unwrapped optional again.
Maybe throw this in with your guard statement at the top.

```
guard isBuffered,
    let buffer = self.buffer,
    let audioFile = self.audioFile,
    let floatChannelData = buffer.floatChannelData else {
        return
}
```

dave234: Might as well through fadedBuffer in the guard statement too as it's integral to what the function does.

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
ryanfrancesconi: ok will do. I have wondered about this convention. I find the syntax kind of more strange to create a new variable to unwrap then just guarding to make sure the original one isn't nil. But if this is the best convention, changing!

ryanfrancesconi: is it in the case of if another thread could change audioFile ?

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
ryanfrancesconi: good catch!

AudioKit/Common/Nodes/Playback/Player/AKPlayer.swift
dave234: It is exactly that.

dave234: It's also just better to work with non optionals as a best practice.

PR #1154, opened 2017-12-10
Change AKTming

dave234:
This has a few breaking changes.

AKTiming adopters will have to rename a few functions. I don't believe anyone is adopting it at this point so probably no biggie.

AKClipRecoder.play was renamed to start. This might affect a few early adopters.

### DIFF
--- a/AudioKit/Common/Internals/Synchronization/AKTiming.swift
+++ b/AudioKit/Common/Internals/Synchronization/AKTiming.swift
@@ -12,13 +12,15 @@
     /// Starts playback at a specific time.
     /// - Parameter audioTime: A time in the audio render context.
     ///
-    func start(at audioTime: AVAudioTime?)
+    @objc func start(at audioTime: AVAudioTime?)
 
     /// Stops playback immediately.
-    func stop()
+    @objc func stop()
+
+    var isStarted: Bool { get }
 
     /// Set position in playback timeline (seconds).
-    func setPosition(_ position: Double)
+    @objc func setPosition(_ position: Double)
 
     /// Timeline time at an audio time
     /// - Parameter audioTime: A time in the audio render context.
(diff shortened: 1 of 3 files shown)


AudioKit/Common/Nodes/Playback/Player/ClipPlayer/AKClipPlayer.swift
```diff
@@ -216,8 +186,36 @@ open class AKClipPlayer: AKNode, AKTiming {
         set { playerNode.pan = newValue }
     }
 }
-extension AKClipPlayer {
+extension AKClipPlayer: AKTiming {
```
dave234: @ryanfrancesconi Moving the AKTiming functions to an extension makes it clear that the functions are there to satisfy AKTiming.  Maybe AKPlayer could do the same.

AudioKit/Common/Internals/Synchronization/AKTiming.swift
```diff
@@ -12,33 +12,93 @@
...
-    func play(at audioTime: AVAudioTime?)
+    @objc func start(at audioTime: AVAudioTime?)
 
     /// Stops playback immediately.
-    func stop()
+    @objc func stop()
 
-    /// Start playback immediately.
-    func play()
+    var isStarted: Bool { get }
```
dave234: @ryanfrancesconi I removed play() from AKTiming it's just a convenience for play(at:nil) I'm sorry about isStarted, it had to be done.

ryanfrancesconi: ok cool. I just committed an AKPlayer update, so I'm going to look at this now

dave234:
I think it’s alright.  The thing I like about setPosition and position(at audioTime) is that it’s not a convenience like var position set get, or currentTime set get.  This allows each implementor to have the conveniences as part of their class.  If someone is doing something AKTiming related, it should be with the intent to accurately schedule audio.  That was part of the reasoning behind removing the start() function.

aure:
I think it will be wise to target the next release as 4.1 with this change and the new AKPlayer, and possibly a deprecation marking if not removal of AKAudioPlayer.

ryanfrancesconi:
I would deprecate AKAudioPlayer as soon as some others have success with AKPlayer. I've been testing it in a few different scenarios, but there could be more?

ryanfrancesconi:
I will implement this AKTiming, then I think it's ok to merge AKPlayer

PR #1159, opened 2017-12-14
Cleaning up a few things

ryanfrancesconi:
updated AudioKit swxftlint yml to current syntax: unsure if desired but it didn't compile for me here.
syntax housekeeping, refactoring and safety on the audio unit manager and converter.

converted Dave's example RandomClips to use AKPlayer

### DIFF
--- a/Examples/macOS/RandomClips/RandomClips/ViewController.swift
+++ b/Examples/macOS/RandomClips/RandomClips/ViewController.swift
@@ -15,7 +15,7 @@ class ViewController: NSViewController {
     let drumPlayer = AKClipPlayer()
     let guitarPlayer = AKClipPlayer()
     let mixer = AKMixer()
-    var drumLooper: AKAudioPlayer?
+    var drumLooper: AKPlayer?
     let playButton = AKButton()
     let guitarDelay = AVAudioUnitDelay()
     let reverb = AKReverb()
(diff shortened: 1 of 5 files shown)


Examples/macOS/RandomClips/RandomClips/ViewController.swift
```diff
@@ -104,12 +129,12 @@ class ViewController: NSViewController {
 
                 drumPlayer.currentTime = 0
                 guitarPlayer.currentTime = 0
-
                 drumPlayer.prepare(withFrameCount: 44_100)
                 guitarPlayer.prepare(withFrameCount: 44_100)
 
-                drumLooper.schedule(from: 0, to: drumLooper.duration, avTime: nil)
-                guitarLooper.schedule(from: 0, to: guitarLooper.duration, avTime: nil)
+                drumLooper.play()
```
dave234: These will not be in sync if started this way.  Should be play(at: futureTime)

ryanfrancesconi: sorry typo... fixing

ryanfrancesconi:
damnit

Examples/macOS/RandomClips/RandomClips/ViewController.swift
dave234: Ok looks good as far as the example.  I didn't look at the rest.

PR #1171, opened 2017-12-19
Adjusted start button L&F 🆗

ryanfrancesconi:


aure:
pulled against master instead of develop

PR #1177, opened 2017-12-26
AudioUnitManager: iOS update

ryanfrancesconi:
works now. also cleaned up forced optionals. DropDown has some violations but I don't really feel like fixing those.


aure:
Thank you!

PR #1183, opened 2018-01-01
AkRenderTap fix

dave234:


aure:
If anyone knows this its you :)

Issue #1182, opened 2018-01-01
Defined arguments for AKMusicTrack.clearRange() do not match with implementation


derekleerock:
First of all - thank you for AudioKit and all of the great work you've done with it! I previously implemented some low-level audio in Obj-C and it's been great to use AudioKit on new projects to simplify these complexities. 

I've been experimenting with AudioKit using the AKSequencer with the goal of being able to modify the length of a looping sequence (both adding to it and removing from it in real time). As a result, in trying to use the AKMusicTrack.clearRange() method to remove MIDI notes from a track I've noticed what I believe to be a discrepancy. I'd like to kindly ask for your confirmation and if this is the case I'd be happy to send a PR with an adjustment based on your feedback. 

The clearRange() method currently has the following definition:

    /// Clear some events from the track
    ///
    /// - Parameters:
    ///   - start:    Start of the range to clear, in beats
    ///   - duration: Duration of the range to clear, in beats
    ///
    open func clearRange(start: AKDuration, duration: AKDuration) {

This is directly calling "MusicTrackClear(track, start.beats, duration.beats)":

https://github.com/AudioKit/AudioKit/blob/f99e45103f42b02f995df7271264828d9d3e1dc1/AudioKit/Common/MIDI/Sequencer/AKMusicTrack.swift#L366

However, "MusicTrackClear" expects both a start time as a music time stamp as well as an end time as a music time stamp:

```
/*!
	@function	MusicTrackClear
	@abstract	Removes all events within the specified range
	@discussion	All time ranges are [starttime < endtime]
	@param		inTrack		the track
	@param		inStartTime	the start time for the range of events
	@param		inEndTime	the end time up to which will form the range of the events to clear
*/
@available(iOS 5.0, *)
public func MusicTrackClear(_ inTrack: MusicTrack, _ inStartTime: MusicTimeStamp, _ inEndTime: MusicTimeStamp) -> OSStatus
```

Therefore, with the current implementation, passing in a start AKDuration of, for example, 4 (for beat # 4), and a duration of 4 (to clear out one 4/4 measure of time), this actually passes in a start time and end time of 4 for both, effectively not clearing out any events (and also generating an "Invalid beat range or track is empty" error). 

It's my assumption that this is likely not the desired functionality, therefore I can see two approaches for improving this:
1. Keep the method declaration for clearRange() as it is indicating the start time and duration - but modify the clearRange() method implementation to calculate the end time to pass into the MusicTrackClear() method. 
[...]

PR #1185, opened 2018-01-02
Adjust AKMusicTrack.clearRange() method to calculate range duration correctly

derekleerock:
Simple fix to the AKMusicTrack.clearRange() method to correctly calculate the range of notes to remove:
* Decided to stick with the existing API definition since this was also being utilized by the AKSequencer.clearRange() method. 
* The actual change itself was quite simple, but to have a bit more confidence I added some tests around this and the related methods for AKMusicTrack.
* Cleaned up some formatting to match the current style in the existing files since I was already in there.
* Updated the API documentation to standardize the descriptions for both AKSequencer as well as AKMusicTrack. 
* Note that clearRange() is inclusive of the start time, but *exclusive* of the end time.  This was surprising to me but is also noted in Apple's documentation:

```
	@function	MusicTrackClear
	@abstract	Removes all events within the specified range
	@discussion	All time ranges are [starttime < endtime]
```

### DIFF
--- a/AudioKit/Common/MIDI/Sequencer/AKMusicTrack.swift
+++ b/AudioKit/Common/MIDI/Sequencer/AKMusicTrack.swift
@@ -351,11 +351,11 @@ open class AKMusicTrack {
         return outBool
     }
 
-    /// Clear some events from the track
+    /// Clear all events from this track within the specified range
     ///
     /// - Parameters:
     ///   - start: Start of the range to clear, in beats (inclusive)
-    ///   - duration: Duration of the range to clear, in beats (exclusive)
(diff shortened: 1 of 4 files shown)


derekleerock:
Fix for issue #1182

aure:
You wrote tests!  Thanks for making my new year so far!

aure:
Thanks very much for this!

derekleerock:
My pleasure! I've been doing TDD for the last couple of years so for better or worse, I can't *not* write tests any longer. 😅 Need to be confident that the code I'm contributing is solid and that's the only way I know to feel comfortable. 👍🏼

Thanks again for all of the great work in AudioKit! Hope I'll have more opportunities to contribute in the future.

PR #1188, opened 2018-01-04
Allow AudioKit to throw exceptions rather than consume or fail on them

warpling:
In an attempt to be more Swifty and ultimately give users more control over audio routing related exceptions and errors this PR allows AudioKit to throw many of the exceptions or errors it previously consumed or crashed on using fatalExceptions. 

This likely isn't perfect and should be considered a breaking change so more eyes would be apprecaited! Thank you for the encouragement to do this, @dave234.

### DIFF
--- a/AudioKit/Common/Internals/AudioKit.swift
+++ b/AudioKit/Common/Internals/AudioKit.swift
@@ -278,11 +278,11 @@ public typealias AKCallback = () -> Void
         engine.stop()
         shouldBeRunning = false
 
+        #if os(iOS)
         notificationObservers.forEach { (observer) in
             NotificationCenter.default.removeObserver(observer)
         }
 
-        #if os(iOS)
         do {
             try AVAudioSession.sharedInstance().setActive(false)
         } catch {
--- a/AudioKit/Common/Internals/AudioKit.swift
+++ b/AudioKit/Common/Internals/AudioKit.swift
@@ -43,9 +43,13 @@ public typealias AKCallback = () -> Void
     /// An audio output operation that most applications will need to use last
     @objc open static var output: AKNode? {
         didSet {
-            updateSessionCategoryAndOptions()
-            output?.connect(to: finalMixer)
-            engine.connect(finalMixer.avAudioNode, to: engine.outputNode)
+            do {
+                try updateSessionCategoryAndOptions()
+                output?.connect(to: finalMixer)
+                engine.connect(finalMixer.avAudioNode, to: engine.outputNode)
+            } catch {
+                AKLog("Could not set output: \(error)")
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -352,82 +344,76 @@ public typealias AKCallback = () -> Void
...
+                    let appDoesNotSupportBackgroundAudio = !AKSettings.appSupportsBackgroundAudio
 
-                    // Sends notification after restarting the engine, so it is safe to resume AudioKit functions.
-                    if AKSettings.notificationsEnabled {
-                        NotificationCenter.default.post(
-                            name: .AKEngineRestartedAfterConfigurationChange,
-                            object: nil,
-                            userInfo: notification.userInfo)
+                    if appIsNotActive && appDoesNotSupportBackgroundAudio {
+                        AKLog("engine not restarted after configuration change since app was not active and does not support background audio")
```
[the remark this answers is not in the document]
warpling: Deal with it Hound.

warpling:
This branch is in progress and **should not be merged**. Would like to test and integrate further exception catching (demoed by Dave: [here](https://github.com/dave234/PlayerStartCatch))

aure:
To make this pass tests, we'd have to change the examples to do try AudioKit.start() I guess.

aure:
@warpling did you get any feedback on this on slack?

warpling:
I didn't see any feedback on the PR itself (afaik) just that we need to update tests to pass.

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -205,79 +212,82 @@ public typealias AKCallback = () -> Void
...
-        start()
+        try start()
     }
 
     /// Start up the audio engine
-    @objc open static func start() {
+    @objc open static func start() throws {
         if output == nil {
-            AKLog("AudioKit: No output node has been set yet, no processing will happen.")
+            AKLog("No output node has been set yet, no processing will happen.")
```
dave234: This should throw a custom error.  If a function can throw, than it should when failing.

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -40,12 +40,19 @@ public typealias AKCallback = () -> Void
...
+    /// Notification observers
+    fileprivate static var notificationObservers: [Any] = []
+
     /// An audio output operation that most applications will need to use last
     @objc open static var output: AKNode? {
         didSet {
-            updateSessionCategoryAndOptions()
-            output?.connect(to: finalMixer)
-            engine.connect(finalMixer.avAudioNode, to: engine.outputNode)
+            do {
```
dave234: Maybe this should be ```func setOutput(node: AKOutput) throws```?

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -205,79 +212,82 @@ public typealias AKCallback = () -> Void
...
         // Stop the engine.
         engine.stop()
         shouldBeRunning = false
+
         #if os(iOS)
+        notificationObservers.forEach { (observer) in
+            NotificationCenter.default.removeObserver(observer)
+        }
+
         do {
```
dave234: You could remove this do-catch, just let the error message bubble up to the user if not able to recover internally.

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -352,82 +355,76 @@ public typealias AKCallback = () -> Void
...
+                if AKSettings.notificationsEnabled {
+                    NotificationCenter.default.post(
+                        name: .AKEngineRestartedAfterConfigurationChange,
+                        object: nil,
+                        userInfo: notification.userInfo)
                 }
             }
+        } catch {
+            AKLog("error restarting engine after route change")
+            // Note: doesn't throw since this is called from a notification observer
```
dave234: This is a code smell.  You mentioned thrashing as a motivation for this refactor.  Notifications that throw that can potentially trigger notifications (that also throw) seems like a good place to look for the thrashing.  Refactor if you have any ideas for how to deal with this, otherwise add a TODO, but it's definitely stinky :)

dave234: Maybe there could be a callback that has a default value, engineWasInterrupted?  The default closure could restart the engine, but more serious apps would have an opportunity to control this behavior by setting a custom closure, then choosing whether or not to restart within the closure.

adamnemecek:
I'm not convinced this is a step in the right direction.

dave234:
@adamnemecek Why not?

adamnemecek:
I guess i'm not crazy about plastering tries all over the place. We could also add a Result<T> type for this. I think I'd prefer that.

dave234:
I understand the preference for result types (especially when Swift starts supporting result types), but generally speaking those should be at the application level rather than within a library.  For most users, the only change will be try? AudioKit.start() vs AudioKit.start().  Other than that there's updateSessionCategoryAndOptions which most users won't use, and AKSampler.loadInstrument, which users really should catch if there are loading errors.

warpling:
It's definitely a tough balance but as Dave mentioned I don't think this makes AK less beginner friendly and it does allow us to make less potentially harmful assumptions about how people are using the framework.

aure:
I'm going to bite the bullet and accept this PR, fixing the things it breaks.

PR #1192, opened 2018-01-06
Add ability to override AKRotaryKnob bubble text color

b3ll:
This adds the ability to override the text color of the value shown when dragging a knob to change the value. Whilst you can change the text color based on theme or via the `-textColor` property this doesn't always work when you do really custom styles to the knob (i.e. make the knob all dark and use the indicator dots as indication of the value changing).


aure:
THanks!

PR #1198, opened 2018-01-12
AKPlayer bug fix for looping players

ryanfrancesconi:
changing loop points marks buffer as needing an update.
removed one forced optional from AKSamplePlayer, just happened to notice it.
✌️


aure:
Thanks!

PR #1211, opened 2018-01-22
Add Chorus and Flanger effects, and a demo program

getdunne:
Demo is macOS only at the moment, but I see no reason why these effects shouldn't work on iOS and tvOS.


aure:
I'm not going to worry too much about these hound issues and accept the PR.

aure:
Things look really great and even if there are a few tweaks I'm super excited to have this in the repo!

PR #1212, opened 2018-01-23
Switched AudioKit over to static frameworks

megastep:
The main purpose of switching from the previous dynamic frameworks to static frameworks is to make them easier to use for our end users. We no longer have to worry about the dreaded `fix-framework.sh` script needing to be called in projects, and more of the necessary business is now automatically handled by the compiler for us.

Other advantages: 
- No need to independently sign the framework when building and submitting to the App Store, reducing the time it takes to process apps.
- The compiler is able to optimize out the unused parts of the library, drastically reducing the amount of linked code for most apps using AudioKit.

Changes from before:
- The `-lstdc++` linker flag needs to be explicitly added to targets because of the internal dependency on the C++ standard library.
- It is no longer required to embed the framework; as a matter of fact they should be removed from existing projects as they would just take room unnecessarily.

Resources that were previously bundled in the frameworks are no longer there, mostly out of necessity since the concept of an independent bundle doesn't make sense in a static library and they are no longer embedded. I have replaced the code accessing these resources with equivalents either recreating the files from code in a temporary directory (in the case of the STK modules), or parsing XML from memory (for the `.aupreset` file). *This code should be properly tested before we push out a release.*

I believe the simplified CocoaPod should also properly set up parent projects to link properly, but I haven't verified this yet.

I have updated all included Xcode projects to use the new static version of the library (including all examples). This mostly involved removing the embedded frameworks, the run phase calling the `fix-framework.sh` script and setting the additional linker flags.


aure:
This looks incredible!

dave234:
This is really cool.

dave234:
I'll make sure to test the preset.

PR #1219, opened 2018-01-27
Exponential Fader in AKPlayer and Booster

ryanfrancesconi:
Have added an internal booster to AKPlayer which handles fades. I wasn't satisfied with the Linear ramping so I wrote a new ramper. I only sort of know what I'm doing but it seems to be working ok.

Removed the sample based fading from AKPlayer. From the API point of view it isn't very different. I doubt anyone is fading yet anyway, but this allows AU based ones. A much better option.


aure:
I guess Andy might have some feedback for you, but for now, this looks good.

ryanfrancesconi:
Cool - yeah I'm hoping others might pitch in on this

PR #1220, opened 2018-01-27
AKPlayer: Changes to remove AKAudioFile properties

ryanfrancesconi:
Changes to remove AKAudioFile properties. These are throwing errors in test projects and aren't necessary. Also updated test suite for mac os to have broader range for booster params. A good place to hear the new exponential parameter ramping.


aure:
Keep 'em coming, thanks!

PR #1233, opened 2018-02-03
Add Documentation files for Chorus/Flanger

getdunne:
Note my ModDelay.svg diagram is included in the Documentation directory itself. You may want to move it to AudioKit.io/images, in which case you'll need to update the link in the .md file.

Keeping all related files together might make Git(Hub)-based management easier, however.


aure:
This is great. SVG is so small that I don't mind having that format in the repo.

PR #1234, opened 2018-02-09
Intercept Objective-C exceptions

megastep:
This adds a new `AKTry` utility method that encapsulates code within an Objective-C `@try` code to intercept and propagate NSExceptions issued from within AVFoundation to methods that currently throw exceptions in Swift. I've mostly restructured some of the basic AVFoundation setup to use this new `AKTry` for safety.

Note that while this builds and passes tests and apparently doesn't seem to break anything, I've not yet been able to test the specific use cases where this would be helpful as this mostly happens unprompted on some of my users' devices... However this could probably use a unit test or two.


AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
```diff
@@ -0,0 +1,39 @@
...
+public func AKTry(_ operation: @escaping (() throws -> Void),
+                  finally: (() -> Void)? = nil) throws
+{
+    var error: NSException?
+    
+    let theTry = {
+        do {
+            try operation()
+        } catch let ex {
+            error = ex as? NSException
```
dave234: If ex is a Swift error, rather than an exception, it will just be ignored here.  We need to figure out how to rethrow here, or do away with the do-try-catch here and make operation non-throwing.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
```diff
@@ -0,0 +1,39 @@
...
+
+import Foundation
+
+extension NSException: Error
+{
+    public var domain: String { return "io.audiokit.SwiftExceptionCatcher" }
+    public var code: Int { return 0 }
+}
+
+public func AKTry(_ operation: @escaping (() throws -> Void),
```
dave234: As it is now, operation should not take a throwing closure as any error caught will be ignored.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
dave234: I guess we could try to wrap the error in an NSError.  Re-throwing the original would be ideal though.

megastep: You're probably right about Swift errors/exceptions, however the main goal of this was to specifically intercept Objective-C exceptions, as these are the ones that go uncaught currently. I only use AKTry() around sections of Cocoa code that should only trigger Obj-C exceptions, though there are probably a few exceptions.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
megastep: I'm not sure I follow. There is a catch section that assigns the caught exception, which is in turn thrown back as a Swift exception. How is it getting lost?

AudioKit/Common/Internals/AKSettings.swift
```diff
@@ -222,7 +226,9 @@ extension AKSettings {
 
         // Activate session
         do {
-            try session.setActive(true)
+            try AKTry({
+                try session.setActive(true)
+            })
```
aure: Since this a trailing closure, can't you just do 

```
AKTry {
    try session.setActive(true)
}
```

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
dave234: ```error = ex as? NSException``` If ex is a Swift error (not a NSException),  error will be nil and no exception will be thrown.

AudioKit/Common/Internals/AKSettings.swift
megastep: I don't think so because I added a `finally` optional closure. Swift was balking at that. However I never end up using that anyway, so I could probably simplify it.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
megastep: OK, though I think the declaration above adds Swift Error protocol conformance to NSException. That's not my original code though, so I should probably make sure it works.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
dave234: I agree that the exception cases will be handled, but swallowing Swift errors might lead to troubles down the road.  Rethrowing would be best, but we should at least print them (if not NSException).  I couldn’t figure out a general solution, I worry we might need to handle these on a case by case basis.

megastep: Do we really have cases where both Swift and NSExceptions can be raised by the single call, i.e. methods that are marked `throws`? We could restrict this to calls that have been known to unexpectedly raise NSException, like `engine.prepare()` for instance. I would be OK with moving current throwing calls out of the AKTry closure if that covers all our bases.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
dave234: Maybe the error var should just be an Error? rather than NSException?, then you could just throw that.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
megastep: My focus has been on the current methods we have that throw Swift errors, and the goal was basically to intercept any NSException and rethrow them in the same way, so we don't get crashes.

dave234: AVAudioEngine.start() and some of the throwing AVAudioEngine funcs are suspected to throw exceptions.

AudioKit/Common/Internals/AKSettings.swift
dave234: I never use finally.

dave234:
This is awesome.

AudioKit/Common/Internals/Helpers/ExceptionCatcher.swift
```diff
@@ -0,0 +1,34 @@
...
+    let theTry = {
+        do {
+            try operation()
+        } catch let ex {
+            error = ex
+        }
+    }
+    
+    let theCatch: (NSException) -> Void = { except in
+        error = NSError(domain: "io.audiokit.AudioKit",
```
dave234: Here's what I was doing for the exception wrapping.  Take it or leave it :)

```
var userInfo = [String: Any]()
userInfo[NSLocalizedDescriptionKey] = exception.description
userInfo[NSLocalizedFailureReasonErrorKey] = exception.reason
userInfo["exception"] = exception
return NSError(domain: "io.audiokit.AudioKit", code: 0, userInfo: userInfo)
```
It would be really cool if we could print the stack trace too.

PR #1236, opened 2018-02-10
Wrap NSException in NSError for AKTry

dave234:


aure:
I'd also love to see some documentation for this addition.

PR #1240, opened 2018-02-14
Have added a gain stage to the player by exposing the gain that is in…

ryanfrancesconi:
… the fader node. So basically player has a built in AKBooster that's available to use.


aure:
Cool!

PR #1242, opened 2018-02-15
Update READMEs for ExtendingAudioKit projects

getdunne:
The two README.md files have been edited to reflect the fact that the projects themselves have been moved under the new Developer folder, and also to provide step-by-step instructions for creating similar projects using both the "frameworks" and "subproject" approaches.

### DIFF
--- a/Developer/macOS/ExtendingAudioKit/README.md
+++ b/Developer/macOS/ExtendingAudioKit/README.md
@@ -1,53 +1,74 @@
 # Extending AudioKit
-The **ExtendingAudioKit** example project illustrates how you can write your own Audio Unit code in a way which is compatible with the AudioKit classes, in effect creating your own *extension* to the AudioKit framework.
+The **ExtendingAudioKit** example projects illustrate how you can write your own Audio Unit code in a way which is compatible with the AudioKit classes, in effect creating your own *extension* to the AudioKit framework, in the context of a full AudioKit *application* for testing your new code.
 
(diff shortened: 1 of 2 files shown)


Developer/macOS/ExtendingAudioKit/README.md
```diff
@@ -1,53 +1,74 @@
...
+* In the Xcode Project Navigator, right-click your project icon and choose "New File...", select Header File, name your file as *[your project name]-Bridging-Header.h*, and save it into the same folder as your project file.
 * The file will be created with the old #ifndef … #endif syntax. Replace with #pragma once.
 
 Back in your project’s Build Settings, locate the Objective-C Bridging Header item (search for “bridging”). Set it to *$(SRCROOT)/[your project name]-Bridging-Header.h*.
 
-Next locate the Header Search Paths item (search for “search”), and set it to point to the *AudioKit/AudioKit/Common/Internals/CoreAudio* folder:
+### Prepare to use AudioKit core header files
+Before you can build your own extensions to AudioKit, you need to ensure that your Xcode project can make use of some of AudioKit's core header files.
+
+Locate the Header Search Paths item (search for “search”), and set it to point to the *AudioKit/AudioKit/Common/Internals/CoreAudio* folder:
```
aure: I don't think this is how I implemented it in the end, I think I pointed to the Headers folder generated in the framework because this would contain the headers that were properly accessible once the framework was used whereas the CoreAudio reference would a) not pick them all up and b) may accidentally see a file that wasn't actually included with AK (unlikely but possible).

aure: I'm happy to accept this PR and have another be created as time becomes available.

PR #1251, opened 2018-02-28
Add pitch bend message sending function to MIDI+Sending

cemolcay:
We can send pitch bend messages through the AKMIDI with a UInt16 value between 0 and 16383 where 
- 0 represents lowest pitch bend possible, 
- 16383 highest pitch bend possible and 
- 8192 is zero pitch bending.


AudioKit/Common/MIDI/AKMIDI+SendingMIDI.swift
```diff
@@ -135,4 +135,17 @@ extension AKMIDI {
         self.sendMessage(message)
     }
 
+    /// Send a pitch bend message.
+    ///
+    /// - Parameters:
+    ///   - value: Value of pitch shifting between 0 and 16383. Send 8192 for no pitch bending.
+    ///   - channel: Channel you want to send pitch bend message. Defaults 0.
+    public func sendPitchBendMessage(value: UInt16, channel: MIDIChannel = 0) {
```
dave234: Great minds think alike https://developer.apple.com/documentation/avfoundation/avaudiounitmidiinstrument/1390430-sendpitchbend

Issue #1254, opened 2018-03-01
Umbrella header missing 'AKExponentialParameterRamp.hpp' / Linker Error


josevazquez:
I have a simple macOS MIDI project which works fine when compiled against v4.0.4. When update AudioKit to more recent versions of the `develop` branch (484157dc007e3c5e4dc960dc63ef8f985acf7c41) the library still compiles fine, but my fails to link against it. First, I get the following warning:
```
<module-includes>:1:1: warning: umbrella header for module 'AudioKit' does not include header 'AKExponentialParameterRamp.hpp'
#import "Headers/AudioKit.h"
^
```
Followed by a series of undefined symbol errors during linking. (the full error message ca be seen [here](https://gist.github.com/josevazquez/8c664186cd920f8b319532a03f3f78ee).)

My app is a simple macOS app, straight from an Xcode template. I then dragged the AudioKit project and added the AudioKit framework to my project. Like I said it works fine when I take AudioKit back to v4.0.4.

megastep:
You need to add `-lstdc++` to your linker flags. This is now a requirement with the static library shipping with version 4.1.

josevazquez:
That worked, thank you.

Issue #1163, opened 2017-12-17
Sysex > 256 bytes, unhandled exception


analoguwings:
Hi,

I am trying to build a (MacOS) app that sends patch and bank data to vintage synths via sysex. The banks are 4kb. 

I was getting a runtime error doing this with AKMIDI.sendEvent in my own app, so I tried the sysex send function in the MIDI Utility example, with the same result.

Through trial end error, it looks like the crash happens with > 256 bytes of sysex. From looking at the source, I get the feeling larger sysex commands are supposed to be broken up into MIDIPackets, though I'm uncertain how to do this.

I have two requests:

- Can MIDIEvent throw an exception when it has > 256 bytes of sysex?

- Can we get an example of how to send > 256 byte sysex, maybe in the MIDI Utility project?

Thanks!

aure:
Sorry it has taken us a while to address this issue.  Hopefully this week.

josevazquez:
I ran against this very same problem in my own app. I have a patch working under v4.0.4. I'll open a PR as soon as I can, but I am currently having trouble compiling against `develop` (see issue https://github.com/AudioKit/AudioKit/issues/1254)

josevazquez:
I went ahead and opened the PR https://github.com/AudioKit/AudioKit/pull/1255 with the potential fix. I did not throw an exception when the data is too large (>65536 bytes) I just post an error to AKLog. I'm new to this project so I'm not sure how such situations should be handled.

PR #1255, opened 2018-03-01
Fix unhandled exception when sending SysEx messages larger than 258

josevazquez:
**NOTE** ~~While I verified that this fix worked on v4.0.4, I am currently having trouble compiling `develop` itself (as noted in https://github.com/AudioKit/AudioKit/issues/1254.) So I recommend not merging this PR until it can be verified against `develop`~~ Added the missing `-lstdc++` linker flag. Verified against `develop` now

This PR should fix the issue in https://github.com/AudioKit/AudioKit/issues/1163

The exception was being thrown whenever deallocating `packetListPointer` but the problem really was caused earlier. These are the things that I found to be off:
1. `packetListPointer` was allocated using `UnsafeMutablePointer`'s `allocate(capacity:)` function with a capacity of 1. This created a memory allocation of 272 bytes as that is the size of a single `MIDIPacketList` (at least on a mac) This buffer size did not take into account the size of the incoming data which works in a majority of case, as most MIDI messages tend to be relatively small.
2. The `listSize` parameter of `MIDIPacketListAdd()` was hardwired to 1_024. This value is larger than the actual buffer allocated in 1. and therefore the function can't properly determine that data with more than the available 258 bytes (272 bytes in buffer - 14 bytes of headers from MIDIPacketList and MIDIPacket) but less than 1014, should have not been allowed and returned with null. Instead, it just goes ahead an writes to memory beyond what was allocated, which is why dealloc throws an exception later. Running Address Sanitizer pinpoints the actual location of the error.

To fix it, I took the following steps:
1) Computed the size of the necessary headers. Since this value should not change from time to time, I just calculated this value once. (perhaps there is a better place for me to store these constants? I'm new to this project and I'm getting my bearings)
2) Allocated the memory necessary by using Data (the bridged class formerly known as NSData.) This avoided the need to alloc and dealloc (which would have work as well, this just seemed nicer)
3) Made sure the size of the buffer was passed properly to `MIDIPacketListAdd()`
4) Report an error and bail if `MIDIPacketListAdd()` returns a null (it shouldn't since the buffer is always allocated to be just the necessary size)
5) Check that the buffer size required will not be greater than the maximum allowed 65536 (as documented in the *discussion* section of `MIDIPacketListAdd()`

### DIFF
--- a/AudioKit/Common/MIDI/AKMIDI+SendingMIDI.swift
+++ b/AudioKit/Common/MIDI/AKMIDI+SendingMIDI.swift
@@ -26,7 +26,6 @@ private let sizeOfMIDIPacketListHeader = sizeOfMIDIPacketList - sizeOfMIDIPacket
 private let sizeOfMIDIPacketHeader = MemoryLayout<MIDITimeStamp>.size + MemoryLayout<UInt16>.size
 private let sizeOfMIDICombinedHeaders = sizeOfMIDIPacketListHeader + sizeOfMIDIPacketHeader
 
-
 internal extension Collection where Index == Int {
     var startIndex: Index {
         return 0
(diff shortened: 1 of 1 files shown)


aure:
Nice thank you!

PR #1258, opened 2018-03-04
Adding section about mixing objc and swift

ideaMarcos:
Adding section to README about including a dummy swift file for objective c projects


megastep:
Unfortunately this is no longer correct for a couple of reasons:
- The `Dummy.swift` file is no longer included in the example because it is not really needed;
- There is a better way to do this by setting your library search path to include `$(TOOLCHAIN_DIR)/usr/lib/swift/$(PLATFORM_NAME)`; this is what the pod does.

ideaMarcos:
So setting the library search path does work but then I get a warning:
```
warning: using 'ALWAYS_SEARCH_USER_PATHS = YES' while building targets which define modules ('DEFINES_MODULE = YES') may fail. Please migrate to using 'ALWAYS_SEARCH_USER_PATHS = NO'.
```
Searching around I see that ALWAYS_SEARCH_USER_PATHS is deprecated so I'm not sure is the right way. I'm write Obj-C infrequently so I don't really know which is correct. I'm just trying to clean up the readme.

From Stackoverflow https://stackoverflow.com/a/43488650

megastep:
That seems rather unrelated, though to be safe you'd probably want to have `$(inherited)` in there as well so that it doesn't override any parent setting at the project level.

PR #1269, opened 2018-03-15
EZRecorder / EZMicrophone buffer fix

leonardocustodio:
Fix an issue with EZAudio where the app would crash when trying to Record with a different Buffer size than 512 and 1024.


aure:
Thanks!

PR #1281, opened 2018-03-22
Fix AKCompressor connection bug

dave234:
Fix bug where effect input was only working if connection was made in init.  

Untested, but should work 🤞

### DIFF
--- a/AudioKit/Common/Nodes/Effects/Dynamics/Compressor/AKCompressor.swift
+++ b/AudioKit/Common/Nodes/Effects/Dynamics/Compressor/AKCompressor.swift
@@ -14,7 +14,7 @@ open class AKCompressor: AKNode, AKToggleable, AUEffect, AKInput {
 
     private var au: AUWrapper
 
-    fileprivate var mixer: AKMixer
+    fileprivate var mixer = AKMixer()
 
     /// Threshold (dB) ranges from -40 to 20 (Default: -20)
     @objc open dynamic var threshold: Double = -20 {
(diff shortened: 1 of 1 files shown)


AudioKit/Common/Nodes/Effects/Dynamics/Compressor/AKCompressor.swift
```diff
@@ -114,22 +115,17 @@ open class AKCompressor: AKNode, AKToggleable, AUEffect, AKInput {
         self.releaseTime = releaseTime
         self.masterGain = masterGain
 
-        input?.connect(to: inputGain)
         inputGain.volume = 0
-        mixer = AKMixer(inputGain)
-
-        effectGain = AKMixer(input)
```
dave234: This was the issue, input connects to both inputGain and effectGain on init.  inputGain was the inputNode, so connection after instantiation wouldn't route through the effect.

AudioKit/Common/Nodes/Effects/Dynamics/Compressor/AKCompressor.swift
```diff
@@ -114,22 +115,17 @@ open class AKCompressor: AKNode, AKToggleable, AUEffect, AKInput {
...
         self.internalEffect = effect
-
         AudioKit.engine.attach(effect)
         au = AUWrapper(effect)
-        if let node = effectGain?.avAudioNode {
-            AudioKit.engine.connect(node, to: effect)
-        }
-        AudioKit.engine.connect(effect, to: mixer.avAudioNode)
+
+        input?.connect(to: inputMixer)
```
dave234: Fixed by adding inputMixer and splitting, and returning inputMixer for inputNode.

dave234:
Addresses [this](https://stackoverflow.com/questions/49392586/connecting-aknodes-after-init/49395207#49395207) SO question.

PR #1285, opened 2018-03-29
Audiobus MIDI Sender example

crashingbooth:
An example project, building on the SenderSynth Audiobus set up, for using an AKSequencer to send MIDI through Audiobus 3 MIDI send ports.  Included in the project, in a file called Audiobus+MIDI.swift, are some helper methods which maybe should live in Audiobus.swift


aure:
Thanks!

Issue #1155, opened 2017-12-11
Use one instance of AKSequencer to load MIDI-Files after another


mahal:
Steps to reproduce:
• Use one AKSequencer
• load a midi-file into the sequencer
• let it play
• load another midi-file into the sequencer
• let it play

Expected Result:
• Sequencer nicely layers each new midi-file on top of existing
• or stops first file and starts new file (as loadMIDIFile calls initTracks and clears existing)
• and some documentation in AKSequencer how it's supposed to be

Actual Result:
• AKSequencer logs deinit and sometimes then it crashes

Testcode (add another midi-file into the playground called "Octave.mid"):
```
//: ## Sequencer
//:
import AudioKitPlaygrounds
import AudioKit

//: Create some samplers, load different sounds, and connect it to a mixer and the output
var piano = AKMIDISampler()
try piano.loadWav("Samples/FM Piano")
var mixer = AKMixer(piano)
AudioKit.output = mixer

//: Create the sequencer after AudioKit's output has been set
//: Load in a midi file, and set the sequencer to the main audiokit engine
var sequencer = AKSequencer(filename: "4tracks")

//: Do some basic setup to make the sequence loop correctly
sequencer.setLength(AKDuration(beats: 4))
sequencer.enableLooping()
sequencer.setGlobalMIDIOutput(piano.midiIn)

AudioKit.start()
sequencer.play()

//: Set up a basic UI for setting outputs of tracks
import AudioKitUI

class LiveView: AKLiveViewController {

    override func viewDidLoad() {

        addTitle("Sequencer")

        let buttonNextMidi = AKButton(title:"new midifile") { _ in
            sequencer.stop()
            sequencer.loadMIDIFile("Octave")
            sequencer.setLength(AKDuration(beats: 4))
            sequencer.enableLooping()
            sequencer.play()

        }
        addView(buttonNextMidi)
        let buttonStop = AKButton(title:"stop") { _ in 
            sequencer.stop()
        }
        addView(buttonStop)
    }
}

import PlaygroundSupport
PlaygroundPage.current.needsIndefiniteExecution = true
PlaygroundPage.current.liveView = LiveView()

```

aure:
We'll take a look at this this week, thanks for your patience.

crashingbooth:
I've written some code that will add tracks from a new MIDI file onto the existing sequencer (i.e., Sequencer nicely layers each new midi-file on top of existing). It seems to work and seems like a useful thing to be able to do.
[...]

PR #1296, opened 2018-04-05
Change to AKSequencer.loadMIDIFile() for issue 115

crashingbooth:
Re: https://github.com/AudioKit/AudioKit/issues/1155
Getting rid of old tracks using MusicSequenceDisposeTrack seems to be necessary before replacing the sequence with the one generated by the new MIDI file. Repeated calls to AKSequencer.loadMIDIFile() will now remove old tracks, create the new sequence, then build new tracks from the new sequence, which seems like expected default behaviour. I think this addresses the issue.

The other idea (i.e., overlaying a new tracks over the old ones) is a useful idea, and I will add it in the future with a different name (e.g., overlayNewMIDIFile) once I've tested it more thoroughly.

Also, I changed loopEnabled to private because it just keeps track of what the setter methods have done.

### DIFF
--- a/AudioKit/Common/MIDI/Sequencer/AKSequencer.swift
+++ b/AudioKit/Common/MIDI/Sequencer/AKSequencer.swift
@@ -22,7 +22,7 @@ open class AKSequencer {
     var musicPlayer: MusicPlayer?
 
     /// Loop control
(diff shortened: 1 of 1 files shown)


aure:
Thanks!

PR #1306, opened 2018-04-06
Normalize buffers and refactoring for AKPlayer and buffer management

ryanfrancesconi:
AKPlayer now has a normalize to peak value option isNormalized. Audio buffers have normalize, fade, reverse added as part of the extension.


ryanfrancesconi:
@dave234 updated AVAudioBufferConvenience. Please review!

PR #1312, opened 2018-04-09
Changed file name of the CDelay class from Delay to CDelay

knagle:
Updated references to the .h file throughout AudioKit. The Delay.cpp and Delay.h file would clash with files of the same name (class name is Delay) in the STK library if a developer wished to use that class. STK's Delay is currently not added to the AudioKit project, but I wonder what the harm would be in renaming the file in Devoloop just in case.


aure:
Thanks!

PR #1313, opened 2018-04-09
Fix bypass bug in filters

JagCesar:
I think this PR fixes a bug in AudioKit.

The process function in the filters is ignoring the `isPlaying` property.
So if a filter had `bypass` enabled it would still process the frames
that were sent through it.

I considered refactoring to make a cleaner solution, but decided to do as little as possible to get this reviewed by you guys first. If this is a valid fix we can discuss options on how to implement the `isPlaying`-check in a better way. But that should probably be a separate PR so we can get this fix in production.

Keep up the good work!


aure:
Oh wow, thanks a lot!  I'm very open to changes to make the code more efficient or more readable.

PR #1314, opened 2018-04-10
Clamp AKDynamicsProcessor parameter values to correct ranges

evenbrenna:
Clamps the parameter values of `AKDynamicsProcessor` to the ranges that
`AUDynamicsProcessor` accepts when opened in a DAW. 

The most significant change is `ExpansionTreshold`, which was clamped to `1...50`,
making it unusable when trying to set it up as a gate. This changes its range to `-120...0`


aure:
Thank you!

evenbrenna:
Glad to contribute! Thank you for a great framework (and amazing response time!)

PR #1322, opened 2018-04-10
Add clearSysexEvents

tengyifei:
I discovered that if there are Sysex events added to a track via `addSysex`, then `clear()` won't work properly. The track length remains the same and `MusicTrackClear` prints an error similar to "Invalid beat range or source track is empty".
The solution, for now, is to offer a way to conveniently remove those Sysex events. We already have an API to add those events so it makes sense to offer the reverse right?


aure:
Thanks!  Keep the updates coming!

PR #1323, opened 2018-04-11
Fix how route/category change observers are managed

warpling:
** UPDATED **

In my attempt to move from selector to block based notification observers in #1188 I inadvertently introduced the ability to add the same observer multiple times. Once a route change did happen and multiple route change observers were in place and `restartEngineAfterRouteChange(…)` could be called multiple times in quick secession ~~which in turn called `start()` and seemingly causing a cascade of restarts. A fun symptom of this was "volume thrashing" in control center:~~

…this was actually due to issues in _my_ code related to poor management of a `AKMicrophoneTracker` singleton. Engine restart thrashing was a thing for me but there's no evidence so far to suggest this affected users of AudioKit otherwise! Regardless this refactor should be an improvement to how these two notifications are managed. 
untitled

### Merge once…

- [x] Replace both TODO comments with explanations for future contributors.
- [x] Potentially find a Swiftier way to check if the observers exist before attempting to remove them.
- [x] Confirm that replacing an observer reference so that it loses scope removes it.

### DIFF
--- a/AudioKit/Common/Internals/AudioKit.swift
+++ b/AudioKit/Common/Internals/AudioKit.swift
@@ -33,11 +33,8 @@ public typealias AKCallback = () -> Void
 
     @objc static var finalMixer = AKMixer()
 
-    /// Notification observers
-    internal static var notificationObservers: [Any] = []
-
     // MARK: - Device Management
-    
+
     /// An audio output operation that most applications will need to use last
     @objc open static var output: AKNode? {
         didSet {
@@ -45,7 +42,8 @@ public typealias AKCallback = () -> Void
                 try updateSessionCategoryAndOptions()
                 output?.connect(to: finalMixer)
                 engine.connect(finalMixer.avAudioNode, to: engine.outputNode)
-            } catch {
+
+                } catch {
                 AKLog("Could not set output: \(error)")
             }
         }
@@ -136,7 +134,7 @@ public typealias AKCallback = () -> Void
         return nil
     }
 
-    /// Change the preferred input device, giving it one of the names from the list of available inputs.
+        /// Change the preferred input device, giving it one of the names from the list of available inputs.
     @objc open static func setInputDevice(_ input: AKDevice) throws {
         #if os(macOS)
             try AKTry {
@@ -211,16 +209,5 @@ public typealias AKCallback = () -> Void
     /// Disconnect all inputs
     @objc open static func disconnectAllInputs() {
         engine.disconnectNodeInput(finalMixer.avAudioNode)
-    }
-
-    // MARK: - Deinitialization
-
-    deinit {
-        #if os(iOS)
-            NotificationCenter.default.removeObserver(
-                self,
-                name: .AKEngineRestartedAfterRouteChange,
-                object: nil)
-        #endif
     }
 }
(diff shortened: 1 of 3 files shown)


aure:
I think that once this is merged, we'll come out with a 4.2.4

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -28,7 +28,31 @@ public typealias AKCallback = () -> Void
     // MARK: - Internal audio engine mechanics
 
     /// Reference to the AV Audio Engine
-    @objc open static var engine = AVAudioEngine()
+    @objc open static var engine = AVAudioEngine() {
+        didSet {
+            #if os(iOS)
+            if routeChangeObserver != nil { NotificationCenter.default.removeObserver(routeChangeObserver!) }
```
dave234: Use if let rather than checking for nil and force unwrapping.
`if let rtc = routeChangeObserver { NotificationCenter.default.removeObserver(routeChangeObserver) }`

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -28,7 +28,31 @@ public typealias AKCallback = () -> Void
...
+            // Subscribe to route changes that may affect our engine
+            // Automatic handling of this change can be disabled via AKSettings.enableRouteChangeHandling
+            routeChangeObserver = NotificationCenter.default.addObserver(forName: .AVAudioSessionRouteChange,
+                                                                         object: nil,
+                                                                         queue: OperationQueue.main,
+                                                                         using: { (notification) in
+                                                                            AudioKit.restartEngineAfterRouteChange(notification)
+            })
+
+            if configurationChangeObserver != nil { NotificationCenter.default.removeObserver(routeChangeObserver!) }
```
dave234: Use if let here too.
Are configurationChangeObserver and routeChangeObserver being conflated here?

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -452,10 +453,12 @@ public typealias AKCallback = () -> Void
 
     deinit {
         #if os(iOS)
-            NotificationCenter.default.removeObserver(
-                self,
-                name: .AKEngineRestartedAfterRouteChange,
-                object: nil)
+        if AudioKit.routeChangeObserver != nil {
```
dave234: Use if let for these too.

AudioKit/Common/Internals/AudioKit.swift
```diff
@@ -28,7 +28,31 @@ public typealias AKCallback = () -> Void
     // MARK: - Internal audio engine mechanics
 
     /// Reference to the AV Audio Engine
-    @objc open static var engine = AVAudioEngine()
+    @objc open static var engine = AVAudioEngine() {
```
dave234: Note: didSet will only be called in the case of `engine = myEngine`, it won't be called on init.  I wasn't aware that the engine was settable, if it wasn't a breaking change I would prefer that it be a let constant.  Either that, or we manage all of the implications that are associated with dynamically changing the engine.

warpling: (Sorry for the delay here.) I'm not sure about this one. I know an engine *can* be released but I'm not sure how or if AK would handle someone actually doing this. One reason to potentially release an engine is if it's `input` node was touched to do mic analysis (via a get or set), which would shut down the taptic engine on devices with them; the only way I've found so far to nil out the input node once touched is to annihilate the engine by nil-ing it out.

PR #1332, opened 2018-04-16
Editing MIDI files with AKSequencer

crashingbooth:
These changes (and example project) address some of the ideas mentioned in AudioKit's ['Improving MIDI'](https://github.com/AudioKit/AudioKit/projects/3) project.  In particular, processing and saving MIDI files and syncing multiple MIDI files to a single AKSequencer.

The basic idea is to give AudioKit apps an easy way to access the MIDI note data on AKMusicTrack.  That data can be edited and written back to the internal MusicTrack, copied to a new AKMusicTrack etc.

---

Changes:
Added ```AKMIDINoteData``` struct to hold note data from MusicTracks (including position)

```AKMusicTrack```:
added 3 methods making use of ```AKMIDINoteData```:
- ```open func getMIDINoteData() -> [AKMIDINoteData]``` - extracts an array of AKMIDINoteData from the internal MusicTrack
- ```open func replaceMIDINoteData(with trackMIDINoteData: [AKMIDINoteData])```  - clears an internal MusicTrack and writes the array of AKMIDINoteData in its place
- ```open func add(midiNoteData: AKMIDINoteData``` basically a helper method for replaceMIDINoteData, but might have general use

```AKSequencer```:
added 5 methods:
- ```public convenience init(fromUrl fileURL: URL)```  - because loadFile uses both filename and URL
- ```open func addMIDIFileTracks(_ url: URL, useExistingSequencerLength: Bool = true)```
- ```open func addMIDIFileTracks(_ filename: String, useExistingSequencerLength: Bool = true)``` makes a temp sequencer then copies MusicTrack data to the the existing sequencer, re-asserts looping behaviour, and optionally truncates new tracks to the original sequencer length
- ```func addMusicTrackNoteData(from tempSequencer: AKSequencer, useExistingSequencerLength: Bool)```  helper method for addMIDIFileTracks
- ```open func deleteTrack(trackIndex: Int)``` because if we're adding tracks we should probably be able to delete them too

Also added a new project in iOS examples called 'MIDIFileEditAndSync' demonstrating how these methods can be used in context.  The README also contains details of some of the quirky stuff one might encounter dealing with Sequencer and MusicTrack in this way.

---
If there any suggestions, changes, things I should add, things I should take out, etc. please let me know!  I would love to make this better, if I can.


aure:
Cool, there are a few small things I will change but I'd like to accept this and attend to the matters on develop.

PR #1339, opened 2018-04-23
Allow AKSamplePlayer to load and reload files of different sizes

knagle:
...rather than setting the maximum size to the first file loaded.

Currently the user specifies maximumSamples at init time, or if not specified, the maximumSamples property is set by the size of the first file loaded. This means loading a new file that's longer than the previous file will cause a crash.
This new approach resizes the buffer by clearing out the sp_ftbl objects if they've already been initialised and creating new ones each time a file is loaded. I was worried this might cause a memory leak, but haven't seen any evidence of such.

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Sample Player/AKSamplePlayer.swift
+++ b/AudioKit/Common/Nodes/Playback/Sample Player/AKSamplePlayer.swift
@@ -308,8 +308,8 @@ open class AKSamplePlayer: AKNode, AKComponent {
         let sizeToUse = UInt32(file.samplesCount * 2)
         if maximumSamples == 0 {
             maximumSamples = Int(file.samplesCount)
-            internalAU?.setupAudioFileTable(sizeToUse)
         }
+        internalAU?.setupAudioFileTable(sizeToUse)
         let buf = AVAudioPCMBuffer(pcmFormat: file.processingFormat, frameCapacity: AVAudioFrameCount(file.length))
         try! file.read(into: buf!)
         let data = buf!.floatChannelData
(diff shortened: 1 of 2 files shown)


aure:
Thanks.

eljeff:
We don't necessarily want to delete / reallocate memory on every load though - the maxSamples field was to basically give you a buffer you could play with in realtime, without allocating memory.

I will probably revert this change unless I'm missing something

aure:
@knagle sorry for the accept/revert here, I got a bit ahead of myself and this is really jeff's baby, maybe you two can hash it out.

knagle:
In the app I'm currently working on I can set maxSamples to be equal to some arbitrary large file size (e.g. 10 minutes), but the app allows the user to record from the microphone as well as load samples from an external file. There's potentially always a bigger file, and a way to crash the app if the user decided to load a sample (or record audio) of 10 minutes and 1 second. 

I felt resizing the memory on each file load was a safer way to handle this scenario, but perhaps deleting/reallocating for each load causes other problems I hadn't considered (like fragmentation maybe?). 

I took inspiration from the Recorder example project, but with AKSamplePlayer instead of AKPlayer. The AKPlayer seems to handle the loading of different length audio files without issue, but the AKSamplePlayer offers more control in terms of how you can affect playback in realtime.

I thought it might be a useful feature in AKSamplePlayer to dynamically resize the buffer, but if not I can live with it.

getdunne:
A compromise approach is "lazy" reallocation: the buffer size can grow (get reallocated larger) when necessary, but setting it smaller does not reallocate--you just use less of it. Ideally one would also implement a forced-reallocate function (to reallocate smaller when you have to use less memory) and a user-settable upper limit on buffer size (with a suitable default) to avoid accidentally overflowing memory.

eljeff:
Yea the idea is you do need a limit *somewhere* on how much memory is allocated, so currently we just set it on init.
I don't really want to do it 'lazy' because I think it should be clear that loading in new samples and allocating memory are different processes - I don't like the idea of the same function having fundamentally different behaviour under the hood.
If you need more time, you can init a new object, which I think makes sense since you are literally using a new chunk of memory by doing so.

PR #1340, opened 2018-04-23
Revert "Allow AKSamplePlayer to load and reload files of different sizes"

aure:
Reverts AudioKit/AudioKit#1339


aure:
As per @eljeff

PR #1343, opened 2018-04-26
Added AKTimePitch into AKPlayer.  New properties: rate and pitch.

ryanfrancesconi:
This is needed in order to keep fade scheduling in sync with a player that has been timestretched. Has the added bonus of having time stretching and pitch integrated. The stretcher will stay bypassed unless it's needed.


aure:
Cool

PR #1344, opened 2018-04-26
PPQN and tests for AKSequencer and AKMusicTrack changes

crashingbooth:
- I added a timeResolution computed var to get PPQN for Apple's sequencer
- replaced hardcoded values with defaults for the non-obvious part of the time signature event
- made AKMIDINoteData equatable and gave it a public init

I added a bunch of tests for AKSequencer and AKMusicTrack
- mostly they cover the additions I made for editing musicTrack/MIDI files (better late than never)
- I also wrote tests to document some of the quirky ways that adding/deleting/copying tracks and/or loading/copying MIDI files affect sequence  length and tempo.  The way these interact is not always obvious, so this might help someone in the future.


crashingbooth:
Okay I will go in and update the example project now (I added default arguments that I need to include in the protocol I used in the project)

crashingbooth:
I'm on it

aure:
THanks!

PR #1347, opened 2018-04-27
Variable type parameter ramper integration

ryanfrancesconi:
I need multiple rampTypes in AKBooster so I've implemented it as included in this PR. Any fade tool will have exponential, linear and logarithmic options. Currently there is now exp and linear. Need to make log. Being able to adjust slope dynamically would be even better but...

I don't know if the way I handled the C++ ramper code is the best option, so please let me know about that if it's bad.

image


aure:
Nice!

Issue #1353, opened 2018-05-02
Enhancement: Read Time Signature in MIDI File


mahal:
I started implementing a new feature in AKSequencer but couldn't make it work with Swift. I don't know if this is the best procedure but I created this Issue and made a pull request. Hopefully some of you core guys has better insights into optional mutable unsafe buffer pointer. There are some more comments in the [PR 1354](https://github.com/AudioKit/AudioKit/pull/1354)

Here's the code that works with ObjC. The upper and lower part of the method is in the pull request. 

        MusicEventIteratorGetEventInfo(iterator, &timestamp, &eventType, &eventData, &eventDataSize);
     
        if (eventType == kMusicEventType_Meta){
            metaEvent = (MIDIMetaEvent *) eventData;
            if (metaEvent->dataLength > 1) {
                timeSignatureTop = [NSNumber numberWithUnsignedInt:metaEvent->data[0]];
                timeSignatureBottomMIDI = metaEvent->data[1];
                hasNext = NO;
            }
        }
        MusicEventIteratorNextEvent(iterator);
    }
    
    DisposeMusicEventIterator(iterator);
    
    // According to MIDI spec, second byte is log base 2 of time signature 'denominator'
    switch (timeSignatureBottomMIDI) {
        case 1:
             timeSignatureBottom = [NSNumber numberWithInt:2];
            break;
        case 2:
            timeSignatureBottom =  [NSNumber numberWithInt:4];
            break;
        case 3:
            timeSignatureBottom =  [NSNumber numberWithInt:8];
            break;
        case 4:
            timeSignatureBottom =  [NSNumber numberWithInt:16];
        default:
            break;
    }

// the return part in swift (as in the pull request)
/*
        if (timeSignatureTop != nil && timeSignatureBottom != nil) {
            outValue = (timeSignatureTop, timeSignatureBottom) as? (UInt8, AKSequencer.TimeSignatureBottomValue)
        }
        return outValue
}*/

adamnemecek:
Sorry I’m out right now so I might not be of much help right now. You should check out mikmidi https://github.com/mixedinkey-opensource/MIKMIDI . As a framework it’s bloated however it’s standard compliant and it should answer your question.

mahal:
Thanks for the link. As mikmidi is in Objective-C it doesn't help me getting the values from the buffer as in lines 362ff 

                        let arrayBuffer = UnsafeBufferPointer(start: data, count: Int(data.pointee.dataLength))
[...]

PR #1354, opened 2018-05-02
Develop: read time signature of midi file

mahal:
First idea of reading time signature of midi file. Does not yet work as I'm not fluent enough in Swift. See Issue [#1353](https://github.com/AudioKit/AudioKit/issues/1353)


aure:
I'm going to accept it, because I don't want this to get stale, still hoping that you get some feedback from some of our MIDI wizards.

PR #1357, opened 2018-05-03
AKPlayer, Mostly to keep properties in sync

ryanfrancesconi:


aure:
Thanks for the vigilant update!

PR #1363, opened 2018-05-04
Clip recorder refactor

dave234:
This should really be two+ PRs.

The clip recorder was refactored to use an result type instead of the many fielded callback that was before.
Reordered the clip "tap" callback and the write to disk so that the tap can alter the audio before it gets written to disk.

There's also a change in AudioKit.start() where I addressed a deadlock caused by dispatch sync being called on the same thread it was dispatching to.

AVAudioTime can use both sample and host times, made a change so that hostTime is preferred when both are present.

I created the LoopbackRecording iOS example where we play the metronome, record it and play the recording of it along with itself. We get the recording, and playback scheduled as close to perfect as possible without applying device specific heuristics.

### Notes

I use AudioKit.engine.inputNode rather than AKMicrophone in the example. AKMicrophone currently uses an internal mixer, but this hides the timing info needed for accurate scheduling when tapping it. It would be nice to refactor AKMicrophone, or create AudioKit.inputNode to get around having to access the engine so much.

I also don't call AudioKit.start(), there were multiple instances where I was getting off base warnings and unnecessary restarts, I was also forced to enable background mode, avoiding the use of AudioKit.output and AudioKit.start() seemed to make the issues go away. It is my opinion that the AVAudioSession management, along with any restarting code is application logic, and doesn't belong in AudioKit. At very least it should be separated from the rest of the audio management as much as possible.


AudioKit/Common/Internals/AudioKit+StartStop.swift
```diff
@@ -130,12 +130,18 @@ extension AudioKit {
                 // Note: doesn't throw since this is called from a notification observer
             }
         }
+        if Thread.isMainThread {
```
dave234: @warpling If the notification. was called on the main thread you would get a deadlock on DispatchQueue.main.sync.  In general, use async unless there's a concrete reason to use sync on the main thread.

aure:
This is cool.  Feel free to continue to discuss issues raised here, either directly on this pr or in the slack group.

AudioKit/Common/Internals/AudioKit+StartStop.swift
dave234: Not really, no.

I'm not sure why it's there in the first place, but it was `DispatchQueue.main.sync` which was deadlocking on me.  So this just fixes the deadlock without having to read up why there was a dispatch_sync here originally.

dave234: I suppose I’m just assuming that sync was a mistake.

ryanfrancesconi: there's a number of examples like this one online (LIKE https://gist.github.com/sgr-ksmt/4880c5df5aeec9e558622cd6d5b477cb), with the isMainThread and sync on main, but doing a little more reading I found that it is not in fact safe as you found. I would assume that whoever wrote that saw this sort of safe dispatch example thinking it was exactly that. Seems like it's best to just get rid of the if statement entirely. I don't see why sync should be called on the main queue ever?

dave234: I’d let @warpling answer this.  The fix is using async, not sync, so it should be good to go AFAIK.

PR #1367, opened 2018-05-09
Many small fixes for AKSequencer/MusicTrack with tests

crashingbooth:
**Additions**

- added computed var ```allTempoEvents``` to get return array of (MusicTimeStamp, bpm)
- added ```getTempo(at position: MusicTimeStamp)```

**Bug  Fixes/Other changes**

- currently the tempo computed var won’t return correct values when the tempo event and the current position are the same (e.g. at MusicTimeStamp 0.0).  I think this could be easily fixed by using a MusicTimeStamp just in front of the current value, but I have left this intact, because there might be some reason for this setup.  But I think it could in principle be replaced by ```getTempo(at position: currentPosition.beats)```.
- The ```setTempo(bpm:)``` method included a call to ```MusicTrackClear()```, but as it turns out, this was not actually deleting any tempo events.  If ```setTempo``` was called 15 times on a sequence, that sequence would then contain 15 extendedTempoEvents at MusicTimeStamp 0.0.  Apple’s sequencer is smart enough to ignore all but the most recently added event so this doesn’t cause problems, but clearly this isn't the desired behaviour.  I have added a ```clearTempoEvents``` helper function to stop these tempo events from accumulating.
- When making the clearTempoEvents method I discovered that calling ```MusicEventIteratorDeleteEvent```  causes the following event to become the de facto current event, so the subsequent call to MusicIteratorNextEvent was causing the iterator to skip over any event that immediately followed a deleted one.  Testing showed that all AK methods using ```MusicEventIteratorDeleteEvent``` (about 5 in total) had this problem. This is now fixed, and they have tests.
- Additionally, the ```clearHelper()``` method used by  ```clear()```, ```clearMetaEvents()```, and ```clearSysexEvents()``` contained a trivial ```eventType == eventType``` comparison that caused it to delete ALL events.  This is also fixed
- I have also removed a method called ```isTempoTrackEmpty```.  This is not part of the API, and is not called internally anywhere within AK.  Presumably it was there for internal testing . . . but when I tried to use it I discovered that it always returns true. It’s easy enough, to fix but ```allTempoEvents.isEmpty``` will do the same thing, so it would be redundant.

### DIFF
--- a/AudioKit/Common/MIDI/Sequencer/AKMusicTrack.swift
+++ b/AudioKit/Common/MIDI/Sequencer/AKMusicTrack.swift
@@ -257,15 +257,17 @@ open class AKMusicTrack {
         }
     }
 
-    func clearMetaEvents() {
+    /// Clear meta events from the track
+    open func clearMetaEvents() {
         clearHelper(kMusicEventType_Meta, from: "clearMetaEvents")
     }
-
-    func clearSysexEvents() {
+    
+    /// Clear SysEx events from the track
+    open func clearSysexEvents() {
         clearHelper(kMusicEventType_MIDIRawData, from: "clearSysexEvents")
     }
 
-    private func clearHelper(_ eventType: UInt32, from functionName: String) {
+    private func clearHelper(_ targetEventType: UInt32, from functionName: String) {
         guard let track = internalMusicTrack else {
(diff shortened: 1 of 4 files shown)


aure:
Thanks, sorry for the delay in accepting.

Issue #1387, opened 2018-05-30
AKAudioFile exportAsynchronously calls callback with error but continues and calls callback again


CrossProd:
See this line: https://github.com/AudioKit/AudioKit/blob/master/AudioKit/Common/Internals/Audio%20File/AKAudioFile%2BProcessingAsynchronously.swift#L402

It calls the completion handler with an error but continues and successfully exports the file and calls the completion handler again.

The method documentation states that the callback will be called when export is completed: 

"callback: AsyncProcessCallback function that will be triggered when export completed."

CrossProd:
I don't mind making a pull request for this. Will do that today.

The underlying problem seems to be that the AKAudioFile reports a length & sampleCount of 0. Not sure how to fix that. I now changed my code to use recorder.audioFile instead for exporting.

aure:
Yeah, I am no fan of AKAudioFile anymore, it is a target for refactoring, but of course, we will at least wait until WWDC to see what gifts / kicks in the nuts are in store first.

CrossProd:
I'm trying to push some code to a branch but it appears I have no rights to do this. Error on my side or do I need to be given permissions?

aure:
Reopen if the fix from the PR isn’t sufficient.  Thank you!

PR #1388, opened 2018-05-30
Fix for double callback call in `exportAsynchronously` in `AKAudioFile`.

CrossProd:
This PR should solve the double callback mentioned in issue #1387.


aure:
Thank you!

PR #1394, opened 2018-06-06
AKSequencer currentRelativePosition 

mahal:
Maybe I did not understand what the relative position was meant for, but as the comment and usage implies, this should return a Double value between 0..1. 

It was a modulo and returned a AKDuration. Alternatively I suggest changing the description.

### DIFF
--- a/AudioKit/Common/MIDI/Sequencer/AKSequencer.swift
+++ b/AudioKit/Common/MIDI/Sequencer/AKSequencer.swift
@@ -523,8 +523,8 @@ open class AKSequencer {
     }
 
     /// Current Time relative to sequencer length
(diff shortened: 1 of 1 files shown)


crashingbooth:
This would be a breaking change. As it is, currentRelativePosition tells you that if you have an 8 bar sequence, and your current position is 32.5, you are effectively at beat 0.5.  I can see how getting a value between 0 and 1 might be useful, but perhaps it should be a separate method entirely to avoid breaking current code.

mahal:
Ok got it; thanks for the detailed answer.  thought that there is some use for it. I'll create a new PR with improved docu.

PR #1400, opened 2018-06-11
AKPlayer: change to make time pitch support more dynamic

ryanfrancesconi:
I've noticed a drastic performance problem (crackling audio output in the entire audio engine) when using many AKTimePitch nodes in conjunction with AKPlayer internally. To handle this, I've changed AKPlayer to only install the AKTimePitch if it is requested by a non standard rate or pitch event.


aure:
Thanks!

PR #1404, opened 2018-06-13
Ensure any previously created timers are invalidated before creating a new instance

colinhumber:
The creation and invalidation of timers should be balanced so that any previously created timers that haven't fired yet are invalidated and removed from the run loop before creating a new one to ensure the previous one will not fire.

Also, the timers should be invalidated on the same thread where they were created, as per the special considerations note on `NSTimer` docs.

When calling `setPosition` on the player after a `resume` call has been made, because the timers are being created and invalidated on different threads, the old timers aren't always removed from the run loop and still fire.


colinhumber:
🙌🏼

aure:
This got merged into Ryan's pull request so I think we're good.  Thanks!

PR #1413, opened 2018-06-22
Fixed incorrect StackOverflow link

appsandwich:
The StackOverflow link markdown was invalid, causing display issues and linking to the wrong SO tag.


appsandwich:
Messed up the PR. Will try again...

PR #1434, opened 2018-07-12
Prevent out-of-bound access in rare cases

tengyifei:
When instantiating `AKMIDIEvent` in `MIDIDestinationCreateWithBlock`, occassionally the code segfaults in the initializer.


aure:
Thanks!

PR #1439, opened 2018-07-13
Updated "Code smell in Swift" URL

markjeschke:
Updated the "Five Signs of Code Smell in Swift" link, because the current one in the ReadMe is broken.


aure:
YOu were using a README for AudioKit 3.7 that has since changed a lot.

PR #1442, opened 2018-07-18
Fix[Playgrounds]: Playback - Recording Nodes

RockerHX:
Change Label -> AKLabel for run error.
&
ExportAsynchronously failure with error "AKAudioFile export: In time must be less than Out time."


aure:
Thanks.

PR #1443, opened 2018-07-18
Added Tap Tempo and set beatTime to 0 when metronome is restarted.

markjeschke:
Added Tap Tempo button and functionality, and changed the stop/play button to always start at beatTime 0. That way, the metronome restarts on the first downbeat, rather than continue from where it left off.


aure:
Thanks!

PR #1449, opened 2018-07-19
Further refinement of sysex handling

eljeff:
Found a few cases where we weren't properly truncating sysex. Made status lookup / computation more robust and accurate. Updated macOS midi utility to default to a blank midi device so user must select one


eljeff:
Github might be glitching out - I specifically tried to merge into develop, then the dropdown changed back to master when I went to the PR page, I submitted not knowing this, now when I try to PR develop, it claims I already have a PR. Yet the PR says it wants to merge into master.

Very confusing.

PR #1452, opened 2018-07-23
Fix: AKNodeRecorder instantiation with existing AKAudioFile

probi86:
Fixed an issue where the internalAudioFile created in AKNodeRecorder init, had an incorrect file format, when an AKAudioFile is passed as parameter.


aure:
Open source at work! Thank you!

PR #1459, opened 2018-08-07
DSP Object Leak Fix

JamesOrdner:
Fixes leaking of DSP objects when nodes are removed and cleans up leaked soundpipe objects. **This required some changes in AKDSPBase (adding a virtual destructor) and AKSoundpipeDSPBase (removing the `destroy()` function) which causes the current commit to not compile without first altering child classes.** AKClipperDSP was modified as an example of what needs to be changed. As a quick summary: if a child class declares a destructor but never defines the destructor, this will now cause a linking error. The destructor declaration should be removed. However, it is no problem if a child class declares _and_ defines a destructor. Additionally, if the DSP class derives from AKSoundpipeDSPBase, the `destroy()` function should be renamed to `deinit() override`, and the function should no longer call `AKSoundpipieDSPBase::destroy()`.

### DIFF
--- a/AudioKit/Common/Internals/AKSoundpipeDSPBase.hpp
+++ b/AudioKit/Common/Internals/AKSoundpipeDSPBase.hpp
@@ -46,10 +46,6 @@ public:
     virtual void setParameter(AUParameterAddress address, AUValue value, bool immediate) override {}
     virtual AUValue getParameter(AUParameterAddress address) override { return 0.0f; }
 
-    void destroy() {
-        //printf("AKSoundpipeKernel.destroy(), &sp is %p\n", (void *)sp);
-    }
-
     virtual void processSample(int channel, float* in, float* out) {
         *out = *in;
     }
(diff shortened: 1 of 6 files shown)


aure:
I've made a branch to continue the work on.

JamesOrdner:
Hmm okay, I haven't gotten this error in my project but we completely rebuild the audio graph upon each audiokit restart. My guess would be that `initWithComponentDescription` is called only once for each node, when it is first created, which creates the kernel. Seems like `allocateRenderResourcesAndReturnError` and `deallocateRenderResources` are being called upon each audiokit start and stop, in which case the solution would be to leave `_kernel->deinit()` inside `deallocateRenderResources`, because it frees anything created with `_kernel->init()`, and move `delete _kernel` to the equivalent of the AudioUnit's destructor, or something called during the AudioUnit's cleanup. I don't know much about Obj-C, do the classes have a built in destructor we can use?

JamesOrdner:
Alright perfect, I added

```
- (void)dealloc {
    delete _kernel;
}
```

to both AKAudioUnitBase and AKGeneratorAudioUnitBase. It still fixes the memory leak, and now `_kernel` is persistent for the lifetime of the AudioUnit object and fixes that issue!

aure:
Why should I take your credit, its always to good to have proper attribution for changes, so please do submit a PR.  Thanks!

JamesOrdner:
Sounds good :) Submitted.

ryanfrancesconi:
Excellent. Thanks! I'll give it a try.

PR #1476, opened 2018-08-24
Delete _kernel objects in destructor

JamesOrdner:
Moved deleting _kernel DSP objects from `deallocateRenderResources` to destructor.


aure:
Thanks!

ryanfrancesconi:
it seems good now. will report back if i find any other crashes. thanks!

PR #1477, opened 2018-08-25
updated docs for clearRange, set Tempo

crashingbooth:
Updated docs in reference to https://stackoverflow.com/questions/52019770/deleting-tempo-events-from-aksequencer


aure:
Not sure how this could have made Travis tests not pass, but I'll accept it since it could not be the cause.

PR #1501, opened 2018-09-17
Restructured ModulatedDelay class

megastep:
We ran into issues with the visibility of some of the code when distributing the framework, so we need to fix this issue specifically for the `AKModulatedDelayDSP` class and related classes.

My solution here leaves the `ModulatedDelay` mostly as is, though it is now officially public and renamed to `AKModulatedDelay` to make it clear. It is also technically outside of the AudioKitCore namespace because of that.

I had to encapsulate some other internal dependencies inside of the cpp file, and added a couple of accessors for some of these properties, but I think this should take care of this issue.


aure:
I'm good with this if Shane is.

PR #1502, opened 2018-09-17
Prevent error where invalid timestamp is passed to extrapolateTime()

dave234:
# UNTESTED!


aure:
I'll test

aure:
Works for me!

PR #1504, opened 2018-09-18
Renamed node 'disconnect' function overrides to instead override 'detach'

JamesOrdner:
This updates nodes which had not been changed to reflect the depreciation of disconnect(). This will allow easier node cleanup for all nodes with a single node.detach() call as designed.


aure:
Sorry I hadn't seen this pull request till now.

PR #1509, opened 2018-09-22
Remove AKMicrophoneTrackerEngine

dave234:
Doing this as step one to remove EZAudio audio system IO access. The ultimate goal is to reduce the bug surface. We shouldn't have two components separately accessing the microphone or the output unit.

I haven't tested this, just going on faith that AKMicrophone and AKFrequencyTracker will work as a replacement.


dave234:
@warpling Could you test this?

dave234:
That commit message should have said remove AKMicrophoneTrackerEngine instead of AKFrequencyTracker :/

warpling:
Oh man, thanks for the heads up on this. All for reducing the bug surface. I believe the reason Aure set this up for me was because using AKFrequencyTracker would touch the audio engine's input node which disabled device haptics until the input node was relinquished by replacing the engine. By using a separate tracker engine we could enable it for mic-based stuff and then throw it away immediately after to resume use of haptics.

I'll try to test if there's a way around this now but won't be able to until October 1st or so :[

dave234:
October works for me!

PR #1522, opened 2018-10-03
Add a hard shutdown option to AudioKit (e.g. for Media Service Resets)

### LINKED ISSUE #1474: Crash / Failure When Handling MediaServicesReset on iOS
Hi,

I apologize in advance if it ends up that StackOverflow is the appropriate place to post this - however I have searched the codebase to the best of my ability and I cannot find a reference to the iOS MediaService notifications / events and given that by "default" usage of AudioKit I experience a crash, I felt it appropriate to post an issue to GitHub.  Happy to change if needed, just let me know.

So, the issue I am seeing is that when I trigger a `Reset Media Services` from the `Settings -> Developer` menu on an iOS device, the next subsequent `AKPlayer.play()` call will fail with an assertion exception that looks something like:
`... required condition is false: _engine->IsRunning() Terminating app due to uncaught exception 'com.apple.coreaudio.avfaudio', reason: 'required condition is false: _engine->IsRunning()'`

I poked around a bit in `AudioKit+StartStop.swift` and it seems that there is some code to handle other similar events (route / configuration changes); fair enough, maybe I just need to do something similar in my own `AVAudioSessionMediaServicesWereReset` notification observer.

However, I can't seem to come up with a combination of calls that will properly reset the AudioKit engine without that "restart" failing separately (or crashing, depending on how I call things) with the following logs:
```
2018-08-21 16:15:12.634808-0700 Drivetime[7967:2553564] [aurioc] 921: failed: -50 (enable 2, outf< 2 ch,  44100 Hz, Float32, non-inter> inf< 2 ch,      0 Hz, Float32, non-inter>)
2018-08-21 16:15:12.672142-0700 Drivetime[7967:2553564] [avae] AVAEInternal.h:103:_AVAE_CheckNoErr: [AVAudioEngineGraph.mm:1227:Initialize: (err = PerformCommand(*outputNode, kAUInitialize, NULL, 0)): error -50
2018-08-21 16:15:12.672700-0700 Drivetime[7967:2553564] *** Terminating app due to uncaught exception 'com.apple.coreaudio.avfaudio', reason: 'error -50'
```

Based on my (limited) understanding of AVAudioEngine et al., it seems like the internal engine initialization on my "second start" / "restart" is considered to be passing bad parameters of some sort to the system.

One operating theory I have is that perhaps the existence of the `finalMixer` internally from the previous engine is breaking things when calling start a second time (I've seen multiple suggestions / references elsewhere that in the event of a media services reset all nodes etc. should be completely dropped and re-created).

However, this appears to be set to `internal` so I can't test out resetting the `finalMixer` myself without building the library from code / scratch (not ideal for me but I can if necessary).

Anyways, to make a long story short - I feel like I'm missing something obvious here and I was hoping maybe you could help me understand why a media services reset doesn't appear to be handled like route / configuration changes already are, if these crashes / issues are expected as a result of a media services reset in the current codebase, and what I should do to handle them properly myself while using AudioKit.

Finally, for context, here are the two most relevant examples of what I've tried doing to handle this so far (without success) - keep in mind I am dispatching the call to this function to the main queue from the MediaServicesReset observer based on the example I saw in AudioKit+StartStop.swift:
```
func resetAudioKit() {
    do {
      // I've tried just calling start directly (and also calling engine.start() directly) as that's more or less how I see it handled in AudioKit+StartStop.swift
      mixer = AKMixer()
      AudioKit.output = mixer
      try AudioKit.start()  // I have also tried calling AudioKit.engine.start() directly as well.
    } catch {
      print("AudioKit Reset ERROR!")
    }
  }
```
```
func resetAudioKit() {
    do {
      // If we've already setup before, I have also tried stopping AudioKit first before starting it.
      if mixer != nil {
        try AudioKit.stop()  // I have also tried calling try AudioKit.engine.stop() directly as well.
      }
      mixer = AKMixer()
      AudioKit.output = mixer
      try AudioKit.start()  // I have also tried calling AudioKit.engine.start() directly as well.
    } catch {
      print("AudioKit Reset ERROR!")
    }
  }
```

cojo:
This PR fixes #1474 (Crash / Failure When Handling MediaServicesReset on iOS).

It does so by adding a `shutdown` function to `AudioKit+StartStop.swift` which completely resets the core `AVAudioEngine` and `AKMixer`, setting `output` to nil, and marking `shouldBeRunning` as false.

This allows a user to then redo all of their internal / app-specific `AudioKit` setup after calling `shutdown()`, calling `start()` again when ready, and have a working `AudioKit` instance even if they have experienced a Media Services Reset (see http://devstreaming.apple.com/videos/wwdc/2016/507n0zrhzxdzmg20zcl/507/507_delivering_an_exceptional_audio_experience.pdf?dl=1 for more discussion / description on these events and how they should be handled).

It also adds an `AKTry` around `AKPlayer`'s `handleCallbackComplete` as we found a related potential crash / exception while testing this fix / workflow in our app (which uses `AKPlayer` heavily).

Let me know if this is not something you wish to have in core; completely understandable if that is the case but thought I'd offer it up both in case you wish to integrate it and as a way of wrapping up / closing out #1474 .

### DIFF
--- a/AudioKit/Common/Internals/AudioKit+StartStop.swift
+++ b/AudioKit/Common/Internals/AudioKit+StartStop.swift
@@ -96,6 +96,13 @@ extension AudioKit {
         }
         #endif
     }
+    
+    @objc public static func shutdown() throws {
+        engine = AVAudioEngine()
+        finalMixer = AKMixer()
+        output = nil
+        shouldBeRunning = false
+    }
 
     // MARK: - Configuration Change Response
 
--- a/AudioKit/Common/Internals/AudioKit.swift
+++ b/AudioKit/Common/Internals/AudioKit.swift
@@ -23,7 +23,7 @@ public typealias AKCallback = () -> Void
     // MARK: - Internal audio engine mechanics
 
     /// Reference to the AV Audio Engine
-    @objc public static let engine = AVAudioEngine()
+    @objc public static internal(set) var engine = AVAudioEngine()
 
     /// Reference to singleton MIDI
 
--- a/AudioKit/Common/Nodes/Playback/Players/Player/AKPlayer+Playback.swift
+++ b/AudioKit/Common/Nodes/Playback/Players/Player/AKPlayer+Playback.swift
@@ -234,8 +234,14 @@ extension AKPlayer {
                 self.pauseTime = nil
                 return
             }
-            if self.currentFrame >= self.frameCount {
-                self.handleComplete()
+            do {
+                try AKTry {
+                    if self.currentFrame >= self.frameCount {
+                        self.handleComplete()
+                    }
+                }
+            } catch let error {
+                AKLog("Failed to check currentFrame and call completion handler: \(error)... Possible Media Service Reset?")
             }
         }
     }


aure:
Thanks for the follow through.

PR #1524, opened 2018-10-06
Fix typo

niklassaers:
frquency => frequency


aure:
Every little bit helps! Thanks.

PR #1526, opened 2018-10-08
Resume AKDynamicPlayer if it was playing before TimePitch was removed

juusokos:
This pr simply resumes AKDynamicPlayer if it was playing when setting pitch or rate back to default value. Currently playing will stop in these situations as play() isn't called after stop() in removeTimePitch().


aure:
Thanks!

PR #1535, opened 2018-10-12
add microphone privacy usage for Mojave

ZAYEC77:
When I was trying to run few demos at macOS Mojave I had found issue with privacy.


aure:
Thanks, I was wondering why that wasn't working!

PR #1543, opened 2018-10-17
Update README.md with correct Matthew Fecher link (https://github.com/analogcode)

VictorNouvellet:
Update README.md with correct Matthew Fecher Github link (https://github.com/analogcode)


aure:
Thanks.

PR #1553, opened 2018-10-26
Fix EXC_BAD_ACCESS in AKConvolution processing

Bautastein:
Before this fix, if start() is called on an instance of AKConvolution
before inserting it into the audio processing chain, there will be an
EXC_BAD_ACCESS in sp_conv_compute. If start() is called immediately
after inserting it into the chain, a race condition will cause the same
failure to occur inconsistently.

The fix does not change functionality or interface of any AudioKit Swift
classes. It does add one new member function and changes the
functionality of the internal AU classes AKConvolutionAudioUnit and
AKConvolutionDSPKernel.

The bug is caused by resources (conv0, conv1) being allocated in the
init() function of the AKConvolutionDSPKernel, and initialized in the
start() function. As init() gets called again when the AKNode is
inserted into the sound chain, new resources are allocated for the same
variables, and then used uninitialized during sound processing. This
could be solved by moving both allocation and initialization into the
start() function, but this may cause further race conditions, and if
start() is called again during processing, new EXC_BAD_ACCESS faults
will probably occur.

This fix instead ensures both allocation and initialization is performed
during construction of the AKConvolution class, so the object is
ready-to-use after construction. However, this must be done in a
callback method which is not guaranteed to be finished when the
constructor returns. For this reason, the fix makes
AKConvolutionDSPKernel turned off by default after construction, and
then it's turned on at the end of the callback, when all resources are
allocated and initialized.

### DIFF
--- a/AudioKit/Common/Nodes/Effects/Reverb/Convolution/AKConvolution.swift
+++ b/AudioKit/Common/Nodes/Effects/Reverb/Convolution/AKConvolution.swift
@@ -67,12 +67,12 @@ open class AKConvolution: AKNode, AKToggleable, AKComponent, AKInput {
     @objc open func start() {
         internalAU?.start()
     }
-    
+
     /// Function to stop or bypass the node, both are equivalent
     @objc open func stop() {
         internalAU?.stop()
     }
-    
+
     private func readAudioFile() {
         Exit: do {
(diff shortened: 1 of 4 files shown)


aure:
Thanks a lot.  Great write-up too.

PR #1555, opened 2018-10-27
Adjust access level for AKTiming methods to synchronizing multiple nodes

derekleerock:
In researching how to synchronize two AKTiming nodes I discovered that both the syncStart() and synchronizeWith() methods did not have access modifiers which allowed them to be called outside of AudioKit. 

I confirmed with @dave234 on Slack that these should be accessible and this PR is simply to make that adjustment.


aure:
That's fine, thanks for the PR!

PR #1559, opened 2018-11-05
Choose bundle based on the location of AKAudioFile so we can load fil…

wahidac:
…es inside XCTest, which consumes resources in a different bundle that the main bundle

Couldn't load files from XCTest because it was using the Main Bundle by default, and the resources were located in a different bundle. Basically used this solution: https://stackoverflow.com/questions/19151420/load-files-in-xcode-unit-tests/31651573


aure:
Thanks!

Granfalloner:
@wahidac yes, for sure.
@aure this change needs to be reverted.

PR #1563, opened 2018-11-08
Fix AKCoreSampler mono-non-legato click bug

getdunne:
Clicks occurred because notes were being abruptly started, not re-started.
This commit includes one file from the AKSampler VST demo project, providing option to start up with a pure sine wave, useful for testing for clicks.


aure:
Woohoo!

PR #1568, opened 2018-11-10
Added NSMicrophoneUsageDescription to Info.plist

nodeful:
Otherwise, macOS Mojave doesn't allow to capture any Input Device Audio.


aure:
Thanks!

PR #1574, opened 2018-11-21
Fix bug in loading SFZ file with AKSampler

### LINKED ISSUE #1573: [Bug] Error loading samples in SFZ file with AKSampler
I was having problems with getting loading the SFZ file I have with `AKSampler`. Upon further investigation, I noticed the sample path was being read incorrectly. For example the following line:

```
<region> sample=test.wav hirand=0.250
```

would leave the sample variable to be `test.wav hirand=0.250` when it should be `test.wav`

mrniket:
Fixes #1573

This PR removes the assumption that the sample attribute associated with the `region` tag has to be at the end of the list of attributes. This should hopefully make the `loadSFZ` function compatible with a few more SFZ files.


aure:
Thanks!

PR #1578, opened 2018-11-22
Fix AVAudioFormat init bug

netizen01:
When connecting a USB device with a complex channel layout the standard AVAudioFormat constructor will fail and return nil, crashing when it’s force-unwrapped in AKMicrophone init. This copies the input’s channel layout directly.


aure:
Thanks very much!

PR #1579, opened 2018-11-26
BufferedAudioUnit as base class for AKAudioUnitBase and AKGeneratorAudioUnitBase

dave234:
BufferedAudioUnit is a base AUAudioUnit class that only deals with the buffer management aspect of an AUAudioUnit subclass. 

Subclasses should now override `processEventsBlock` instead of internalRenderBlock. ProcessEventsBlock is a pared down version of internalRenderBlock where the output buffer is always valid, and the input buffer is valid unless `shouldAllocateInputBus` is overridden  to return false (Generator).

Ideally BufferedAudioUnit should handle multiple inputs, but I don't currently have the time to implement a system for dynamically adding/removing busses and connections. That being said, BufferedAudioUnit is a good place to add this functionality in the future. My only reservation about this implementation is that processEventsBlock is set up for a single input. It might be better to change the block signature so that we have multiple input buffers available, but just render one for now. I'm all ears.

### DIFF
--- a/AudioKit/Common/Internals/CoreAudio/BufferedAudioUnit.h
+++ b/AudioKit/Common/Internals/CoreAudio/BufferedAudioUnit.h
@@ -1,0 +1,32 @@
+//
+//  BufferedAudioUnit.h
+//  AudioKit
+//
+//  Created by Dave O'Neill., revision history on Github.
+//  Copyright © 2018 AudioKit. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import <AVFoundation/AVFoundation.h>
+#import <AudioToolbox/AudioToolbox.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void(^ProcessEventsBlock)(AudioBufferList        * _Nullable inBuffer,
+                                  AudioBufferList        * _Nonnull  outBuffer,
+                                  const AudioTimeStamp   * _Nonnull  timestamp,
+                                  AVAudioFrameCount                  frameCount,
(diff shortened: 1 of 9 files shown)


AudioKit/Common/Internals/CoreAudio/BufferedAudioUnit.h
```diff
@@ -0,0 +1,32 @@
...
+#import <AVFoundation/AVFoundation.h>
+#import <AudioToolbox/AudioToolbox.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void(^ProcessEventsBlock)(AudioBufferList        * _Nullable inBuffer,
+                                  AudioBufferList        * _Nonnull  outBuffer,
+                                  const AudioTimeStamp   * _Nonnull  timestamp,
+                                  AVAudioFrameCount                  frameCount,
+                                  const AURenderEvent    * _Nullable eventsListHead);
```
dave234: If we wanted to future proof for multiple inputs we could change ProcessEventsBlock to something more like this:
```
typedef void(^ProcessEventsBlock)(
    AudioBufferList * _Nullable * _Nonnull  inBuffers,
    int                                     inBuffersCount,
    AudioBufferList             * _Nonnull  outBuffer,
    const AudioTimeStamp        * _Nonnull  timestamp,
    AVAudioFrameCount                       frameCount,
    const AURenderEvent         * _Nullable eventsListHead);
```

I'm just hesitant to do this now because we might end up with something entirely different once we actually implement it.

PR #1581, opened 2018-11-27
AudioEngineUnit

dave234:
AudioEngineUnit is an BufferAudioUnit subclass that contains an internal AVAudioEngine. This allows for a single audio unit that contains multiple audio units within. In this form it is intended to be used as a base class, but there's nothing stopping a client from accessing the (public) internal audio engine and making connections directly.

Subclass and override `shouldAllocateInputBus` to return `false` if using as a generator.


aure:
Looks cool, I went ahead and accepted since I'm about to get on a plane without internet and I want to try it out.  The Travis failure looks like it could just be an anomaly.

PR #1587, opened 2018-11-29
Revert AKAudioUnitBase and AKGeneratorAudioUnitBase to AUAudioUnit base

dave234:
The BufferedAudioUnit base class was causing problems with the [Sampler Demo](https://github.com/AudioKit/SamplerDemo). Reverting until we can find the cause.

PR #1594, opened 2018-12-02
BufferedAudioUnit as base class for AKAudioUnitBase and AKGeneratorAudioUnitBase (again)

dave234:
This is re-introducing the changes originally made in https://github.com/AudioKit/AudioKit/pull/1579. There was an incompatibility with AKSampler where notes would seemingly hang, so the existing base classes were reverted in https://github.com/AudioKit/AudioKit/pull/1587.  

The issue was that the existing base classes zeroed out the output buffer before rendering, whereas the new base class does not. This is addressed by zeroing the output buffer before rendering in AKSamplerDSP->processWithEvents.

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Samplers/Sampler/AKSamplerDSP.mm
+++ b/AudioKit/Common/Nodes/Playback/Samplers/Sampler/AKSamplerDSP.mm
@@ -277,6 +277,14 @@ void AKSamplerDSP::handleMIDIEvent(const AUMIDIEvent &midiEvent)
     }
 }
 
+void AKSamplerDSP::processWithEvents(const AudioTimeStamp *timestamp, AUAudioFrameCount frameCount, const AURenderEvent *events)
+{
+    for (int i = 0; i < _outBufferListPtr->mNumberBuffers; i++) {
+        memset(_outBufferListPtr->mBuffers[i].mData, 0, _outBufferListPtr->mBuffers[i].mDataByteSize);
+    }
+    AKDSPBase::processWithEvents(timestamp, frameCount, events);
+}
+
 void AKSamplerDSP::process(AUAudioFrameCount frameCount, AUAudioFrameCount bufferOffset)
 {
     // process in chunks of maximum length AKCORESAMPLER_CHUNKSIZE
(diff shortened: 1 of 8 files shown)


AudioKit/Common/Nodes/Playback/Samplers/Sampler/AKSamplerDSP.mm
```diff
@@ -277,6 +277,14 @@
     }
 }
 
+void AKSamplerDSP::processWithEvents(const AudioTimeStamp *timestamp, AUAudioFrameCount frameCount, const AURenderEvent *events)
+{
+    for (int i = 0; i < _outBufferListPtr->mNumberBuffers; i++) {
+        memset(_outBufferListPtr->mBuffers[i].mData, 0, _outBufferListPtr->mBuffers[i].mDataByteSize);
```
dave234: @getdunne Here's where I'm zeroing things out.

dave234: Another option would be to change the sampler’s internals so that it does not generate additively.

getdunne: @dave234 I thought about that, but since BufferedAudioUnit is already all about buffer management, it makes sense to zero the output buffer there for two reasons:
1. Additive generation is the most natural approach for all multi-voice instrument AUs.
2. It ensures that a "null" AU (which does no DSP at all) will output silence.

getdunne: @dave234 I will test myself this morning and approve this PR if no problems. BTW thanks for tracking this down.

getdunne: @dave234 I just looked again and realized you're simply adding buffer zeroing to my code. I can do that at a lower level in AKCoreSampler.cpp, with no need to add a processWithEvents() member to AKSamplerDSP.

aure:
This looks good. Any followups and edits can be added later.

PR #1596, opened 2018-12-03
Divide by file duration if endTime - startTime is zero

### LINKED ISSUE #1577: currentTime is nan after loading audio file
Hi all! We have updated AudioKit in our app from version 4.3 to 4.5.4 today and have discovered some interesting behavior around AKPlayer's currentTime property. If we try to get the player's currentTime right after loading a song it is nan, not a number. Previously it was 0. This is occurring in iOS 10-12. I suspect this is not intended behavior. ;)

nevosegal:
When calculating `currentTime`, there is a division by `endTime - startTime` which, before playing the audio file, is 0.0. This fix checks for this and sets it to `duration` if true, as `duration` is set when loading the file.

fixes #1577.


nevosegal:
Looks like the travis failure is unrelated to the PR... Some sort of flaky behaviour. Found this reporting the same issue - https://github.com/Microsoft/azure-pipelines-tasks/issues/8840

aure:
Thank you I love the issue fixing PRs the most!

PR #1600, opened 2018-12-08
Replace void* and UnsafeMutableRawPointer with AKDSPRef.

dave234:
This PR replaces all occurrences of void* and UnsafeMutableRawPointer with AKDSPRef where the pointers is pointing to an AKDSPBase. AKDSPRef is typedef'd to a void* with clang's `swift_newtype` attribute. This offers some stronger typing in Swift as AKDSPRef is recognized as it's own type.

### DIFF
--- a/AudioKit/Common/Internals/Utilities/AKInterop.h
+++ b/AudioKit/Common/Internals/Utilities/AKInterop.h
@@ -11,8 +11,10 @@
 
 #ifdef __OBJC__
 #define AK_ENUM(a) enum __attribute__((enum_extensibility(open))) a : int
+#define AK_SWIFT_TYPE __attribute((swift_newtype(struct)))
 #else
 #define AK_ENUM(a) enum a
+#define AK_SWIFT_TYPE
 #endif
 
 /* EXAMPLE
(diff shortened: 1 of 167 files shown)


AudioKit/Common/Internals/Utilities/AKInterop.h
```diff
@@ -29,4 +31,7 @@
  let direction: AKDirection = .up
 */
 
+/** Pointer to an instance of an AKDSPBase subclass */
+typedef void* AKDSPRef AK_SWIFT_TYPE;
```
dave234: Here's the AKDSPRef definition.

AudioKit/Common/Internals/Utilities/AKInterop.h
```diff
@@ -11,8 +11,10 @@
 
 #ifdef __OBJC__
 #define AK_ENUM(a) enum __attribute__((enum_extensibility(open))) a : int
+#define AK_SWIFT_TYPE __attribute((swift_newtype(struct)))
```
dave234: Here's the attribute adding macro.

dave234:
The changes are largely find and replace except where commented.

PR #1602, opened 2018-12-08
AkAudioUnitBase inheritance improvements

dave234:
Moves AKDSPBase ivar from C++ ivar in implementation file to AKDSPRef property in Base class. This allows subclasses to access the DSP kernel.


aure:
❤️

Issue #1380, opened 2018-05-23
There is no way to access export progress of an AKAudioFile.


cgossain:
In order to present nice looking UI to the user, it is nice to show an progress bar showing the activity. Currently there appears to be no way to access export progress to update my UI when exporting via the exportAsynchronously method on the AKAudioFile.

aure:
That would be nice to have, but there's nothing in place that would make that possible with AKAudioFile that I know.

colinhumber:
@aure The documentation in `AKAudioFile+ProcessingAsynchronously.swift` (https://github.com/AudioKit/AudioKit/blob/2473c7d58d1d731df7cac884a3e0822329889671/AudioKit/Common/Internals/Audio%20File/AKAudioFile%2BProcessingAsynchronously.swift#L278) mentions that the `ExportSession` object that is created to wrap the `AVAssetExportSession` has a `progress` property that can be used to track progress. However, this property is never defined in the internal class.

As @cgossain mentioned the wrapped `AVAssetExportSession` already reports progress, so I'm not sure if the internal documentation is incorrect and there used to be a `progress` property available externally.

```
/// As soon as callback has been triggered, you can use ExportSession.status to
/// check if export succeeded or not. If export succeeded, you can get the exported
/// AKAudioFile using ExportSession.exportedAudioFile. ExportSession.progress
/// lets you monitor export progress.
```

aure:
Yeah, the primary author of this, Laurent Veliscek is no longer working on it, so I am not sure what to make of this, if it was intention it to add it or if was there and someone else clobbered it.  There are a lot of references to an internal export session, maybe this could just be exposed as a property?  Could one of you guys who could use this feature extend AudioKit and test the change in their own project?

aure:
Hmm, anybody still working on or interested in this feature?

ryanfrancesconi:
The AKAudioFile code looks complicated, but there is almost nothing happening in its export code once you get rid of the excess stuff. AKConverter also uses the AVAssetExportSession if necessary. I would be happy to add a progress delegate to that class, but I don't really want to mess with AKAudioFile as it is too limited for macOS usage. AKConverter is a much more flexible conversion class.

however if you want to do your own export, it's pretty easy. Just check the session.progress value periodically for progress values.

`        let asset = AVURLAsset(url: inputURL)
        guard let session = AVAssetExportSession(asset: asset, presetName: AVAssetExportPresetAppleM4A) else { return }

        //session.progress

[...]

PR #1611, opened 2018-12-17
added the currentExportSession class variable to expose progress data…

ryanfrancesconi:
… as per issue: https://github.com/AudioKit/AudioKit/issues/1380

kinda late to the game on this one, sorry. but here's how this would be solved fairly easily.


aure:
Thanks!

PR #1615, opened 2018-12-22
Further cleanup of Bank audio units

getdunne:
At this point, I think I have eliminated all the macros in C++ code. There are still quite a few macros used in Objective-C code.


aure:
Cool!

PR #1618, opened 2018-12-28
AKPlayer.pauseTime, and AudioKit.output -> finalMixer

ryanfrancesconi:
This PR is mainly for the pause bug in AKPlayer. I've also included recent changes I made to AudioKit.output. This may be debatable from certain perspectives, or not. Please look at the alteration of AudioKit.finalMixer?


ryanfrancesconi:
@dave234 requesting review of changes to AudioKit.swift:
https://github.com/AudioKit/AudioKit/pull/1618/commits/53c86bf1e814eaf9188b8b73c89a03a64f5cc805

ryanfrancesconi:
@dave234 this is the part your review is requested for:

Debatable change here, but it is desirable from my point of view:  …
if the assigned output is already a mixer, avoid creating an additional mixer and just use the assigned input as the finalMixer.
This avoids the creation of an extra mixing node and another potential spot for sample rate conversion.

Also made finalMixer an optional for this reason.

I know this is ultimately going to be trashed when your new I/O stuff takes over, but it's how I use AudioKit.output over here....

PR #1622, opened 2019-01-01
Add macOS OSC examples

getdunne:
New example project "HelloOSC" under _Examples/macOS_. Uses Devin Roth's _SwiftOSC_ library.


aure:
So cool!

PR #1642, opened 2019-01-16
Improvements to AKMIDI port selection and sysex message handling

karnlund:
Port selection now uses MIDIUniqueID instead of a String name, so that ports with the same name can be identified correctly.

Sysex received data is now correct, and does not contain up to 256 0’s after each byte. 

Manufacturer and Model were swapped in the EndpointInfo.

AKMIDI code cleaned up a bit.

Added a command line target to the dev project for testing Sysex or any other features.  Included compete definition of Kawai K5000 sysex dump requests.


aure:
Only an example is breaking, this is just too interesting not to merge and play with. Might make some tweaks along the way.

aure:
Thank you!

PR #1646, opened 2019-01-18
1 and 2 byte system messages were not being processed correctly

karnlund:


aure:
I'm going to accept this and then fix the Travis build. Thanks!

PR #1653, opened 2019-01-24
Make AKCoreSynth public

dave234:
I was getting an error when importing `<AudioKit/AudioKit.h>` into an Objective-C++ file.


dave234:
I'm realized that I didn't need to add AKCoreSynth to the umbrella header. I don't really know enough about the implementation to have an opinion on this. I just removed that addition so that there were minimal changes.

PR #1663, opened 2019-02-03
AKPlayer creating loop artifacts in official playground, AKAudioPlayer fixes it

irae:
Hey folks! Noob here! Just started learning AudioKit. For some reason, one of the playgrounds was giving me loop artifacts. I compared the code with the previous Playground and noticed the difference of using `AKPlayer` and `AKAudioPlayer`. Switching to the leather fixes it.

Maybe it is just API differences or some bug, I'll let you guys that know better dig into it. But while you investigate, maybe replace it by the one that works?

### DIFF
--- a/Playgrounds/AudioKitPlaygrounds/Playgrounds/Basics.playground/Pages/Balancing Nodes.xcplaygroundpage/Contents.swift
+++ b/Playgrounds/AudioKitPlaygrounds/Playgrounds/Basics.playground/Pages/Balancing Nodes.xcplaygroundpage/Contents.swift
@@ -12,8 +12,8 @@ import AudioKitUI
 
 //: This section prepares the players
 let file = try AKAudioFile(readFileName: "drumloop.wav")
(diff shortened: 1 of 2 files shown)


irae:
Well, as I went through and played with each playground, I found more examples that didn't quite work. Unfortunately, the playback playground cannot be easily fixed. So maybe this should actually be an issue.

https://github.com/AudioKit/AudioKit/blob/50d75151417fd5a3c008d6876456c11ab3f078f2/Playgrounds/AudioKitPlaygrounds/Playgrounds/Playback.playground/Pages/Audio%20Player.xcplaygroundpage/Contents.swift#L11

I was reading the documentation and realized both classes are not feature compatible. So I am now trying to figure out if this can be fixed in another way, or if we have a bug with current version of `AKPlayer`.

aure:
Nice someone is showing interest in the playground, but you should also be aware that they're on the move - the ones you should probably using are here https://github.com/AudioKit/Playgrounds

PR #1665, opened 2019-02-04
Improving LoopbackRecording Example (closes #1662)

### LINKED ISSUE #1662: Details on AudioKit.start() bug mentioned on LoopbackRecording Example
I am about to start writing a multi-track recorder app and I was studying and exploring AudioKit documentation, Examples and StackOverflow questions related to recording in sync with a track playing. I came across a intriguing piece of code by @dave234 in the `LoopbackRecording` example:

https://github.com/AudioKit/AudioKit/blob/ff43976916a96756aa325a205f3087307e82ad04/Examples/iOS/LoopbackRecording/LoopbackRecording/ViewController.swift#L93-L103

Is there still some kind of problem with `AudioKit.start()` as of AudioKit 4.5.5? Should I be avoiding it for advanced recording scenarios? And lastly, in case this is outdated, could someone already proficient with `AKClipRecorder` and `AudioKit.engine` update the example with current best practices?

Thank you very much for your time (and this awesome open source library, really)

irae:
In #1662 I discussed some of @dave234 comments about avoiding `AudioKit.start()`.

I tested a lot of combinations and turns out there are still some combinations that might cause crashes. But If I replace the direct manipulation of AVAudioSession by AKSettings instead I can reach something stable.

This PR changes from direct access to `AVAudioSession.sharedInstance()` to `AKSettings.session` and users `AudioKit.start()` as pointed as best practices in #1662.

Since I am here, I was wondering if the example could be made better in some other way. For instance, `AKSamplerMetronome` is not documented at all, and looks like the older brother of `AKMetronome`. Those don't have API parity, so I was unsure how to make the example work with the newer and documented metronome. Also imagining if the calculations based on `lastRenderHostTime` could be turned into some abstraction in the official AudioKit APIs. Maybe something like `nextSyncOportunity`? I have some time in my hands, so if someone knowledgable wants to point me in some direction to make the example even better, I am all ears!

### DIFF
--- a/Examples/iOS/LoopbackRecording/LoopbackRecording/ViewController.swift
+++ b/Examples/iOS/LoopbackRecording/LoopbackRecording/ViewController.swift
@@ -71,32 +71,33 @@ class ViewController: UIViewController {
     func setUpAudio() {
 
         do {
-            let audioSession = AVAudioSession.sharedInstance()
-            try audioSession.setActive(true)
-            try audioSession.setCategory(.playAndRecord, mode: .default, options: [.defaultToSpeaker, .mixWithOthers])
+            AKSettings.audioInputEnabled = true
+            AKSettings.defaultToSpeaker = true
+            AKSettings.sampleRate = AudioKit.engine.inputNode.inputFormat(forBus: 0).sampleRate
+            try AKSettings.setSession(category: .playAndRecord)
+
 
             // Measurement mode can have an effect on latency.  But you end up having to boost everything.
             // It's a must if you want accurate recordings.  It turns of the os input processing.
             // Uncomment / Experiment
-            /*
 
-            try audioSession.setMode(AVAudioSessionModeMeasurement)
+            // try AKSettings.session.setMode(AVAudioSession.Mode.measurement)
 
-             */
 
         } catch {
             fatalError(error.localizedDescription)
         }
 
-        // Make connections, using engine.outputNode so that I can skip AudioKit.start()
-        [metronome, player] >>> mixer >>> AudioKit.engine.outputNode
+        // Make connections
+        metronome.connect(to: mixer.inputNode)
+        player.connect(to: mixer.inputNode)
+        AudioKit.output = mixer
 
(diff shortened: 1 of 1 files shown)


Examples/iOS/LoopbackRecording/LoopbackRecording/ViewController.swift
```diff
@@ -71,32 +71,33 @@ class ViewController: UIViewController {
...
 
         } catch {
             fatalError(error.localizedDescription)
         }
 
-        // Make connections, using engine.outputNode so that I can skip AudioKit.start()
-        [metronome, player] >>> mixer >>> AudioKit.engine.outputNode
+        // Make connections
+        metronome.connect(to: mixer.inputNode)
+        player.connect(to: mixer.inputNode)
```
dave234: I think these can just be `node.connect(to: otherNode)`. The connect function will end up using inputNode implicitly.

dave234:
AKSamplerMetronome was originally created to demonstrate AKTimeLine, it's mostly unrelated to AKMetronome. If I were to improve the metronome, I would probably base it off of AKSampler and just do away with AKSamplerMetronome altogether. AKSamplerMetronome is a bit of an anomaly in that it subclasses AVAudioUnitSampler and adds a timing tap to it. While this works, it's harder to extend than other nodes as it uses the built in node, which is somewhat of a black box.

dave234:
WRT `nextSyncOportunity`, I think it is a great idea to generalize the timing mechanisms. I originally hoped that we could use/extend [AKTiming] (https://github.com/AudioKit/AudioKit/blob/develop/AudioKit/Common/Internals/Synchronization/AKTiming.swift) to encapsulate much of this logic. You can see an example of this [here in `syncStart`](https://github.com/AudioKit/AudioKit/blob/68e44d9a2c2ac1333caca1bd7706ccd34831dcd8/AudioKit/Common/Internals/Synchronization/AKTiming.swift#L58). I'm not sure if this is helpful, just putting it on your radar.

PR #1668, opened 2019-02-05
Fix: RecordingNodes playground was unable to play recorded sounds

irae:
The recording example was not able to play the recorded sounds. The player was attached to a file but recorder was attached to no file at all. Attaching both to the same `AKAudiofile` fixes it.

The PR also includes cosmetic changes to buttons, that were feeling "stuck" to a different state.

I am aware that Playgrounds are moving to another repo, but I could not run the other repo at all, so the PR fixes it on the only place I could make it work.


aure:
Thanks, I'll definitely get the changes here over to the new playground if/when that does happen.

aure:
BTW, if you make more pull requests, its preferred practice to make them on the develop branch of AudioKit.  This one didn't have any conflicts, but other changes could. Thanks!

irae:
Sure, I'll try to do it.
The reason I tried to stay on master is because I am still relearning Swift (used to program in Swift 1 when it was released and never again) and also learning AudioKit.

Running the stable release feels better. When something is wrong I can almost 100% believe it is on me and not a delelopment branch bug. 😄

PR #1672, opened 2019-02-08
Add AKMusicTrack+Events extension to MacOS

gertjanjansen:
Hi!

My first ever commit to an open source repository! 🙈 

I noticed the AKMusicTrack+Events extension added by @eljeff was not present in the MacOS distribution of AudioKit. I couldn't find a reason why this was the case so I added and tested it on MacOS. It seems to work fine, but maybe I'm missing something. Please have a look.


eljeff:
Good catch! Thank you

PR #1673, opened 2019-02-10
Small refactor in AKAudioUnitManagerDelegate

oettam:
- Passes the `AKAudioUnitManager` itself in the delegate methods, making it easier to use multiple instances of `AKAudioUnitManager`
- Adds specific types for notifications

I needed this change for a project I'm working on, so I thought it could be helpful for others. 
Of course this breaks compatibility with clients that are already using the `AKAudioManager`.


aure:
Next version of AK will have several API updates so I think this is good to get in now. Thanks!

PR #1681, opened 2019-02-14
Bugfix: MIDI input not working with AKAudioUnitInstrument

oettam:
Hi, I wanted to use the `AKSequencer` with a `AKAudioUnitInstrument`, so I checked the sequencer example, but it wasn't working.

I found these 2 problems:

1. from my understanding, MIDI system command are the ones starting with status byte `f0` and above. But the check at [MIDIPacket+Extensions.swift:15](https://github.com/AudioKit/AudioKit/compare/develop...oettam:bugfix/midi-au-instrument?expand=1#diff-ba2f52e5a7b5247169428fcbbd455441R25) would identify all the packets as system command
1. the MIDI packets coming from an endpoint (set via the `midiIn` property of the `AKMIDIInstrument`) didn't reach the `AKAudioUnitIntrument`


aure:
Thanks!

oettam:
@maksutovic the problem was affecting subclasses of `AKPolyphonicNode`, so `AKMIDICallbackInstrument` should also be fixed

PR #1704, opened 2019-03-12
Add RAII to DSP Kernel

sieren:
Adding RAII to the DSP Kernel so we can
initialise all resources when instantiating subclasses.

ping @aure

### DIFF
--- a/AudioKit/Common/Internals/AKSoundpipeKernel.hpp
+++ b/AudioKit/Common/Internals/AKSoundpipeKernel.hpp
@@ -28,10 +28,23 @@ public:
     //    }
 
     sp_data *getSpData() { return sp; }
+
+    // The default constructor should be deleted,
+    // but we're keeping it to not break the API
+    AKSoundpipeKernel() = default;
+
+    AKSoundpipeKernel(int channelCount, double sampleRate) :
+        AKDSPKernel(channelCount, sampleRate) {
+        sp_create(&sp);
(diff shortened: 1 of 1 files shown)


aure:
Hmm, so this affects only two nodes for some reason.  I will check them out to see if they still sound about the same and the precision tests can be updated.

sieren:
@aure I didnt know we had tests on AK so I never ran them. I can also look into it. (And worst case force the reinitialising of `sp` from the `init()` call - which im right now not doing if theres already an instance).

aure:
Its just AKPinkNoise and AKPluckedString

> PinkNoiseTests
>     ✗ testAmplitude, XCTAssertTrue failed - fe160cb0de91109574515f9bef8e3286
>     ✗ testDefault, XCTAssertTrue failed - f5032eb3fc926f68a73fec67e29c6ac7
>     ✗ testParameterSweep, XCTAssertTrue failed - db9904288c8872a06a87a6d4913eb942
> PluckedStringTests
>     ✗ testDefault, XCTAssertTrue failed - 382cdbd27558fed9c8723ba435cdb4cf

aure:
Actually its just the operation versions which is weird.

sieren:
Yeah, and the root cause is
```        if (sp == nullptr) {
            sp_create(&sp);
        }```
as in repeated calls to init. Mhmmmm.

Issue #1709, opened 2019-03-19
AKKeyboardView init ignores polyphonic parameter


bakeddean:
The AKKeyboardView doesn't do anything with the 'polyphonic' parameter. Passing true has no effect, so you need to do:
```
keyboardView.polyphonicMode = true
```

The initializer currently:

```
/// Initialize the keyboard
@objc public init(width: Int,
                      height: Int,
                      firstOctave: Int = 4,
                      octaveCount: Int = 3,
                      polyphonic: Bool = false) {
    self.octaveCount = octaveCount
    self.firstOctave = firstOctave
    super.init(frame: CGRect(x: 0, y: 0, width: width, height: height))
    oneOctaveSize = CGSize(width: Double(width / octaveCount - width / (octaveCount * octaveCount * 7)), height: Double(height))
    isMultipleTouchEnabled = true
    setNeedsDisplay()
}
```

Since the polyphonicMode property has an observer which calls setNeedsDisplay(), you could replace the ``` setNeedsDisplay()``` in the initializer with ```polyphonicMode = polyphonic```.

bakeddean:
Done.

PR #1710, opened 2019-03-19
Fix for AKKeyboardView initializer ignoring polyphonic parameter.

bakeddean:
Fix for #1709.

- Set polyphonicMode property using the polyphonic parameter in the initializer.
- Use the setNeedsDisplay() in the polyphonicMode property observer.


aure:
Thanks!

Issue #1706, opened 2019-03-14
Changing mic input doesn't actually change it


GeorgeL9:
I am building an app that works with multi-channel mixers. I need to be able to select any number of those inputs and make them into one output. While waiting on the mixer to arrive, I created a small sample project as a proof of concept. 

Unfortunately it doesn't seem that changing the AKMicrophone device does anything. I have 4 inputs available to me, the 3 iPhone X microphones, and my external headphone/mic. If I select the internal mic, it still stays connected to the external mic.

See example project:
https://github.com/GeorgeL9/audiokit-device-bug

aure:
Hopefully the pull request addresses this, let us know if not.

Issue #1532, opened 2018-10-11
Use built-in microphone with headphones connected


firewallsol:
I'm trying to change the internal microphone with the connected headphones but I do not change it

im using        

try! AudioKit.setInputDevice(inputs[1])
        try! mic.setDevice(inputs[1])

I have tried inputs 0,1 and 2 but there is no change.

Thank you

aure:
Noting that this is also on StackOverflow: https://stackoverflow.com/questions/52765236/use-built-in-microphone-with-audiokit-with-headphones-connected but I'll leave this issue open as well. Conversation should happen over on StackOverflow however.

aure:
Hopefully this is now fixed.

Issue #808, opened 2017-03-20
AKMicrophone.setDevice() doesn't seem to work after the fact


pcbeard:
I'm looking for a solution for configuring the preferred input before creating an AKMicrophone instance. Calling `AKMicrophone.setDevice()` doesn't seem to change anything once `AudioKit.start()` has been called. My current workaround is to call `AVAudioSession.setCategory()` before accessing `AudioKit.availableInputs`

     try? AKSettings.session.setCategory(AVAudioSessionCategoryPlayAndRecord, with: .mixWithOthers)
     if let inputs = AudioKit.availableInputs, !inputs.isEmpty {
      if let device  = inputs.last {
        try? AudioKit.setInputDevice(device)
      }
    }

This is needed, because all of the available inputs don't show up unless `setCategory()` has been called already. Ideally, one could call `AKMicrophone.setDevice()` anytime after the audio engine has started, and things would just work.

aure:
I don't really know what to do about this.  What do you think would be a good solution?  It's more like a basic design issue with AVAudioEngine than anything AudioKit could be doing better, isn't it?

pcbeard:
Isn't it possible to detect the configuration change using the AVAudioEngineConfigurationChangeNotification? Changing the input device will likely change the sampling rates, so I think this comment is relevant:

> When the engine's I/O unit observes a change to the audio input or output hardware's
> channel count or sample rate, the engine stops itself (see `AVAudioEngine(stop)`), and
> issues this notification.
> The nodes remain attached and connected with previously set formats. However, the app 
> must reestablish connections if the connection formats need to change (e.g. in an 
> input node chain, connections must follow the hardware sample rate, while in an output only
> chain, the output node supports rate conversion).

So, to properly respond to input routing changes, which may include hardware sample rate changes, we'd need to restart the engine, and potentially tear down all the nodes.

pcbeard:
Apologies, I see that you are already handling route changes in AudioKit.restartEngineAfterRouteChange.

aure:
So, you're making pull requests and you're knowledgeable about the code base, do you think there's anything we can do to address this issue?  If yes, let me know, if not, can we close this issue.  And have I inviited you to slack?  Would you like to be part of the AudioKit Slack?

pcbeard:
I'm working on it. Please do invite me to the Slack. I think the recent change I made to the move the notification handlers to the main queue has fixed most of the issues I've been seeing.

aure:
[...]

Issue #1654, opened 2019-01-24
Unable to use iPhone Speaker as output while bluetooth headset is used as mic


PrinceAbraham:
Currently this is my code.
`        AKSettings.playbackWhileMuted = true
        AKSettings.audioInputEnabled = true
        AKSettings.useBluetooth = true
        let mic = AKMicrophone()
        let session = AKSettings.session
        do {
            try session.setCategory(.playAndRecord, mode: .default)
            try session.overrideOutputAudioPort(.speaker)
            try session.setActive(true)
        } catch {
            print("Couldn't override output audio port")
        }

        AudioKit.output = mic
        try! AudioKit.start()`
Seems like the session is not overriding the audio port from my bluetooth headphones to phone speaker.

aure:
Is this a problem with AudioKit?  It seems like you're using AVAudioSession directly.

PrinceAbraham:
The issue might be that since I'm using a headset I can't playback on a different speaker. Can you confirm this behavior? Solution would be to use a bluetooth mic (only mic), right?

eljeff:
I can't comment on this exact problem, but I have done it reverse - I have been able to use the phone microphone and a bluetooth speaker at the same time.

aure:
Can you provide a sample project for us to reproduce this issue?

PrinceAbraham:
It's the same code as given above. I'm using a bluetooth mic, which for some reason shows up as output, I'm trying to get iput from the mic and play back through a bluetooth speaker

PrinceAbraham:
> I can't comment on this exact problem, but I have done it reverse - I have been able to use the phone microphone and a bluetooth speaker at the same time.

How did you achieve this? Currently my bluetooth speaker cuts off most of the words I say

kingsofteng:
> I can't comment on this exact problem, but I have done it reverse - I have been able to use the phone microphone and a bluetooth speaker at the same time.

Can you please share your code snipped for that?

kingsofteng:
> > I can't comment on this exact problem, but I have done it reverse - I have been able to use the phone microphone and a bluetooth speaker at the same time.
> 
> How did you achieve this? Currently my bluetooth speaker cuts off most of the words I say

@PrinceAbraham  Could you resolve this issue?

BrianBecker:
Does it cut off the beginning of your words or after short pauses? If so this is due to your bluetooth speaker going into bypass mode to save battery. 

[...]

PR #1714, opened 2019-03-24
AudioKit setInputDevice bug fixes [iOS]

bakeddean:
Hopefully fixes issues like #808, #1706, #1532, where users (including myself), cannot change the input device.

- Refactor the AudioKit.setInputDevice(_ input: AKDevice) method. Because of differences in the way AudioKit.inputDevice &  AudioKit.inputDevices instantiate their AKDevice, String comparisons like ```dev.uid == input.deviceID``` are false, and fail to set ```AVAudioSession.sharedInstance().setPreferredInput()```
eg "Built-In Microphone" == "Built-In Microphone Back"
The method also potentially sets the preferred device twice, and is generally a bit hard to follow. Also, let the errors propagate, at the moment they are swallowed for iOS but not mac.

- Added convenience initialiser to AKDevice to take a AVAudioSessionPortDescription. Allows easier instantiation of AKDevices in AudioKit.inputDevice & AudioKit.inputDevices, and is more consistent.

- Added isEqual in AKDevice to make it easier to compare devices.
eg ```cell.accessoryType = (inputDevice == currentInputDevice) ? .checkmark : .none```

- Added example of input device switching to the MicrophoneAnalysis example. Built the AudioKit framework from my changes and tested MicrophoneAnalysis and every seems to be working much better.

Cheers.


aure:
I'll get this passing Travis and then I'll be happy to accept this awesomeness.

bakeddean:
Thanks. Silly me, I didn't realise AVFoundation was iOS only.
I've thought about this overnight, and think this could be nicer still.
Have you done any work on this yet? I could push some more changes if not.

Ta.

aure:
I haven't done anything yet, I'll let you give it a go.

bakeddean:
Thanks.

bakeddean:
Righto. I'm happier with this, see what you think.

**AKDevice**
- Added computed properties for portDescription & dataSource
- Fix the conditional compilation issues eg. #if !os(macOS)

**AudioKit**
- Simplify the setInputDevice() method further.

**AKError**
- Added DeviceNotFound error

**MicrophoneAnalysis**
- Add nav bar and moved Input button to said nav bar
- Improved landscape support
- Code simplification
- Added README & screen shots.

bakeddean:
Hmm. Maybe portDescription & dataSource should be open vars's?

aure:
Do you think you could address issues like this with similar changes, except on output? https://github.com/AudioKit/AudioKit/issues/1654

bakeddean:
I don't have the hardware devices to test his exact problem at the moment. It might be when he sets the output, AudioKit potentially creates a route change. Of overrideAudioOutputPort, Apple says "This change remains in effect only until the current route changes"

I think your saying, don't worry about all this, and implement ```AudioKit.setOutputDevice```, which is not implemented yet on iOS. I was thinking about this last night. Its seems logical that iOS should support multiple output devices. Dongles like the iRig HD2 or even simple USB headset/mic devices plugged into the camera connection kit should present as output options.

At the moment I've only got input devices like the Sonuus i2m, but I was going to get a cheap dongle like this:
s-l225

Hopefully its class compliant.

And see if it comes up as an output device, and implement setOutputDevice and see how things go.

Cheers.

aure:
I'd be happy to reimburse you for an inexpensive device that would allow you to delve further, just let me know.

bakeddean:
I'll give these cheapies a go first, they are only a few dollars. iOS is pretty good with USB audio as long as its class compliant/driverless.

PR #1715, opened 2019-03-25
Made AKMicrophone init a failable initializer

maksutovic:
1. Made `init` for AKMicrophone a failable initializer.
2. Made `setAVSessionSampleRate(sampleRate: Double)` a throwing function

`init` will fail under the following conditions:
- `format` returns nil
- `setAVSessionSampleRate` throws an error

I think the 2nd `setAVSessionSampleRate(sampleRate: Double)` throwing, thus failing the entire init for AKMicrophone might be a little heavy handed, but in any case that there is samplerate mismatch, we should never hard crash our apps, as alot of users (maybe just me 😜 ) seem to have this crash.

Happy to revise in anyway, thanks so much team!


aure:
This PR should have been made to the develop branch, but I recreated it as such.

PR #1694, opened 2019-02-20
Move AKCoreSynth.hpp to project scope in iOS project (#1691)

sparist:
Issue: #1691

PR #1722, opened 2019-03-27
Add missing AKCoreSynth.hpp header to Audiounit.h

sieren:
Should resolve some compiler errors people have been seeing.

Should address: https://github.com/AudioKit/AudioKit/pull/1694#issuecomment-477105302


sieren:
ping @aure

PR #1728, opened 2019-04-01
AudioKit Mac Sandbox issue fix

relpot:
In order to fix the annoying popup alert when using AudioKit for Mac with SandBox,
the method init(type: OSType, subType: OSType) from the AKComponent.swift class file must be updated.
At the line 93, the value "0"  must be replaced by "AudioComponentFlags.sandboxSafe.rawValue".
So at the end, the method init(type: OSType, subType: OSType) should look like that:
```
public init(type: OSType, subType: OSType) {
        self.init(componentType: type,
                  componentSubType: subType,
                  componentManufacturer: fourCC("AuKt"),
                  componentFlags: AudioComponentFlags.sandboxSafe.rawValue,
                  componentFlagsMask: 0)
    }
```


aure:
made the change on develop, https://github.com/AudioKit/AudioKit/commit/6d6b2d317ed0f1c65e161a45db311ace450e3db2

Thanks!

PR #1736, opened 2019-04-08
Fix for broken universal frameworks

megastep:
While the main fix for the issue was basically inverting the order of the arguments to the `lipo` command used to assemble the universal frameworks, this PR also includes a variety of cleanups I made while trying to switch to a source-based pod.

I am glad that the actual fix ended up being that simple (at least for now), because two things became quickly apparent while I was trying to fix all of this:

- CocoaPods is just *not* currently capable of building a project like AudioKit from source. Their build system is just too deficient and simply not able to handle C++ mixed with Objective-C and Swift the way we are. The same thing could be said about the Swift Package Manager, but at least it never attempted to cover anything beyond Swift.
- It is also *not* capable of handling per-platform frameworks the way it should. The fact that we are able to ship a truly "universal" framework with slices for what are in effect separate platforms (`iphoneos` and `iphonesimulator`) is actually an unsupported artifact that a lot of people have been relying on. Apple explicitly says that framework slices should *not* be combined across platforms. This also explains why we had to do some incredible scripting acrobatics when we were trying to ship a dynamic framework.

I will be submitting a feature request to the CocoaPods project to at least try to address the frameworks issue as it is the easier thing to fix here. This suddenly broken behavior with Swift 5 and Xcode 10.2 just brought the problem to the fore and there is no telling when these pseudo-universal frameworks will completely stop working in Xcode. We need to get in front of this problem.

In the meantime, this PR should produce a working pod that should let people use AudioKit and CocoaPods together again.


megastep:
So it looks like while this is actually fixed on the Swift side, when importing the framework in Objective-C it still fails to see the symbols - which is really odd to say the list. So effectively right now the HelloObjectiveC example doesn't build when targeting a device. But all other projects are fine, so it's an improvement.

aure:
What I'll do is merge but not make a release just yet.

colinhumber:
@megastep Out of curiosity, would this fix an issue I just started having on our CI after updating to version 4.7? Seems like there's a missing architecture for `AKConverter`. Our CI is setup to build AudioKit 4.7 with Xcode 10.2 but has been failing since we made the update. However, I can still successfully build on my local dev machine with Xcode 10.2.

```
Undefined symbols for architecture armv7:
  "_swift_updateClassMetadata2", referenced from:
      type metadata completion function for AudioKit.AKConverter in AudioKit(AKConverter.o)
ld: symbol(s) not found for architecture armv7
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

megastep:
@colinhumber if you were just using our pod it should actually help you - missing symbols seem to no longer be a problem when linking, though they still are in Xcode proper when using AK in Objective-C somehow.

I'm pushing a staging build right now so you can try that pod in a little bit.

colinhumber:
@megastep Awesome, I'll give it a go once it's up. Hopefully that solves it. I was super confused when those linking issues started popping up only on our CI. 🤷‍♂️

megastep:
A staging pod for version 4.7.1.b1 is now available. Instructions here: https://github.com/AudioKit/Specs

eljeff:
I can confirm this fixed the issues I was having with CocoaPod and pre-built binary. Things that used to only build in simulator or device now build for both.

colinhumber:
I'm still seeing the armv7 error I noted above on our CI only, but I'm thinking there must be something else going on. I can build and archive locally just fine.

To clarify, the app we're building pulls in one of libraries that has AudioKit as a dependency. The library's pod is set to build as a static library. I'm not really sure why it would fail on CI but work locally, but I think that's out of the scope of what this was fixing, and don't want to confuse the issue too much. I'm going to do some more digging.

colinhumber:
Ugh, so it looks like the version of Xcode 10.2 that Circle CI currently supports isn't actually the GM, which I'm guessing is causing these issues as I'm locally running the GM. Ok, think that's sorted out. Sorry for the brain dump, and thanks for all the hard work!

PR #1743, opened 2019-04-12
New Notification .AKEngineStarted added.

trevor-sonic:
This update mainly adding; AudioKit engine started (and ready use) notification. 
Some times, AudioKit need to be restarted (for AKSession changes or other..) 
```AudioKit.stop()```
making changes ... like session etc..
```AudioKit.start()```
making some other audio node changes can cause crash because AudioKit engine is not ready properly yet. We could add some delay but this is not proper way.

Now by listening; ```.AKEngineStarted``` notification we can insure that audio engine is ready.
````
NotificationCenter.default.post(
                    name: .AKEngineStarted,
                    object: nil,
                    userInfo: nil)
````

Other improvement is about some MIDI related crashes.


aure:
Looks like this still refers to internalData which I think is now just data.  I believe that is the extent of the errors Travis-CI found.

trevor-sonic:
Ok, I'll update and re-request.

Issue #1730, opened 2019-04-02
AKPlayer.currentTime is not accurate when AKPlayer.resume() is called


apprithm:
AKPlayer.currentTime is not accurate when AKPlayer.resume() is called. The currentTIme always start to increment from 0 instead of the previous pauseTime.

It should increment from the pauseTime.

apprithm:
It looks like the following lines are messing up the logic:
In AKPlayer+Playback.swift
```
line 73 startTime = previousStartTime
```

In AKPlayer+Playback.swift
```
line 306  let current = startTime + playerTime.truncatingRemainder(dividingBy: currentDuration)

```

wangrenzhu:
I solved with code like this:

`do {
                try self.bgMusicPlayer = AKAudioPlayer(file: self.backgroundMusicFile!)
                self.bgMusicPlayer.internalPlayer.installTap(onBus: 0, bufferSize: 512, format: self.bgMusicPlayer.audioFile.processingFormat) { [weak self] (pcmbuffer, time) in
                    guard let nodeTime = self?.bgMusicPlayer.internalPlayer.playerTime(forNodeTime: time), let startTime = self?.bgMusicPlayer.startTime else {
                        return
                    }
                    
                    let currentTime = Double(nodeTime.sampleTime ) / nodeTime.sampleRate
                    self?._bgCurrentTime = (currentTime < 0 ? 0.0 : currentTime) + startTime
                    AKLog("currentTime:", self?._bgCurrentTime ?? 0.0)
                }
                self.bgMusicPlayer.volume = self.backgroundVolume
                self.bgMusicPlayer.looping = self.isBgMusicLooping
                self.bgMusicPlayer.startTime = currentTime
            } catch {
                
            }`

Madrox24:
Someone managed how to solve this problem with AKPlayer?

Issue #1675, opened 2019-02-12
AKPlayer reporting wrong startTime after finished playing the track.


apprithm:
Im not sure if this is intended or a bug, but it only happens on 3.6 and the dev branch:

**Reproduce:** set pauseTime to a time other than 0, call resume(), wait till it finish play the whole track. Then call play(). It will start from where it previously paused instead of the beginning.

**Expected behavior:** The startTime should report 0 after the track is finished playing. This will effectively play the track from the beginning.

ryanfrancesconi:
Thanks I'll check!

PR #1746, opened 2019-04-14
Fix for currentTime starting from zero when resuming from pause https…

juusokos:
pauseTime and isPaused live different lives now so pauseTime can be used to give correct currentTime value after pausing. This makes it possible to keep the fix for https://github.com/AudioKit/AudioKit/issues/1675 (startTime is not changed when resuming from pause) while also fixing https://github.com/AudioKit/AudioKit/issues/1730 (currentTime starting from zero when resuming from pause)

The preserved pauseTime is additionally always cleared when the main play() is called so stuff such as seeking to different time don't mess stuff up.


aure:
Thanks!

mglodack:
Thanks @juusokos! I was just experiencing the exact same issue and it's awesome to see there is a fix for it 🎉

PR #1748, opened 2019-04-15
Loop completion handler

jacobp100:
Adds a completion handler specific to looping. A use-case is a speed trainer that increments the tempo on every loop.


aure:
This should have been to develop branch, but I have added it anyway, thanks.

jacobp100:
Thanks 😀

Issue #1757, opened 2019-04-22
Cannot close midi output port


andreiantonescu:
Steps:
- open a midi output port based on a UID
- try and then close the respective midi output port based on the same UID

Result:
- midi output port is not closed, throws error -10832. 

From digging a bit around this issue it seems that the error code above refers to kMIDIWrongEndpointType - 'A source endpoint was passed to a function expecting a destination, or vice versa'.

A sample app that exemplifies this issue can be found here: https://github.com/andreiantonescu/audiokitMidiOutTest. I've built a table view which lists the available output ports and a keyboard from which you can send data. You can see that you can successfully open output ports and send data to them but you cannot close them.

andreiantonescu:
I am not sure if what I am trying to do is supported by AudioKit, but by the naming of the API I would presume this is the case. 

I've submitted a PR in which I describe what I did to fix this issue for my project: https://github.com/AudioKit/AudioKit/pull/1758

eljeff:
> kMIDIWrongEndpointType - 'A source endpoint was passed to a function expecting a destination, or vice versa'.

It sounds like somewhere we were just using the wrong function on the wrong type? Is there a way to actually close the port using the correct arguments?

aure:
Closing since the PR was accepted.

PR #1758, opened 2019-04-22
attempt to fix midi close output port

andreiantonescu:
Attempt to fix https://github.com/AudioKit/AudioKit/issues/1757.

Solution:
- revert to previous way of closing output ports by just removing them from the endpoints listing
- do not close previous output port when opening a new one

### DIFF
--- a/AudioKit/Common/MIDI/AKMIDI+Sending.swift
+++ b/AudioKit/Common/MIDI/AKMIDI+Sending.swift
@@ -173,13 +173,6 @@ extension AKMIDI {
             return
         }
 
-        // close any previous output port and dispose of it
-        if outputPort != 0 {
-            let uid = getMIDIObjectIntegerProperty(ref: outputPort, property: kMIDIPropertyUniqueID)
-            closeOutput(uid: uid)
-            outputPort = 0
-        }
-
         outputPort = tempPort
 
         let destinations = MIDIDestinations()
@@ -200,7 +193,7 @@ extension AKMIDI {
         }
     }
 
-    /// Open a MIDI Input port by name
+    /// Close a MIDI Input port by name
     ///
     /// - Parameter name: Name of port to close.
     ///
@@ -212,7 +205,7 @@ extension AKMIDI {
         closeInput(uid: uid)
     }
 
-    /// Open a MIDI Input port by index
+    /// Close a MIDI Input port by index
     ///
     /// - Parameter index: Index of destination port name
     ///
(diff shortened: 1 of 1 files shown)


AudioKit/Common/MIDI/AKMIDI+Sending.swift
```diff
@@ -230,24 +223,11 @@ extension AKMIDI {
     ///
     public func closeOutput(uid outputUid: MIDIUniqueID) {
         let name = destinationName(for: outputUid)
-        AKLog("Closing MIDI Input '\(String(describing: inputName))'")
+        AKLog("Closing MIDI Output '\(String(describing: name))'")
         var result = noErr
         if let endpoint = endpoints[outputUid] {
-            result = MIDIPortDisconnectSource(outputPort, endpoint)
```
eljeff: 'Source' to me sounds like it would be an input, so that makes sense if this is throwing the 'wrong type' error. Can you find the correct function to use (if there is one) to close the output port properly? @karnlund what do you think?

AudioKit/Common/MIDI/AKMIDI+Sending.swift
```diff
@@ -230,24 +223,11 @@ extension AKMIDI {
...
-            if result == noErr {
                 endpoints.removeValue(forKey: outputUid)
-
-                result = MIDIPortDispose(outputPort)
```
eljeff: Maybe this function still passes? Can you try this keeping this line intact and just skip the MIDIPortDisconnectSource?

andreiantonescu: This function passes but it seems to close all midi outputs altogether. I.e. if I have the following situation:
- Output 1 opened
- Output 2 opened
and then close Output 1, midi will not be sent to Output 2 anymore because of this (or any other output if you have more). 

From what I experimented, it seems that the outputPort variable functions as a sort of global midi output port. It just has to be opened once for the first midi output and then when you attempt to open subsequent midi outputs it suffices to add them to the endpoints listing and midi will be sent to them.

Therefore, when you open a new midi output, I've seen that you don't necessarily have to assign a new value to outputPort via tempPort = MIDIOutputPort(client: client, name: outputPortName), unless of course the outputPort hasn't been initialized yet.

Referring strictly to the MIDIPortDispose part you talked about above, I would use it only when there are no more endpoints left in the endpoints listing, i.e. when you have closed all midi output ports.

karnlund: That's correct @andreiantonescu.  I find it often helpful to refer to this diagram I drew of CoreMIDI:

https://github.com/AudioKit/AudioKit/blob/master/docs/CoreMIDI%20Design.svg

You also should be careful to not dispose of physical ports that you don't create.

andreiantonescu: Thanks for the info @karnlund, I've pushed a new commit to address this.

andreiantonescu:
Regarding the latest commit, besides the items discussed with @eljeff & @karnlund above, I've also noticed and fixed some other inconsistencies.

To sum it up, it seemed that the close & open output functions by index/name were either having incorrect comments or using incorrect functions / arrays:
- the open output by name function had comments referring to destination index instead of destination name
- the open output by uid function had comments referring to destination name instead of destination uid
- the close output by name function had comments referring to closing an input by name instead. On top of this, it was looking for the uid in inputUIDs and not in destinationUIDs / through uidForDestinationAtIndex
- the close output by index function had comments referring to closing an input by index instead. Similar to the above point, it was looking for the uid using uidForInputAtIndex vs uidForDestinationAtIndex

karnlund:
Most of those sound like typical me, copy and paste errors from when I switched us to uid's instead of names.  Sounds like a good clean up.

PR #1773, opened 2019-05-10
AKLog: realized it was printing Optional() on every content item

ryanfrancesconi:
🤷🏽‍♂️fixed.

### DIFF
--- a/AudioKit/Common/Internals/AudioKitHelpers.swift
+++ b/AudioKit/Common/Internals/AudioKitHelpers.swift
@@ -89,7 +89,13 @@ public func fourCC(_ string: String) -> UInt32 {
 public func AKLog(fullname: String = #function, file: String = #file, line: Int = #line, _ items: Any?...) {
     guard AKSettings.enableLogging else { return }
     let fileName = (file as NSString).lastPathComponent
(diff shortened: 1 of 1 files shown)


AudioKit/Common/Internals/AudioKitHelpers.swift
```diff
@@ -184,8 +188,8 @@ extension Double {
             // exponential taper
             var adjustedMinimum: Double = 0.0
             var adjustedMaximum: Double = 0.0
-            if range.lowerBound == 0 { adjustedMinimum = 0.000_000_000_01 }
-            if range.upperBound == 0 { adjustedMaximum = 0.000_000_000_01 }
+            if range.lowerBound == 0 { adjustedMinimum = 0.00_000_000_001 }
```
aure: Why this change? Seems wrong to me.

ryanfrancesconi: oh, the swift lint did that. feel free to revert.

aure: I actually realized that it was correct this way, which is why I accepted the PR as is rather than ask for a change.

PR #1779, opened 2019-05-14
AKDynamicPlayer: Fixed a bug where the timePitch node would be removed if you disabled…

ryanfrancesconi:
… either rate or pitch, even though the other variable pitch or rate is still on


aure:
Thank you!

PR #1785, opened 2019-05-27
Add MIDIUniqueID to AKMIDIListener callbacks

cemolcay:
Adds `portID` parameter to all `AKMIDIListener` callbacks.
Allows implementor to know which port has received the MIDI message.


aure:
Do. you think it would make sense to make the port ID be an optional parameter with default value nil so that this becomes a non-API breaking change?

cemolcay:
@aure, yup, makes sense, I made them optional with default nil value.

PR #1789, opened 2019-06-12
Allow paths with whitespaces for samples in SFZ loader

iamluciano:
The following SFZ line example (1):

`<region> loop_mode=loop_continuous loop_start=3510 loop_end=3984 sample=samples/Mizmar A-Sha3bi-1.wav `

Was incorrectly interpreted as:

`<region> loop_mode=loop_continuous loop_start=3510 loop_end=3984 sample=samples/Mizmar
`

This patch allows referencing samples with whitespaces in their paths (like 1).


aure:
I'm happy to accept this but there are two versions of this sfz loader floating around now and its kind of annoying. They need to be merged.

PR #1793, opened 2019-06-21
added method for computing master set from 12et and deviations in cents

marcussatellite:
added method for computing master set from 12et and deviations in cents


marcussatellite:
@aure don't merge this yet...need feedback from Luciano Lam

iamluciano:
@marcussatellite I confirm this patch works great and allows creating a tuning table from a cents array very easily 👍🏻

Issue #1813, opened 2019-07-23
AKMicrophone has to be created before AudioKit.output is assigned


tcwalther:
This may be expected behaviour, but if so, it should be mentioned more prominently in the documentation:

When instantiating an `AKMicrophone` *after* assigning `AudioKit.output`, it only records zeros (i.e., silence). It took me a fair amount of debugging to figure this out. Simply moving
```swift
let mic = AKMicrophone()!
```
before 
```swift
AudioKit.output = ...
```
fixed it.

I had wrongly assumed that the order of these operations doesn't matter, as long as they all happen before `AudioKit.start()`.

This is on the most recent 4.8 release.

aure:
Yeah, this is well known and an unfortunate side effect of so much happening when you set the output. In the future, we're architecting a system that will be less sketchy.

tcwalther:
Thank you for reacting to my issue so swiftly, I really appreciate it.

If it is well known, I suggest adding a comment in the playgrounds that explains this. It took me almost two days to figure this out; a small comment in the code might save the next person some time :-).

aure:
Good suggestion, might you like to make a pull request doing it? I like it when the affected party makes the fix! :)

tcwalther:
Yes, of course. I just filed a pull request.

I'm having some more issues right now with zero-valued samples coming from AKMicrophone when headphones are connected, but I will try to get to the bottom of that first and will then file a separate issue (or, if I can, a PR).

tcwalther:
I fixed my other issue. I didn't have a fully connected graph; my microphone wasn't routed to the output.

Interestingly, that worked okay for the internal mics but not for the headphones. @aure if you have a suggestion where this kind of mistake should best be explained, happy to make another PR.

PR #1835, opened 2019-08-16
Explain that AKMicrophone nodes have to be created before setting AudioKit.output

tcwalther:
Adds a short comment explaining that `AKMicrophone` instances always have to be created before `AudioKit.output` is assigned. See #1813.


aure:
Thanks!

Issue #1319, opened 2018-04-10
Question: Echo cancellation no present in AudioKit?


temkich:
Echo cancellation no present in AudioKit? 
Can you advise how to solve the echo problem? 
What are the solutions?

aure:
Inactionable, perhaps try StackOverflow.

mlostekk:
any new ideas about a solution?

PR #1838, opened 2019-08-19
added echo cancellation to EZMicrophone & AKSettings

mlostekk:
Hi,

i was thinking if this would not solve the requirement for echo cancellation? This is the only flag that actually has to be set

It would also resolve this request. (and we are planning to use echo cancelation with AK too)

https://github.com/AudioKit/AudioKit/issues/1319

### DIFF
--- a/AudioKit/Common/Internals/AKSettings.swift
+++ b/AudioKit/Common/Internals/AKSettings.swift
@@ -99,6 +99,11 @@ open class AKSettings: NSObject {
     @objc public static var bluetoothOptions: AVAudioSession.CategoryOptions = []
     #endif
 
+    #if !os(macOS)
+    /// Enable / disable voice processing (echo canellation)
+    @objc public static var enableEchoCancellation: Bool = false
+    #endif
+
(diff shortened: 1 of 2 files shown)


aure:
Cool, I hope this helps you and others out.

mlostekk:
@aure do you have examples out there to use EZMicrophone in a swift based project?

aure:
@mlostekk Other than how we use it in AK?

mlostekk:
Well, i see that it can simply be created and then its already running. it does does not have to be attached to AudioKit as AKNode. So my case is solved. But anyhow, how do you attach those `EZxxx` devices as they are not a `AKNode`s?

aure:
The AKMicrophoneTrackerEngine uses the EZMicrophone directly.  Also, AudioKit uses EZAudioDevice optionally for device management. AudioKit never meant to wrap all of EZAudio's capabilities, but we do make use of it and we embed it in AudioKit for those reasons.

mlostekk:
okay, thanks for the clarification

PR #1842, opened 2019-08-22
Timestamp for EZMicrophone input

mlostekk:
Simple extension of the protocol of EZMicrophone to pass down the incoming timestamp


aure:
hope this helps you and others.

PR #1846, opened 2019-08-27
Low pass resonant filter added to generator banks

colinhallett:
Added a moog ladder lowpass filter and filter envelope to the generator banks from the DSP to the swift bits. 

The banks now have extra parameters for - 
Filter cutoff frequency 
Filter resonance
Filter ADSR
Filter envelope strength

### DIFF
--- a/AudioKit/Common/Internals/AudioKit.h
+++ b/AudioKit/Common/Internals/AudioKit.h
@@ -26,6 +26,7 @@ FOUNDATION_EXPORT const unsigned char AudioKitVersionString[];
 #import "AKDSPKernel.hpp"
 #import "AKSoundpipeKernel.hpp"
 #import "AKBankDSPKernel.hpp"
+#import "AKFilterSynthDSPKernel.hpp"
 #import "ExceptionCatcher.h"
 
 // Analysis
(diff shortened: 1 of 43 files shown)


colinhallett:
Yeah I agree, it'd good to leave a basic version untouched and have a more extended synth as a separate class - especially if further things are added. Would it then make sense to have a new class instead of AKBankKernel, so the original osc banks don't need to inherit the additions, or does it make no difference?

colinhallett:
I've added a new lot of synths named AKFilterSynth, added headers and changed the AKBanks back to what they were. Looks like it's all compiling and working

colinhallett:
I've also just added in a filter LFO with depth and rate parameters. Probably needs a little bit of tweaking to make match the values with actual the actual LFO value given but it sounds alright. 

Also, if the ramp duration is set to 0.5s the cutoff takes forever to reach its given value. Does this having something to do with the difference between setImmediate and setUIValue in the parameter ramper?

colinhallett:
Also think it'd be worth having a base class for all the .swift files as all the synths mostly have the same parameters and would save a lot of copy pasting when adding features.

aure:
I need to fix a couple of things but I thought I would accept it and move on to the fixes myself. (Mostly its macOS and tvOS support, and maybe to write some tests)

aure:
Okay, I pushed a version with fixed OSes and moved the Filter Polysynths folder into Polysynths since it is a category of poly synth in a way.

colinhallett:
Sounds good! I'll be doing more work with it so will continue to update when and if required.

PR #1856, opened 2019-09-11
Node Automation, Offline Rendering for AKPlayer, and other things

ryanfrancesconi:
The two main additions here are offline rendering for AKPlayer support and AKParameterAutomation. 

In my experiments I ended up touching many files though the main additions are those two things.


aure:
Booster tests fail which is probably expected.

ryanfrancesconi:
What is needed for it not to fail?

PR #1857, opened 2019-09-11
AKPlayer Scheduling bug in realtime vs offline

ryanfrancesconi:
AKPlayer still has a strong sense of hostTime scheduling when realtime playback is happening, so this is currently needed.

```
        if renderingMode == .offline {
            let nowTime = AVAudioTime(hostTime: refTime, sampleTime: 0, atRate: sampleRate)
            avTime = nowTime.offset(seconds: scheduledTime)
        } else {
            avTime = AVAudioTime.secondsToAudioTime(hostTime: refTime, time: scheduledTime)
        }
```


aure:
Pull request on wrong branch, hence 124 changed files.

PR #1871, opened 2019-09-19
added documentation

mlostekk:
I had to debug the framework just to find out that for proper scheduling the player may not be in the paused state. 
Just added some comments


aure:
Thank you, every bit helps and I love documentation pull requests!

PR #1881, opened 2019-10-03
Leak fix for AKTimelineTap

ryanfrancesconi:
the tap was leaking it's renderTap + block. this is my solution to fix that, but Dave may want to do it another way. ARC wasn't collecting it.

Some other debug output cleanup / commenting out of printf's in AKAutomation stuff


aure:
Thanks!

PR #1882, opened 2019-10-06
Leak fix for AKTimelineTap

ryanfrancesconi:
Correct fix included for leak in the timeline tap. Render flags was being kept around. Able to remove the detach now and ARC picks it up.


aure:
Good plumber!

ryanfrancesconi:
😎

PR #1923, opened 2019-11-21
Fix: Make AKMIDIEvent handle MIDI Clock messages

### LINKED ISSUE #1822: AKMIDIListener not receiving Clock events
I have problems receiving MIDI Clock events fron an external device (MC-909) into my iOS App.

I have a simple MIDIListener:
```swift
class MyMIDIListener: AKMIDIListener {
    func receivedMIDINoteOn(noteNumber: MIDINoteNumber, velocity: MIDIVelocity, channel: MIDIChannel, portID: MIDIUniqueID? = nil, offset: MIDITimeStamp = 0) {
        print("Note ON")
    }
    
    func receivedMIDISystemCommand(_ data: [MIDIByte], portID: MIDIUniqueID? = nil, offset: MIDITimeStamp = 0) {
        print("System Command")
    }
}
```

When I trigger notes on the external device, I get "Note ON" messages in the log. So the setup is working. And when I start or stop the external sequencer, I get "System Command" messages. But the problem is, that I get no continuous clock events during playback. Only one event on start and one event on stop of the sequencer.

If have tried with another app ("Midi Tool Box") and the clock events do show up there. Just not through AudioKit.

I am using AudioKit version 4.8.

Not sure, if this may be related to #1798.

mopotmo:
Fixes #1822 

Since Clock events contain no additional data bytes, they were not interpreted as `AKMIDISystemCommand`.

### DIFF
--- a/AudioKit/Common/MIDI/AKMIDIEvent.swift
+++ b/AudioKit/Common/MIDI/AKMIDIEvent.swift
@@ -66,7 +66,7 @@ public struct AKMIDIEvent: AKMIDIMessage {
     public var command: AKMIDISystemCommand? {
         //FIXME: Improve this if statement to catch valid system reset commands (0xFF)
         // but ignore meta events (0xFF, 0x..., 0x..., etc)
-        if let statusByte = data.first, (statusByte != AKMIDISystemCommand.sysReset.rawValue && data.count > 1) {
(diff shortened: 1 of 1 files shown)


aure:
Thanks!

PR #1924, opened 2019-11-24
AKPlayer: fallback processingFormat

ufoq:
AVAudioFormat will fail for pretty much any channel
count that is >2 when no channel layout is given.
It's better to fallback to non-standard format from
audio file.

I tested it with 6 channels file from there: https://www2.iis.fraunhofer.de/AAC/ChID-BLITS-EBU-Narration.mp4

It failed even with simplest playback like:
```swift
file = try! AKAudioFile(readFileName: "test.mp4")
player = AKPlayer(audioFile: file)
player.buffering = .always
AudioKit.output = player
try! AudioKit.start()
player.play()
```

Because `processingFormat` was returning null it couldn't connect nodes and was infinitely calling these lines:
https://github.com/AudioKit/AudioKit/blob/b28526ad8c103947676ab9cb8a2f0c7d6cd671a4/AudioKit/Common/Nodes/Playback/Players/Player/AKPlayer%2BPlayback.swift#L167-L169

What is purpose of that helper method anyway? Why cannot we just use format that was read from file?

### DIFF
--- a/AudioKit/Common/Nodes/Playback/Players/Player/AKPlayer.swift
+++ b/AudioKit/Common/Nodes/Playback/Players/Player/AKPlayer.swift
@@ -198,8 +198,14 @@ public class AKPlayer: AKAbstractPlayer {
 
     @objc public var processingFormat: AVAudioFormat? {
         guard let audioFile = audioFile else { return nil }
-        return AVAudioFormat(standardFormatWithSampleRate: audioFile.fileFormat.sampleRate,
+        let format = AVAudioFormat(standardFormatWithSampleRate: audioFile.fileFormat.sampleRate,
                              channels: audioFile.fileFormat.channelCount)
+        if format == nil {
+            // fallback
+            return audioFile.processingFormat
+        }
+
+        return format
     }
 
     // MARK: - Public Options


ufoq:
Looks like CI issue:
```
Ineligible destinations for the "iOSTestSuite" scheme:

		{ platform:iOS, id:dvtdevice-DVTiPhonePlaceholder-iphoneos:placeholder, name:Generic iOS Device }

		{ platform:iOS Simulator, id:dvtdevice-DVTiOSDeviceSimulatorPlaceholder-iphonesimulator:placeholder, name:Generic iOS Simulator Device }

The command "./Tests/travis.sh" exited with 100.
```

aure:
Yeah, the tests are basically succeeding, just wt to get @ryanfrancesconi 's review

ryanfrancesconi:
@ufoq it may be problematic to allow processingFormat to be > 2 channels for AKPlayer. processingFormat is the basis for the audio unit chain. I will take a look!

ryanfrancesconi:
Actually I think it is fine to just use the audioFile?.processingFormat as you originally asked:

    @objc public var processingFormat: AVAudioFormat? {
        return audioFile?.processingFormat
    }

If you don't mind, I have a PR in progress here, so I will change this on my end.

Thanks for the heads up!!

ufoq:
@ryanfrancesconi 
Sure, you can handle it in your PR.
Also what did you meant by the fact that 2+ channels in processingFormat may be problematic?
I've seen in the codebase that channels stuff is handled in pretty much generic where, I didn't encounter hardcoded parts with `if stereo/mono`.

ryanfrancesconi:
@ufoq i was concerned some AU's could have issue with 6 channels, but I did a few quick tests here and it seems fine.

ryanfrancesconi:
@aure you can close this one.

Issue #1851, opened 2019-09-04
crash format. SampleRate == hwFormat. SampleRate


cs571393:
When I start AKMicrophone() initializing, required condition is false: format. SampleRate == hwFormat. SampleRate leads to crash.
I want to ask whether I need to modify the AudioKit. Engine. InputNode. InputFormat (forBus: 0). SampleRate. Or how to solve the problem

        AKSettings.sampleRate = 44100
        AKSettings.bufferLength = .veryLong
        do {
            if #available(iOS 10.0, *) {
                try AKSettings.setSession(category: .playAndRecord, with: [.allowBluetoothA2DP, .allowBluetooth, .defaultToSpeaker])
            } else {
                try AKSettings.setSession(category: .playAndRecord, with: [.allowBluetooth, .defaultToSpeaker])
            }
        } catch {
            AKLog("Could not set session category.")
        }
        
        AKSettings.defaultToSpeaker = true
        AKSettings.useBluetooth = true
        AKSettings.playbackWhileMuted = true
        AKSettings.audioInputEnabled = true
        
        self.mic = AKMicrophone()

cs571393:
AudioKit.engine.inputNode.inputFormat(forBus: 0).sampleRate 
value is 16000

roussetjc:
@cs571393 Do you use any external devices with a 16K sample rate?
I am struggling with the microphone sample rate crash for quite a long time now too. My app connects to different USB microphones with different sample rates (48K, 96K or 192K) and I still can't make this work seamlessly when plugging/unplugging.
There seems to be something wrong with `AudioKit.deviceSampleRate` which is not responding to any kind of sample rate setting/formatting.

cs571393:
@roussetjc Yes，My devices is 16K sample rate.

roussetjc:
Does it work with `AKSettings.sampleRate = 16000`?

warpling:
I often see the same crash when switching to AirPods

roussetjc:
When an AKMicrophone is initialized, the function `@objc public init?(with format: AVAudioFormat? = nil)` is called.
Let's say we pass a format with a 16K sample rate. Everything is fine until...
```
#if os(iOS)
// we have to connect the input at the original device sample rate, because once AVAudioEngine is initialized, it reports the wrong rate
do {
    try setAVSessionSampleRate(sampleRate: AudioKit.deviceSampleRate)
} catch {
    AKLog(error)
    return nil
}
```

[...]

PR #1929, opened 2019-12-02
AKMicrophone fix per #1851

warpling:
Thank you to everyone who participated in the #1851 discussion, especially @eofs who seems to have found the straight forward golden ticket of a solution: let the system handle it!

This fix has been tested in production on 10-30k devices running iOS 10-13 with a 100% drop off of this previously frequent crash (see #1851). Your mileage may vary!

### DIFF
--- a/AudioKit/Common/Nodes/Input/AKMicrophone.swift
+++ b/AudioKit/Common/Nodes/Input/AKMicrophone.swift
@@ -46,24 +46,8 @@
 		AKSettings.audioInputEnabled = true
 
 		#if os(iOS)
-		// we have to connect the input at the original device sample rate, because once AVAudioEngine is initialized, it reports the wrong rate
-		do {
-			try setAVSessionSampleRate(sampleRate: AudioKit.deviceSampleRate)
-		} catch {
-			AKLog(error)
-			return nil
-		}
-
 		AudioKit.engine.attach(avAudioUnitOrNode)
 		AudioKit.engine.connect(AudioKit.engine.inputNode, to: self.avAudioNode, format: format ?? formatForDevice)
-
-		//Now set samplerate to your AKSettings sampling rate, it may be heavy handed to make the init fail here, but taking all percautions to avoid all the hard crashes with AKMicrohpone init issues of late.
-		do {
-			try setAVSessionSampleRate(sampleRate: AKSettings.sampleRate)
-		} catch {
-			AKLog(error)
-			return nil
-		}
 		#elseif !os(tvOS)
 		AudioKit.engine.inputNode.connect(to: self.avAudioNode)
 		#endif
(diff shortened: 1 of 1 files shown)


eljeff:
I love how simple this is. What have you done to test it, or what could we do to test it? We go back and forth on this mic stuff so frequently, often what fixes it for one person breaks it for another, so I want to check for regressions

aure:
I'm going to accept this due to the high amount of benefit it gave to the crashing in BlackBox.

warpling:
@Eljeff right?? 

I’d like to clarify that the crash actually has still occurred after this change but at such a dramatically reduced rate that I missed it the first time I checked. AKMicrophone crashes in general have dropped by over 90%. The most common crash still being `required condition is false: IsFormatSampleRateAndChannelCountValid(hwFormat)` from `AKMicrophone.init()`


To test that this change was effective I loosely tested that the microphone could survive the following:
- Siri
- music (started before and after app launch and mic init)
- Incoming phone calls
- Mute switch
- Backgrounding
- Voiceover 
- Music playing over AirPods before app launch
- AirPods connected after app launch, before and after mic init

eljeff:
That's awesome @warpling thank you. What I find usually shows these problems, not necessarily crashing, is something like trying to make a recording while using the mic mixed with other AK generators

maksutovic:
@warpling this has been plaguing me for at least a year, I can't thank you enough for this fix <3

warpling:
@Maksutovic tell me about it!! At least a year if not more of realllly embarrassing crash-free user percentages here haha. All credit goes to @eofs though. I just made the PR!

eofs:
👋  Glad to hear that my workaround helped you guys too :)

> I’d like to clarify that the crash actually has still occurred after this change but at such a dramatically reduced rate that I missed it the first time I checked. AKMicrophone crashes in general have dropped by over 90%. The most common crash still being `required condition is false: IsFormatSampleRateAndChannelCountValid(hwFormat)` from `AKMicrophone.init()`

Same here! We've seen two crashes identical to you but otherwise 99.99% drop in crashes. Haven't had the time to investigate further though.

eljeff:
@warpling @eofs @aure As predicted, this re-introduced bugs when using the microphone and recording a file at the same time.
We've gone back and forth on this mic stuff so many times, one person's fix is another person's glitch.
I'm gonna see what I can do to get around it, but I just want to let yall know we are not out of the weeds yet.
If anybody does end up testing this and re-creating the problem, it'd be good to have a simple, common ground test we could all use.

warpling:
@eljeff classic. Sorry this was rammed through. Let's open a new issue with the offending error and whatnot and get a discussion going there!

aure:
Fixing Ryan's crashes is pretty important and the code is simpler. The discussion lasted on the Issues page for months, so I hardly think of this as rammed through.  Perhaps instead of going back and forth we can have two microphone classes. The regular one kept simple and another one that attempts to match the sampling rate or whatever it needs to do above and beyond. The more complex one could inherit from the simpler one if that will help.

PR #1935, opened 2019-12-09
Basics playground typo fix

paulthorsteinson:
While going through the basics playground I found a typo and this simply addresses that.


aure:
Thanks.

PR #1946, opened 2020-01-04
Start to replacing print based AKLog with os_log based one

aure:
os_log seems to be the best way to do logging and is quite robust and supported by Apple:

https://developer.apple.com/documentation/os/logging

Question: In the last line of that file they suggest not putting in symbols and line numbers, which have traditionally done. But, I am not sure why its bad or how the log shows that anyway (some seem to suggest that it doesn't work even on Catalina). 

@ryanfrancesconi I know you're the most interested in updating logging, aside from myself. What do you think?


ryanfrancesconi:
i implemented it on my end and i kept the basic output syntax of AKLog. I don’t think we should change that as that’s what people are used to. Or we could make it an option in the logging class.

I’m personally not a fan of a top level upper case function like AKLog(). I’d personally rather see AKLog.trace() or something like that. But, in general your changes look ok to me. I’ve switched now to my own logging system anyways, but I’m glad for AK internal stuff to use os_log.

For reference, I did it like this:

//
//  Log.swift
//
//  Created by Ryan Francesconi on 12/1/19.
//  Copyright © 2019 Audio Design Desk. All rights reserved.
//

import os

private let subsystem = Bundle.main.bundleIdentifier ?? "com.audiodesigndesk.*"

struct Log {
    static var isEnabled: Bool = true
    static let errorLog = OSLog(subsystem: subsystem, category: "Errors")
    static let debugLog = OSLog(subsystem: subsystem, category: "Debug")

    static func error(fullname: String = #function, file: String = #file, line: Int = #line, _ items: Any?...) {
        // guard isEnabled else { return } // always print errors currently

        let fileName = (file as NSString).lastPathComponent
        let content = (items.map {
            String(describing: ($0 ?? "nil"))
        }).joined(separator: " ")
        let message = "\(fileName):\(fullname):\(line):\(content)"

        os_log("%{public}@", log: errorLog, type: .error, message)
    }

    static func debug(fullname: String = #function, file: String = #file, line: Int = #line, _ items: Any?...) {
        guard isEnabled else { return }

        let fileName = (file as NSString).lastPathComponent
        let content = (items.map {
            String(describing: ($0 ?? "nil"))
        }).joined(separator: " ")

        let message = "\(fileName):\(fullname):\(line):\(content)"

        os_log("%{public}@", log: debugLog, type: .debug, message)
    }
}

aure:
Yeah, I think we're close enough in spirit, and I left the AKLog function just so I wouldn't have to change 700 lines of code that already use AKLog(). I'll go ahead and commit this to develop since its already pretty far along.

aure:
Oops, I mean I'll commit the other pull request.

ryanfrancesconi:
cool sounds good,

PR #1951, opened 2020-01-12
Preliminary support for XCFrameworks, enable frameworks for distribution

megastep:
This lays the groundwork for generating true universal XCFrameworks that contains every single build of AudioKit (iOS, macOS, Apple TV and Catalyst).

This is not complete because to make this actually functional we'll need to rename the `AudioKit` class to something that doesn't clash with the name of the enclosing module. This has large implications needing to patch most examples, playgrounds, tests and docs. It also breaks the API (although in relatively minor way) so it might be better left off for the next major release (AK 5 maybe?)

This is still helpful for the current release because:
- We now generate frameworks with the library evolution feature turned on, which in theory should now solve the issue with binaries becoming incompatible with future versions of Swift
- Fixed a few small things that were getting in the way of successfully building the XCFramework (more importantly several bugs in the compiler and other quirks)

I think the rest of the work to complete XCFramework support should go in a beta branch of some sort, particularly since a lot of the usefulness of this hinges on other things being still pending release: Travis support for Catalina, CocoaPods 1.9 (still in beta) with native XCFramework support. Generally the tooling support even from Xcode still feels pretty "beta".

I have the rest of this ready to commit (and it works even with the new Catalyst examples) but don't think it should to the current develop branch.


megastep:
Unless there are objections, I think we should merge this PR as is for now for the benefits we get from library evolution being turned on. We can add the rest for the XCFrameworks later (mostly all the big changes with renaming the AudioKit singleton) when CocoaPods/Travis have caught up.

aure:
You are the release czar, so if you want to make this change, I approve.

PR #1969, opened 2020-02-20
add abillity to monitor amount that AKDynamicRangeCompressor is working

osheroff:
Hi, 
I've been working on an audiokit based project that includes AKDynamicRangeCompressor as a last-chance limiter.  I want to be able to see how much the compressor is working and feed that back to the user.  here's a patch that does this.

Questions:

- is averaging the amount of compression performed correct?  would RMS be better?  max amount of compression per buffer?
- would you prefer to output this value in DBs or as a float-point?

thanks,
osheroff


aure:
In response to your questions, dB might be nice, but in no way should that hold up the acceptance of the pull request. Seems like its been a while since I've received a pull request like this from a new person. Thank you!

PR #1981, opened 2020-03-07
Replace Travis CI with GitHub Actions

megastep:
This switches AudioKit fully to GitHub Actions for CI and builds.

Many advantages here:
- It builds and tests all 3 platforms in parallel, allowing us to run all available tests and still complete jobs much faster than Travis.
- The macOS images are much more up to date and are able to build Catalyst binaries and produce XCFrameworks.
- Build artifacts are saved for each run, so you're able to easily get the latest frameworks straight from GitHub instead of building them yourself.

Works just like before: merging to the staging branch now uploads only the XCFramework to S3 (will work on switching CocoaPods to that next), tagging triggers a new release with all the files attached.

(I suggest squashing my commits when merging as there was a lot of trial-and-error going on there)


megastep:
Example of a run: https://github.com/catloafsoft/AudioKit/actions/runs/51328416

PR #2022, opened 2020-04-08
Cleaner AKAudioUnit creation, init, and deinit

JamesOrdner:
Nearly all (if not all?) of AudioKit's Soundpipe AUs leaked memory, due to the init() function of AKDSPBase and subclasses being called twice -- once (improperly) from create[AudioUnit]DSP, and once just before beginning processing.

The first was from create[AudioUnit]DSP, which was ultimately called from AKAudioUnitBase  initWithComponentDescription. This function's purpose as Apple intended is not to actually initialize memory and DSP properties of AUs, but to initialize static properties of AUs. The sample rate and other important properties aren't even known at this time. Sample-rate-dependent properties and other memory should be allocated just before processing begins, from allocateRenderResourcesAndReturnError.

Additionally, many DSP classes did not call their superclass deinit() function from within deinit(), causing soundpipe objects created in AKSoundpipeDSPBase to be leaked.

This PR is as much a proposal as a fix, due to one API-breaking change. The following was changed:

1. AKDSPBase::init() is only called once, from allocateRenderResources. It is no longer called at DSP object creation time.

2. All DSP classes now properly call their superclass' deinit() function, and all memory allocated in init() is deallocated in deinit() (not in the object's destructor).

3. initDSPWithSampleRate(sampleRate, channelCount) was renamed to createDSP(void). This is the API-breaking change. It will only affect those who have independent forks of AudioKit with custom AU classes. As part of this PR, I changed all files already part of AudioKit to conform to the new name. I chose to make this change for a few reasons.
    - The name init is misleading, as its purpose is purely to create the DSP object and return a pointer to it. createDSP much more clearly communicates the purpose of the function.
    - The sample rate may not even be known when this function is called. There is no reason to pass a "best guess" sample rate and channel count, as was previously happening, as the exact values will be known and passed to the AU via allocateRenderResources.
    - This API break will cause build failures for those with custom forks of AudioKit with custom AUs, but the fix is a simple function rename. It will also hopefully alert those with custom AUs of a potential memory leak, assuming they have implemented their AUs following how AudioKit has thus far.
    - As this is in the v5-develop branch, it won't break existing v4 projects

### DIFF
--- a/AudioKit/Common/Internals/AKSoundpipeDSPBase.hpp
+++ b/AudioKit/Common/Internals/AKSoundpipeDSPBase.hpp
@@ -29,15 +29,15 @@ protected:
     sp_data *sp = nullptr;
 public:
 
-    void init(int channelCount, double sampleRate) override {
+    virtual void init(int channelCount, double sampleRate) override {
         AKDSPBase::init(channelCount, sampleRate);
         sp_create(&sp);
         sp->sr = sampleRate;
         sp->nchan = channelCount;
     }
 
-    ~AKSoundpipeDSPBase() {
-        // releasing the memory in the destructor only
+    virtual void deinit() override {
+        AKDSPBase::deinit();
         sp_destroy(&sp);
     }
 
--- a/AudioKit/Common/Internals/CoreAudio/AK4/AKAudioUnitBase.h
+++ b/AudioKit/Common/Internals/CoreAudio/AK4/AKAudioUnitBase.h
@@ -27,7 +27,7 @@
  is. I'm not sure the standard way to deal with this.
  */
 
-- (AKDSPRef _Nonnull)initDSPWithSampleRate:(double)sampleRate channelCount:(AVAudioChannelCount)count;
+- (AKDSPRef _Nonnull)createDSP;
 
 /**
  Sets the parameter tree. The important piece here is that setting the parameter tree
--- a/AudioKit/Common/Internals/CoreAudio/AK4/AKAudioUnitBase.h
+++ b/AudioKit/Common/Internals/CoreAudio/AK4/AKAudioUnitBase.h
@@ -55,7 +55,6 @@
 // Convolution and Phase-Locked Vocoder
 - (void)setupAudioFileTable:(float *_Nonnull)data size:(UInt32)size;
 - (void)setPartitionLength:(int)partitionLength;
-- (void)initConvolutionEngine;
 
 @property (readonly) BOOL isPlaying;
 @property (readonly) BOOL isSetUp;
--- a/AudioKit/Common/Internals/CoreAudio/AK4/AKAudioUnitBase.mm
+++ b/AudioKit/Common/Internals/CoreAudio/AK4/AKAudioUnitBase.mm
@@ -77,7 +77,7 @@
  This should be overridden. All the base class does is make sure that the pointer to the
  DSP is invalid.
  */
-- (AKDSPRef)initDSPWithSampleRate:(double)sampleRate channelCount:(AVAudioChannelCount)count {
+- (AKDSPRef)createDSP {
     return (_dsp = NULL);
 }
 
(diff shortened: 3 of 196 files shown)


aure:
Oh my!

aure:
Incredible! Thank you.

aure:
Looks like unit tests are failing. Do you have time to look at that or shall I?

JamesOrdner:
I can check that out tomorrow if you'd like. My guess is that for some AUs, not everything is initialized in init() as is should be, and this change may have brought that to light? I only tested a subset of AUs before submitting the PR. Is there an easy way to see which nodes/AUs are failing the unit tests?

JamesOrdner:
One last thought before logging off for the day. It's possible that some DSP classes have been programmed in a way in which some properties of the class that are meant to be accessed or modified outside of active rendering (before init() is called or after deinit() is called) are initialized inside of init(). This worked before, because init() was called on DSP object creation, but now that we're only calling it once, where it's supposed to be called, this could cause issues. For example, if I create a node and set a parameter before attaching it to an active audio graph, and the parameter's underlying data isn't setup until init(), we get undefined behavior. (Side note: if this is the case, even if it didn't crash before, the parameters would have been reset when init() is called the second time anyways, wiping out any changes for parameters like this).

If this is the case, any DSP class members intended to be persistent across graph rebuilds/starting or stopping AudioKit should be initialized in the class' constructor and deallocated in the destructor, not in init() and deinit().

aure:
In the Tests folder you can find the test suites for the various platforms. Most of the tests rely on generating a tone with soundpipe, so the error happens on creating the table:

image

JamesOrdner:
Got it, so the issue here is that the soundpipe object isn't initialized until init(), and because setupWaveform() is called before init() and relies on *sp, we're accessing a nullptr.

This is a tricky issue. My first thought was that the simple solution would be to move the soundpipe object creation/deletion to the ctor/dtor of AKSoundpipeDSPBase rather than init/deinit. This would ensure that any AKSoundpipeDSP subclasses which rely on the soundpipe object outside of active processing will have access to it. However, because we don't know what the true session sample rate will be until init() is called, and soundpipe functions which rely on the sp object will most likely query it for this (as sp_ftbl_init does), creating the sp object before init() doesn't make much sense.

A working solution would be to move sp_ftbl_create from setupWavetable() to init(), where we are certain of having a valid soundpipe object with a known sr/channel count. This would require adding a vector<float> in AKOscillatorDSP to store the wavetable locally, as setupWavetable and setWaveformValue are called outside of active processing. This would then be copied to the soundpipe-side frametable buffer in init().

Btw this issue also brings to light another memory leak, because there's no deinit companion to setupWaveform(), where sp_ftbl_create is currently called, and the waveform object is leaked. This one is actually a sizeable leak, because we're malloc'ing enough memory to hold the full waveform buffer. If we moved sp_ftbl_create to init() as proposed, it would be cleanly deallocated in deinit(), fixing the leak.

JamesOrdner:
The same thing is happening with AKConvolutionDSP, though instead of setupWavetable, it's initConvolutionEngine instead. Same error trying to access the sp object before init() is called, same memory leak due to no deinitConvolutionEngine function. However, in this case, initConvolutionEngine takes no parameters, and simply creates the convolution soundpipe objects. I can't see any reason to not just simply delete this function and move its contents to init(). I've searched the entire project for instances where initConvolutionEngine is called and can find no reason not to do this. Unless there is something I'm missing (please let me know!), I'll go ahead and do it, along with the fixes above.

aure:
Tests pass, that's good enough for me. Thank you so much!

aure:
Would you like to be on the AudioKit slack group? Do you have more plans for ways to improve AudioKit? There's real opportunities to get involved and make a difference if you're interested.

JamesOrdner:
I don't know how active I would be able to be due to my full-time job, but I do intend to continue making fixes and improvements to AudioKit as I have time. No grand plans, just tightening up things here and there as I find them. If that kind of non-commitment is okay, then I would love to join!

aure:
Sure, just send me an email address to aure at AudioKit dot io.

One thing you might want to check is how these changes conflict with Ryan's previous pull request on v4 that I had to merge in.  Basically it has to do with getting the right sample rate from AKSettigs instead of AKManager. Here' the PR https://github.com/AudioKit/AudioKit/pull/2013/files

PR #2026, opened 2020-04-12
Added AKNode.detach() call to AKNode deinit

JamesOrdner:
AudioUnits and their underlying C++ DSP objects were not being deallocated upon destruction of owning AKNodes, due to a persistent AVAudioUnit reference from AVAudioEngine. Calling detach() in AKNode's deinit function solves this issue and allows the AudioUnit and the underlying C++ DSP objects to be deallocated.


aure:
Thank you!

PR #2045, opened 2020-05-08
Correct documentation for allowHapticsAndSystemSoundsDuringRecording

jamiehunt:
Existing documentation implied that setting a value of true would suppress haptics, this is actually the inverse.

Corrected typo in 'microphone'


aure:
thanks!

PR #2055, opened 2020-05-14
Converted some classes from AKDSPKernel to AKDSPBase

JamesOrdner:
Converted DynaRageCompressor, RhinoGuitarProcessor, and AKSequencer, and converted their AU classes from objc subclassing AKAudioUnit to swift subclassing AKAudioUnitBase. Still a few classes to be converted before we can delete the old objc AKDSPKernel and AKAudioUnit classes, but it's a good start.


aure:
Thank you. I will make an attempt at making this pass tests again.

PR #2056, opened 2020-05-15
Support for multiple AU input busses

JamesOrdner:
AUs now support an arbitrary number of input busses out of the box. `inBufferListPtr` and `outBufferListPtr` in AKDSPBase have been changed to `inputBufferLists` and `outputBufferLists` respectively, each a vector containing the buffer lists for each bus. In DSP code, this means that `inBufferListPtr->` is now `inputBufferLists[0]->`, and likewise for output busses.

To set the number of input busses, simply resize `inputBufferLists` in the DSP class' constructor to the desired size. The input busses must also be connected to other nodes, which will normally be done in the  AKNode's constructor. AKBalancer accepts two input busses and is a concrete example of how this is done.


aure:
Sadly the Github CI is still not able to keep up the continuous integration testing so I'll have to test on my own.

PR #2067, opened 2020-05-24
Sequencer - Remove specific note at beat and all of a specific note

colinhallett:
All is working fine in 4.9.4 but I haven't tested in anything later.


aure:
I think the function to remove all should be named "removeAllInstancersOf(noteNumber:) - a little longer but much more evocative of what its going to do.

colinhallett:
Yeah good point, I've renamed those functions.

PR #2070, opened 2020-05-27
Generalized parameter automation and other parameter cleanup with AKNodeParameter

JamesOrdner:
Consolidates AKNode parameters into AKNodeParameter, which requires a simple one line declaration passing the identifier key of the AUParameter. Once a valid AU is acquired in the node's init(), the AKNodeParameter should be associated with the AUParameter. AKNodeParameters automatically clamp values to the allowed range, offloading this responsibility from AKNodes or DSP code, and allows each parameter's ramp duration, taper, and skew to be set individually.

All DSP parameters now use the ParameterRamper class, rather than AKParameterRamp.

The AKAutomatable protocol has been simplified to only require a single AKParameterAutomation object in any implementing class, and most AKNodes now implement this protocol. All automation-related functions, including start and stop, are now fully inside AKParameterAutomation and not required to be implemented by AKNodes. Additionally, automation points are now identified by their identifier key, rather than their address, to allow decoupling of client-side saved parameter automation and AK-side parameter addresses moving forward.

Each automation point may include an individual taper, skew, and offset (skew/taper/offset are no longer node-wide properties). This is achieved within AKNodeParameter at parameter scheduling time, by or'ing bitwise flags with the most significant bits of the parameter address for each taper/skew/offset property, respectively. This is interpreted by AKDSPBase::startRamp(), which schedules events for each registered parameter automatically. Subclasses no longer implement startRamp().

### DIFF
--- a/AudioKit/Common/Internals/AKComponent.swift
+++ b/AudioKit/Common/Internals/AKComponent.swift
@@ -20,7 +20,7 @@ extension AUEffect {
 
 /// Helpful in reducing repetitive code in AudioKit
 public protocol AKComponent: AUComponent {
-    associatedtype AKAudioUnitType: AnyObject // eventually AKAudioUnitBase
+    associatedtype AKAudioUnitType: AUAudioUnit // eventually AKAudioUnitBase
     var internalAU: AKAudioUnitType? { get }
     var rampDuration: Double { get set }
 }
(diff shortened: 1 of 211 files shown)


ryanfrancesconi:
Thanks James! I will try this tomorrow.

aure:
This looks amazing and the failing tests are something I can fix so I think I'll merge it and make updates. WOW!

PR #2072, opened 2020-05-29
Consolidate the various platform Xcode projects into a single one

megastep:
This addresses a long-standing pain point in the way AudioKit platforms have been managed, by unifying all three platform-specific projects within a single `AudioKit.xcodeproj`.

There are six targets in that project (2 per platform, AudioKit and AudioKitUI). The end usage is not very different from before (mostly have to check you're using the target for the right platform), and the changes to the scripts were minimal.

As a result a number of files have been relocated and some duplicates were removed. This will make it easier to add new cross-platform features from now on, as including new files in an OS target is now just a matter of checking the box for the target in Xcode.


aure:
Killer

PR #2084, opened 2020-06-08
Add argument to change buffer size of AKFFTTap

### LINKED ISSUE #2078: Custom buffer size for AKFFTTap
I am trying to use the AKFFTTap to analyze what frequency is being played, but the default resolution of 1024 is not big enough. It looks like the fft size is hard coded to this value.

Could we possibly add the bufferSize as an argument to the constructor?
I don't know how the underlying code works, but it appears to be quite trivial to change.

The [AKSettings.BufferLength](https://github.com/audiokit/AudioKit/blob/master/AudioKit/Common/Internals/AKSettings.swift#L16) enum could be used to enforce a value in the power of two.

https://github.com/AudioKit/AudioKit/blob/2fab72d4ed30a1a19ff9a964ef03c0945c56de66/AudioKit/Common/Taps/FFT%20Tap/AKFFTTap.swift#L9-L16

Related: @steve228uk did post a modified version using a buffer size of 16 here https://github.com/AudioKit/AudioKit/issues/748#issuecomment-279989826

viktorstrate:
This resolves #2078 

The argument `fftSize` has been added to the init function of AKFFTTap, to define a custom buffer size.

A default value of .veryLong (1024 samples) has been set, this way the old api still works.

### DIFF
--- a/AudioKit/Common/Taps/FFT Tap/AKFFTTap.swift
+++ b/AudioKit/Common/Taps/FFT Tap/AKFFTTap.swift
@@ -9,17 +9,28 @@
 /// FFT Calculation for any node
 open class AKFFTTap: NSObject, EZAudioFFTDelegate {
 
-    internal let bufferSize: UInt32 = 1_024
+    public let fftSize: AKSettings.BufferLength
+    
+    internal var bufferSize: UInt32 {
+        get {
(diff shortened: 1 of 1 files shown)


aure:
Thanks!

PR #2090, opened 2020-06-11
Efficient parameter automation playback and recording

JamesOrdner:
Converted former AKParameterAutomation class from .h/.m pair to Swift/C++ pair. More efficient playback of automation. Parameter automation timings are now expressed in seconds, rather than samples. Automation points are scheduled similar to using a sequencer, where the beginning of the sequence is time 0, allowing a much more intuitive interface. Beginning automation playback in the middle of the sequence, i.e. at time > 0, will automatically handle setting the value to the correct value based on past or current events, including handling starting playback in the middle of a ramp. Added the ability to record parameter automation. AKParameterAutomation is compatible with any AVAudioNode which implements parameters as AUParameters, allowing the use of non-AK AUs.

The largest change is parameter automation recording. This may be enabled for each parameter on a node individually. I'll copy and paste the description of AKParameterAutomation.setRecordingEnabled() below for a description of how it works:

"When armed for recording, existing automation will play normally until the AUParameter receives a setValue() call with the .touch event type. Assuming the use of AKNodeParameter, this is done by calling the beginTouch() function. When a .touch event is received, existing automation will be muted, and any changes to the parameter's value will be written as points to the automation track at the times during playback in which they are received, overwriting existing points. Recording will continue until a setValue() call with the .release event type is received, or endTouch() in AKNodeParameter. When a .release event is received, normal playback will resume."


ryanfrancesconi:
let's merge this in and i'll test out in context. I don't have a UI for the automation recording, but this is a rad feature.

aure:
Woohoo

PR #2096, opened 2020-06-12
Allow specifying buffer size when creating AKAmplitudeTap

tcwalther:
This PR makes the `bufferSize` in `AKAmplitudeTap` configurable at creation time. This is analogous to `AKFFTTap`; however, unlike the solution there which takes an `fftSize: AKSettings.BufferLength` argument, my proposal allows specifying arbitrary `bufferSize`s.

This can be useful when you want to track amplitude in a specific time interval, for example, every 10ms.


aure:
Thanks!

PR #2097, opened 2020-06-13
STKAudioKit

aure:
This is a proof of concept for extracting portions of AudioKit into subframeworks which have to be imported. I hope doing this will create better code separation for these codebases and will allow people who don't need a certain portion of functionality to avoid compiling for it. Perhaps it is also a first step towards putting things in completely separate frameworks.

Changes for the build process and such, I would hope that I can get help from @megastep on. 🙏


megastep:
You had some spurious plists (no doubt created by Xcode when you duplicated targets) which were causing the tests to fail.
I think just adding `#import "AKLinearParameterRamp.hpp"` to the AK umbrella header should fix these warnings you're experiencing.

aure:
@ryanfrancesconi Believe me, I am tempted to remove it in AK5, but this is a first step towards that and its a step that teaches how to build an audio framework on top of AudioKit instead of adding to AudioKit itself. Ideally we could interface to a lot of different code bases, but keep the different kinds of codes out of sight.

PR #2111, opened 2020-06-23
Added Meta Event 09

emurray2:
Added support for Meta Event 09 which determines the Device/Port name each MIDI track is connected to


aure:
Keep the pull requests coming! Woohoo. :)

emurray2:
Thank you! I will!

PR #2117, opened 2020-06-26
EndpointInfo is now Codable.

trevor-sonic:
So we can persist last used ```EndpointInfo``` easy.

_(this update is not effecting any previous use of this object, safe.)_


trevor-sonic:
Sorry for **Colon Violation** , promise next time I'll clean up all take care about.

PR #2122, opened 2020-06-27
Support changing the default audio device on macOS

wtholliday:
Adds support for changing the default audio device on macOS while an app is running.

Also modernizes notification handlers.


aure:
👍 will merge after continuous integration passes

PR #2125, opened 2020-06-28
make maximumFrameCount of renderToFile available for user 

seaniezhao:
I'm developing a app which can export user's performance and background music as one wav file. I need to adjust maximumFrameCount to control the progress callback interval. I think this would be a feature to control the progress callback interval.


aure:
Thanks

PR #2132, opened 2020-07-03
Fix small error in frameworks documentation

### LINKED ISSUE #2114: Small error in documentation about including from source
Could it be that this is a small mistake installation instructions to include AudioKit from source?:

"In the **Build Phases** tab, add AudioKit.framework in **Target Dependencies** for your target. Also add AudioKitUI.framework as needed."

From: https://github.com/audiokit/AudioKit/blob/master/Frameworks/README.md

I don't see **Target Dependencies** under the **Build Phases** tab. I think it should be **Frameworks, Libraries and Embedded Content** under the **General** tab.

gertjanjansen:
Fixes #2114


aure:
Thanks!

PR #2135, opened 2020-07-07
Added Tempo Automation

emurray2:
I added tempo automation support for the `AKMIDITrackView.swift`. Every MIDI file seems to play at the right tempo now. There are a certain 10% of MIDI files which still crash when loaded, most likely due to more unfamiliar meta events. I will continue fixing things and make an example for Mac Catalyst.

On that note, I am able to run this with AudioKit 5.0 which is awesome. However, I updated my iPhone to iOS 14 before and noticed a bug. When setting the MIDI output port for the `AKAppleSequencer` tracks, not all of the MIDI voices are parsed/played to that MIDI port. The sequencer usually picks one voice at random and only plays that. However, if I do not connect any MIDI output ports to the AKAppleSequencer, I notice that all of the default sine wave voices play fine. I feel that this might have something to do with Apple's `MusicSequence` class, and possibly it does not work fully yet on the iOS 14 beta. I downgraded back to iOS 13.5.1, and everything works fine when I build with Xcode 12 on macOS Big Sur. 

I am hoping this won't interfere with my process of creating this for Mac Catalyst. I am not sure if this is something Apple should fix, but I feel as if it is. I will keep investigating and debugging to see if the `MusicSequence` is really the center of this bug.

### DIFF
--- a/AudioKit/Common/MIDI/MIDIFile/AKMIDIFileTrackMap.swift
+++ b/AudioKit/Common/MIDI/MIDIFile/AKMIDIFileTrackMap.swift
@@ -21,19 +21,16 @@ public class AKMIDIFileTrackNoteMap {
     public let midiTrack: AKMIDIFileTrack!
     public let midiFile: AKMIDIFile!
     public let trackNum: Int!
-    //AudioKit midi object reference for later implementation with the AKMIDITempoListener
-    let midi = AKManager.midi
-    public let tempoListener = AKMIDITempoListener(smoothing: 0.98, bpmHistoryLimit: 1)
     public var loNote: Int {
         if noteList.count >= 2 {
-            return (noteList.min(by: { $0.noteNum < $1.noteNum })?.noteNum)!
+            return (noteList.min(by: { $0.noteNum < $1.noteNum })?.noteNum) ?? 0
         } else {
             return 0
         }
     }
     public var hiNote: Int {
         if noteList.count >= 2 {
-            return (noteList.max(by: { $0.noteNum < $1.noteNum })?.noteNum)!
+            return (noteList.max(by: { $0.noteNum < $1.noteNum })?.noteNum) ?? 0
         } else {
             return 0
         }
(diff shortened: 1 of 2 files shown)


aure:
Definitely closely monitor the iOS14 and it might be worth it to alert them about the problem through one their issue tracking systems. They'll definitely balk if you can't make any example of the bug using pure Apple frameworks only (ie. no AudioKit).

emurray2:
Sounds good! I’ll work on doing that!

emurray2:
I have been digging around in the AudioKit and Apple code to continue finding this bug, but I am getting closer! I know the bug is coming from the `AKAppleSequencer`. When I connect the sequencer to a `AKMIDINode` or `AKMIDISampler` by running `setGlobalMIDIOutput(_ midiEndpoint: MIDIEndpointRef)`, that is when the bug happens. Not all of the MIDI events are parsed correctly. However, if I do not connect the sequencer to anything, it plays its default `AUGraph` with the iOS monotimbral synthesizer Audio Unit, and that plays fine. So, I am thinking this problem is related to the new `CoreMIDI` on iOS 14.0 . I have been poking around in `MIDIServices.h` and I found something interesting.

The method `MIDIDestinationCreateWithBlock
( MIDIClientRef client, 
CFStringRef name, 
MIDIEndpointRef * outDest, 
MIDIReadBlock readBlock )` is marked with `API_DEPRECATED_WITH_REPLACEMENT
("MIDIDestinationCreateWithProtocol", 
macos(10.11, API_TO_BE_DEPRECATED),
ios(9.0, API_TO_BE_DEPRECATED), 
tvos(12.0, API_TO_BE_DEPRECATED));` .

I'm not sure if this has anything to do with the problem, but I noticed that AudioKit still uses the method above which is going to be deprecated. In general, I am interested in learning about if something really needs to be updated and not be used when it is deprecated. I have heard different things from other people. For example, I heard Apple decided to deprecate OpenGL, but people still use it.

I am just curious, when is it acceptable to use something which is deprecated? Is it a good idea to use the new methods to prevent bugs?

I am eager to learn more! Please let me know if you have any advice, and I will continue looking into this bug.

PR #2141, opened 2020-07-09
ExtendingAudioKit cleanup

wtholliday:
AudioKit exposes AKDSPBase, so it's no longer necessary to use the core headers directly.

Further clean up and simplify the code.


aure:
Thank you!

PR #2149, opened 2020-07-11
Objc cleanup

wtholliday:
Reduce ObjC interface surface area. This cleans up the code and may improve build times.


wtholliday:
Test failure looks like code sign BS.

PR #2150, opened 2020-07-11
Add property wrapper for node parameters.

wtholliday:
Wraps AKNodeParameter so we can easily assign values to it. Shows how to update `AKOscillator` to use `@Parameter`.

Instead of`osc.frequency.value = 440`, we have `osc.frequency = 440`

Use the $ operator to access the underlying AKNodeParameter. For example:
`osc.$frequency.maxValue`

When writing an AKNode, use:
```
@Parameter("myParameterName") var myParameterName: AUValue
```
This syntax gives us additional flexibility for how parameters are implemented.


wtholliday:
Need a top-level workspace that builds everything.

PR #2154, opened 2020-07-13
Fixed MIDI Packets iOS 14 beta

emurray2:
I changed some of the code in `AKMIDISampler.swift` a little bit to fix the MIDI bug in iOS 14 beta I mentioned.

Now all of the MIDI note events can be sent to a sampler. The only thing I'm concerned about is the fact that I changed it so only events with a length of three (note events) can go through.

Please let me know if you have any advice or feedback. I can always change things and resubmit.

Other than that, I've tested this on both iOS 13.5.1 and iOS 14.0, and everything works great.

Thank you.

### DIFF
--- a/AudioKit/Common/MIDI/AKMIDISampler.swift
+++ b/AudioKit/Common/MIDI/AKMIDISampler.swift
@@ -37,11 +37,14 @@ open class AKMIDISampler: AKAppleSampler {
                          name: String = "MIDI Sampler") {
         CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
             for e in packetList.pointee {
-                let event = AKMIDIEvent(packet: e)
-                do {
-                    try self.handle(event: event)
-                } catch let exception {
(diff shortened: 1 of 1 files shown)


jdanthinne:
Had to do the same fix for `AKMIDIInstrument` as well, in `enableMIDI` method.

emurray2:
@jdanthinne Ah yes I forgot about that. That makes sense. I think all of the `enableMIDI` methods might have misread the MIDI packets since Apple changed the MIDI API a little bit on iOS 14.

jdanthinne:
Will you update the code or do you need me to make a PR?

emurray2:
It would probably be best if you make a PR. I'm busy with school at the moment and my computer has been having some audio bugs. Now that we have `MIDISampler` and `MIDIInstrument` taken care of, you can probably check the other MIDI classes to see if their `enableMIDI` methods need to be updated.

jdanthinne:
Already fixed in [PR 2281](https://github.com/AudioKit/AudioKit/pull/2281).

PR #2160, opened 2020-07-14
Add support for using AKNodeParameter instead of strings in AKParameterAutomation

wtholliday:
We can do:
```
parameterAutomation?.add(point: point, to: $leftGain)
```

instead of

```
parameterAutomation?.add(point: point, to: "leftGain")
```


aure:
Thank you! That is better.

PR #2162, opened 2020-07-15
Avoid using EZAudio in AKManager

wtholliday:
Adds audio device selection code for macOS.

Modernizes AKDevice.


aure:
Major! Thank you!

PR #2170, opened 2020-07-17
fix memory bugs caught by address sanitizer

wtholliday:
MIDIMetaEvent is a variable length C structure which has to be handled carefully from Swift.

Note that I don't have a way of testing AKMusicTrack other than running the test suite.

### DIFF
--- a/AudioKit/Common/Sequencing/Apple Sequencer/MIDIMetaEvent+allocate.swift
+++ b/AudioKit/Common/Sequencing/Apple Sequencer/MIDIMetaEvent+allocate.swift
@@ -1,0 +1,32 @@
+// Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
+
+import Foundation
+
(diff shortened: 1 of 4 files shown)


AudioKit/Common/Sequencing/Apple Sequencer/MIDIMetaEvent+allocate.swift
```diff
@@ -0,0 +1,30 @@
+// Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
+
+import Foundation
+
+extension MIDIMetaEvent {
+    /// `MIDIMetaEvent` is a variable length C structure. YOU MUST create one using this function
```
eljeff, line 6: I read your comment 'YOU MUST' literally... anyway we can lock it down so there is no possible way to do this otherwise?
I create a lot of metaEvents in the file reader - will those just all crash now?

PR #2180, opened 2020-07-19
Add regression testing of internal DSP state

wtholliday:
This allows us to ensure an internal DSP state variable is what we expect for the duration of a test, giving the developer more information when a test fails.

Note that we still need to ensure this doesn't cause any performance issues by using a macro to only call `DebugDSP` during testing.


aure:
hard core

PR #2181, opened 2020-07-20
Use single multi-platform framework target

wtholliday:
Instead of three framework targets, use a single target for all platforms.

This just gets things started, showing it can work.

Next steps:
- is to update the build scripts to use the Multiplatform target.
- rename the target and delete the others.


aure:
Merged now rather than waiting, hope that's cool.

megastep:
Well I see you haven't deleted the other targets yet so at least it keeps building. I'm still on the fence on whether this is necessarily going to solve anything here instead of creating more potential problems with Xcode...

PR #2185, opened 2020-07-21
Switch new automation code to linear automation

wtholliday:
Instead of supporting curved automation in `AKNodeParameter`, provide `AKEvaluateAutomation` which converts curved automation to linear automation. This keeps the DSP side simple and offers compatibility with Apple AUs.


wtholliday:
> There might be some pain on getting ADD back to its functionality now, but this the way forward, so maybe better to get it on the right track now than to wait. I still think my nomenclature suggestions should be considered before merge.

This should not break ADD functionality. I haven't touched the `AKParameterAutomation` class. I've only removed curved automation from the new automation code.

ryanfrancesconi:
so how do i do a tapered curve with automation?

wtholliday:
@ryanfrancesconi 

```
let point = AKParameterAutomationPoint( ... )
let events = AKEvaluateAutomation(initialValue: 0, points: [point], resolution: 0.1)
node.$param.automate(events: events)
```

ryanfrancesconi:
what does resolution do?

wtholliday:
```
///   - resolution: Duration of each linear segment in seconds
```

PR #2193, opened 2020-07-25
Fix delayed start with AKPlayer.

wtholliday:
Also adds regression test.


ryanfrancesconi:
great thanks. i will test this sunday morning.

PR #2204, opened 2020-07-30
Balancer problem, making a PR to see if the issue happens on CI and other people's machines

aure:
This simple change, similar to ones that worked this morning, breaks AudioKit.h. Why?


megastep:
Well I see that it fails to compile because of unknown references. Maybe it has something to do with Taylor's recent changes to make internal APIs less accessible?

aure:
> Well I see that it fails to compile because of unknown references. Maybe it has something to do with Taylor's recent changes to make internal APIs less accessible?

No I don't think. I did this to 50 or 60 nodes in a previous commit without issue.

aure:
I see the issue now. The balancer hpp file was the last place that AKSoundpipeDSPBase.hpp was imported from a public header file.

aure:
I've made the AKSoundpipeDSP.hpp file into a .hpp/.mm pair but still having some issues.

PR #2215, opened 2020-08-03
Fix threading issues in AKSequencerEngine

wtholliday:
- Fix issues discovered by TSAN
- Improve threading correctness.


megastep:
Does that mean we can re-enable TSAN runs in CI?

wtholliday:
@megastep I'm afraid not. This was for some new code I added.

Issue #2196, opened 2020-07-27
5.0.b1: How to export a portion of a file with AudioKit?


pcki11:
I understand that 5.0.b1 is beta software, I am just trying to run my app with Xcode 12 and 4.* does not compile there.
With 4.* I had this code to export an AKAudioFile into a desired format and with defined start and end positions:
```
let sampleRate = audioFile.fileFormat.sampleRate
let fromSample = Int64(audioFile.duration * start * sampleRate)
let toSample = Int64(audioFile.duration * end * sampleRate)
audioFile.exportAsynchronously(name: "audio", baseDir: .temp, exportFormat: exportFormat, fromSample: fromSample, toSample: toSample) { file, error in
    guard let file = file else {
        return
    }
    success(file)
}
```
Currently it seems that I have to use AKConverter for this task, but it does not have an option to specify fromSample and toSample when I convert the file. So how can I convert my code so that it works with v.5.0.b1?

ryanfrancesconi:
Yes, i can add that functionality in.

ryanfrancesconi:
I've added two new extract methods to AVAudioFile in v5-master: https://github.com/ryanfrancesconi/AudioKit/tree/v5-master

```
/// Will return a 32bit CAF file with the format of this buffer
    @discardableResult public func extract(to outputURL: URL,
                                           from startTime: TimeInterval,
                                           to endTime: TimeInterval,
                                           fadeInTime: TimeInterval = 0,
                                           fadeOutTime: TimeInterval = 0) -> AVAudioFile?

/// - Returns: An extracted section of this file of the passed in conversion options
    public func extract(to url: URL,
                        from startTime: TimeInterval,
                        to endTime: TimeInterval,
                        fadeInTime: TimeInterval = 0,
                        fadeOutTime: TimeInterval = 0,
                        options: AKConverter.Options? = nil,
                        completionHandler: AKConverter.AKConverterCallback? = nil)

```

ryanfrancesconi:
the fades are useful in the case where you want to ensure zero crossings to your edits, like 0.01 values.

ryanfrancesconi:
There is also an extract in the AVAudioPCMBuffer now:

```
/// - Parameter from: The time of the in point of the extraction
    /// - Parameter to: The time of the out point
    /// - Returns: A new edited AVAudioPCMBuffer
    public func extract(from startTime: TimeInterval,
                        to endTime: TimeInterval) -> AVAudioPCMBuffer?
```

PR #2216, opened 2020-08-03
Extraction added to buffers, files, and PlayerDemo

ryanfrancesconi:
Example of 2 use cases in PlayerDemo:
* crop - like a sound editor
* extract selection to file


ryanfrancesconi:
Solves: https://github.com/AudioKit/AudioKit/issues/2196

PR #2219, opened 2020-08-04
Add akGetParameterAddress

wtholliday:
- Reduces namespace pollution
- Avoids having to edit AudioKit.h when adding a node
- Demonstrates how to upgrade AKOscillatorDSP.

### DIFF
--- a/AudioKit/Common/Nodes/Generators/Oscillators/Oscillator/AKOscillatorDSP.mm
+++ b/AudioKit/Common/Nodes/Generators/Oscillators/Oscillator/AKOscillatorDSP.mm
@@ -85,7 +85,7 @@ public:
 };
 
 AK_REGISTER_DSP(AKOscillatorDSP)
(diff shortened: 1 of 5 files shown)


AudioKit/Common/Nodes/Generators/Oscillators/Oscillator/AKOscillatorDSP.mm
```diff
@@ -6,6 +6,13 @@
 
 #include "DebugDSP.h"
 
+enum AKOscilatorParameter : AUParameterAddress {
```
aure, line 9: I don't understand how this worked with the typo misspelling of oscillator.  Almost as if nothing refereed to this?

wtholliday: It's true, nothing refers to it. The enum values are stringified by the AK_REGISTER_PARAMETER macro.

PR #2229, opened 2020-08-20
V5 master, isPlaying not being set to false on AKPlayer completion if there is no user supplied completion handler

ryanfrancesconi:
fix for @warpling


ryanfrancesconi:
@warpling merged

PR #2243, opened 2020-09-01
CI rules and script updates for CocoaPods / staging

megastep:
This builds on the existing SPM support in v5 to bring back support for CocoaPods.

It doesn't seem like it would still be possible to build AudioKit as a source CocoaPod, so for people who will keep using that instead of SPM we're still doing it the old fashioned way, via a set of static XCFrameworks which the pod spec references. The Xcode project is generated via the `swift` command and much of the same build processes are handled via `xcodebuild`.

Like previously, merging to the staging branch will trigger the build and upload to the S3 bucket where staging builds are stored. Tagging the repo will also trigger an upload to the corresponding GitHub release. Builds are not produced in any other case, as it takes about 30 minutes to complete.

To be added later: major cleanups of the contents of the Frameworks folder (especially the docs), and try to build frameworks with Apple Silicon slices via Xcode 12. I believe the Github Actions image might be caught up enough now with the Xcode 12 beta to accommodate this.


aure:
Seems reasonable, doesn't affect our nice light happy making SPM stuff. I suppose people will still want Cocoapods.

trevor-sonic:
> It doesn't seem like it would still be possible to build AudioKit as a source CocoaPod, so for people who will keep using that instead of SPM we're still doing it the old fashioned way

@megastep  SPM is the future, I would stick to SPM only for v5. Since I started using SPM, I'm finally able to separate my project in many small reusable framework/project. I'm looking forward to use AK with SPM. I don't see any single interest to use CocoaPod apart being lazy to learn SPM :-D 
(Of course this is my personal opinion, sorry if I so open/direct :-) )

trevor-sonic:
> Seems reasonable, doesn't affect our nice light happy making SPM stuff. 👍🏻

megastep:
@trevor-sonic Hey, that's fine. I think the point of this is we can support both with relatively minor changes to what we did before so I think we're better off offering it at least as a transition. If you prefer SPM then by all means use that, it will definitely be the primary way to use AudioKit v5 going forward.

aure:
btw, as far as I'm concerned its your call as to when to merge this, @megastep

megastep:
@aure still want to do a bit of clean up and tests on my branch. I think I have working Xcode 12 support with Apple Silicon

PR #2268, opened 2020-09-17
Prefix removal

aure:
Other than AKMIDI most of the AK prefixes have been dropped.


megastep:
Maybe I'm missing something but what's the rationale behind this? Wouldn't it make it more likely that symbols in AudioKit are going to conflict with similar ones in other apps/frameworks? Even Apple still uses prefixes except for core language features.

PR #2270, opened 2020-09-18
EndpointInfo (obj) for everything. Clean MIDI  in, out, open, close, send, receive.

trevor-sonic:
Hi every time I use MIDI i face a problem that some functions works with name (string) other refId (number) to many functions make me confuse.
I've cleaned up this work flow. I made everything works with ```EndpointInfo``` (it's an object)
(I didn't delete anything of old functions will not break anything, promise :-))

I've added added 2 major extension
```AKMIDI+Receiving+EndpointInfo.swift```
```AKMIDI+Sending+EndpointInfo.swift```
These are working with a ```EndpointInfo``` object which contains all valuable information.
```EndpointInfo``` (maybe we better refactor it ,something like ```EndpointObj``` or  so)
So with this unique object (```EndpointInfo```) we can open/close midi inputs and outputs confidently.
```AKMIDI.inputs()``` returns  array of input EndPointInfo as array ```[EndpointInfo]```
```AKMIDI.outputs()``` returns  array of input EndPointInfo as array```[EndpointInfo]```

```openOutput(outputInfo: EndpointInfo)``` open given endpoint
```closeOutput(outputInfo: EndpointInfo)``` close given endpoint
```openInput(outputInfo: EndpointInfo)``` open given endpoint
```closeInput(outputInfo: EndpointInfo)``` close given endpoint

```
sendNoteOnMessage(noteNumber: MIDINoteNumber, velocity: MIDIVelocity, channel: MIDIChannel = 0, endpointInfo:EndpointInfo) 
``` 
And some more function on its way to implement (when I find time)

I push here (v4) because I tested on v4.11 OK. For v5 we can inspire from this codes.
(I changed 3 files mainly but lot of file look modified in my pull request sorry maybe clumsiness with git)


aure:
Hi, there's a problem here. You have added a lot of duplicate copies of files. Looks like something happened where the Finder automatically created files with " 2" appended to the name.  There's no way I can add this too the repo. I'd suggest you make a clean fork of the repo and re add just the files you've added to the new repo. ie. Don't just fix the files here because the git history is already corrupted on your fork of AudioKit. This does look like a valid addition to AudioKit otherwise, so I am sorry to have to reject it.

trevor-sonic:
Hi, no problem I understand. I thought you may have some magical power to eliminate wrong files. :-) Apparently no.
I'll follow your advise, push later.

Issue #2284, opened 2020-09-25
Not receiving SysEx messages on iOS 14


yuriizubkov:
Hello devs. Thank you for your great framework!
I am using the MIDI part of AudioKit in my app. 

In some cases I am sending sequential SysEx requests (and receiving responses) from my app to MIDI synth with 100ms gap between messages. All was working stable on iOS 11-13 with no issues at all. But the same build of the app stopped working on iOS 14. I tried both AudioKit 5 and 4.11.2 - same issues on iOS 14.

I am getting the first couple of SysEx messages and then I can’t receive anything, SysEx, CC, Note information.
What I found in the logs. Usually I can see in the logs:

```
AKMIDI.swift:startReceivingSysEx(with:):92:Starting to receive SysEx (AKMIDI.swift:startReceivingSysEx(with:):92)
AKMIDI.swift:stopReceivingSysEx():97:Done receiving SysEx (AKMIDI.swift:stopReceivingSysEx():97)
```

But on iOS 14 at some point it stuck at:
```
AKMIDI.swift:startReceivingSysEx(with:):92:Starting to receive SysEx (AKMIDI.swift:startReceivingSysEx(with:):92)
```

And after that I can’t receive any MIDI message from the synth. But I can still send MIDI messages.
What I found in the AudioKit MIDI implementation, it might be some race condition around:
`AKManager.midi.isReceivingSysEx`

Or maybe internal format of MIDI packets has changed so parser can’t recognise the end of SysEx message.
I see many changes in CoreMIDI implementation on iOS 14, so it can be so the internal behaviour changed somehow.

I am not providing my code here, because I think it is not relevant to this issue, since everything is working on iOS 13.
Tell me if I can help to debug this issue better.
Thank you.

yuriizubkov:
After further investigation I think this issue might be related to this:
https://github.com/AudioKit/AudioKit/issues/2145#issue-655054321

Because **MIDIInputPortCreateWithBlock** is deprecated on iOS 14.
List of all deprecated symbols is here: https://developer.apple.com/documentation/coremidi/midi_services/deprecated_symbols

And you are using it for parsing incoming MIDI messages here: https://github.com/AudioKit/AudioKit/blob/v5-develop/Sources/AudioKit/MIDI/MIDI%2BReceiving.swift

My incoming SysEx message is 350 bytes, so this is why probably I am not getting second part of the message and AKManager.midi.isReceivingSysEx never got set to true.

[...]

PR #2292, opened 2020-10-05
Added support for the new function MIDIInputPortCreateWithProtocol on iOS 14+

yuriizubkov:
This PR is related to this issue: https://github.com/AudioKit/AudioKit/issues/2284

Added support for the new function MIDIInputPortCreateWithProtocol on iOS 14+. 
The old parser behavior has been retained for previous iOS versions.

Only parsing of MIDI 1.0 over UMP Messages is working for now.
We need to add a bunch of new stuff from MIDI 2.0 specification soon...

Not quite sure where to put new enums and helper functions (such as byteArray<T>(from value: T), getMSB(from uint8: UInt8), getLSB(from uint8: UInt8))

I tested it with my own hardware synth and I am able to receive CC, Note On/Off, Clock etc. as well as full patches in SysEx format with size of 350 bytes.

### DIFF
--- a/Sources/AudioKit/MIDI/MIDI+Receiving.swift
+++ b/Sources/AudioKit/MIDI/MIDI+Receiving.swift
@@ -138,6 +138,192 @@ extension MIDI {
         let uid = uidForInputAtIndex(inputIndex)
         openInput(uid: uid)
     }
+    
+    /// Message type of the Universal MIDI Packet
+    ///
+    /// https://www.midi.org/midi-articles/details-about-midi-2-0-midi-ci-profiles-and-property-exchange
+    enum UMPMessageType: UInt8 {
+        case Utility32bit = 0x0
+        case SystemRealTimeAndCommon32bit = 0x1
+        case MIDI1ChannelVoice32bit = 0x2
+        case DataAndSysEx64bit = 0x3
+        case MIDI2ChannelVoice64bit = 0x4
+        case Data128bit = 0x5
+        case Reserved32bit_1 = 0x6
+        case Reserved32bit_2 = 0x7
+        case Reserved64bit_3 = 0x8
+        case Reserved64bit_4 = 0x9
+        case Reserved64bit_5 = 0xA
+        case Reserved96bit_6 = 0xB
+        case Reserved96bit_7 = 0xC
+        case Reserved128bit_8 = 0xD
+        case Reserved128bit_9 = 0xE
+        case Reserved128bit_10 = 0xF
+    }
+    
+    /// Status of each UMP in a System Exclusive message
+    ///
+    /// Chapter 4.4 of M2-104-UM Universal MIDI Packet (UMP) Format and MIDI 2.0 Protocol
+    /// http://download.xskernel.org/docs/protocols/M2-104-UM_v1-0_UMP_and_MIDI_2-0_Protocol_Specification.pdf
+    enum UMPSysEx7bitStatus: UInt8 {
+        case CompleteMessage = 0x0
+        case Start = 0x1
+        case Continue = 0x2
+        case End = 0x3
+    }
+    
+    private func byteArray<T>(from value: T) -> [UInt8] where T: FixedWidthInteger {
+        withUnsafeBytes(of: value.bigEndian, Array.init)
+    }
+    
+    private func getMSB(from uint8: UInt8) -> UInt8 {
+        return (uint8 & 0xF0) >> 4
+    }
+    
+    private func getLSB(from uint8: UInt8) -> UInt8 {
+        return uint8 & 0x0F
+    }
+    
+    /// The most significant 4 bits of the first UInt32 word in every UMP shall contain the Message Type field.
+    /// UMP Message Type can tell us how many 32bit UMP packets we have to read next in order to get full MIDI message.
+    /// A Universal MIDI Packet contains a MIDI message which can consists of one to four 32-bit words.
+    ///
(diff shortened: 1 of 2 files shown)


aure:
Very odd that this doesn't build on Xcode 11.7 and only on 12. @megastep any idea? I'm inclined to accept it because I'm on Xcode 12, but I am also inclined to just support iOS13 so I am a bit less conservative in general.

megastep:
Seems to be mostly because of missing symbols: `MIDIInputPortCreateWithProtocol` and `MIDIEventPacket`; I'm guessing those were added in iOS 14 so the appropriate sections of code need to be surrounded by `@available` blocks and such, as appropriate. I don't see why we couldn't make it compile with proper checks.

aure:
@megastep Yeah, that makes. @yuriizubkov what do you think of updating it to an iOS12 level or putting in checks?

yuriizubkov:
@megastep @aure Yes, this is strange, because I have checks in this PR here:
https://github.com/AudioKit/AudioKit/pull/2292/commits/243be454c307083bc86323986b58a93b6a56cc25#diff-d5817501a888604d2d2eaf4b035c60b4R252
and here:
https://github.com/AudioKit/AudioKit/pull/2292/commits/243be454c307083bc86323986b58a93b6a56cc25#diff-d5817501a888604d2d2eaf4b035c60b4R341
to support older iOS versions and for iOS 14+.
Maybe I missed something?

megastep:
Might need to complement this with some `#if` macros maybe? The APIs are just not visible at all on Xcode 11.7, though if you have `@available` set up this way we should be good. It's just going to fail compiling in Xcode 11.7, but frankly maybe we should stop testing for that.

yuriizubkov:
@megastep I see in the CI logs that the second check `if #available(iOS 14, *)` is also ignored by the compiler. So I think it will fail as well as `@available` on the first check.
Hmmm... not sure how to fix that. I use Xcode 12, everything looks good on Xcode 12.

megastep:
Yeah I don't think Swift has `#if` macros to check for a specific version of iOS, that's what `@available` is for, but it doesn't help if you're using "future" APIs.

I'm thinking of just disabling tests on Xcode 11.7 if we're going to start using iOS 14-only features. It would actually be more useful to restore tests across the various platforms rather than testing across older versions of Xcode.

yuriizubkov:
I mean we can check availability of the specific iOS versions with:
```
if #available(iOS 14, *) {
    // use new APIs
} else {
    // use older code
}
```
or mark the whole method as:
```
@available(iOS 13, *)
func useNewAPI() {
    // new stuff
}
```
https://www.hackingwithswift.com/new-syntax-swift-2-availability-checking
https://www.raywenderlich.com/908-availability-attributes-in-swift
I am new to Swift, so I am trying to make it clear for myself :)

But I think I got your point, if I try to compile this code with new APIs on older Xcode, it will fail anyway in this case.

aure:
I have merged this as it compiled on Xcode 12.

aure:
Oops, I was mistaken, it errors on Xcode 12 too, I may have to revert. Submit another PR when you're able to, thanks.

yuriizubkov:
@aure @megastep I don't understand why it fails on the CI. Because I have compiled it from my local branch and I already using it in my app without any of those errors. My Xcode build is 12.0.1 (12A7300). Can you help me please?

yuriizubkov:
@megastep I think this might be something to do with CI itself?
An example of similar problem:
https://travis-ci.community/t/objective-c-api-availability-checks-give-wrong-results/6144

megastep:
Interesting, but I don't think it's really it though - mostly because on GitHub Actions the Mac environment is already running macOS 10.15 and is fairly up to date with Xcode 12...

https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

Have you tried just doing `swift build` from the root of the AudioKit directory? That's how it's built in Actions, and this is quite a different build environment than Xcode.

megastep:
FYI I just checked out your branch and it does give me the same errors on the command line.

yuriizubkov:
@megastep Thanks for checking. I am reading now about how to add a compile time availability check instead of a runtime check.

megastep:
I think I figured out your problem. You use this:    `@available(iOS 14.0, *)`

However, that is not good enough. `swift build` runs on a macOS target, and these APIs are only available on macOS 11, so you need to explicitly narrow down for the other platforms (including tvOS though that's taken care of by the entire file not being available).

If you compile AK targeting macOS even in Xcode you should see the problem.

yuriizubkov:
@megastep Ha! Thank you very much! I'll try to reproduce this on my local branch.

yuriizubkov:
@megastep  Yes I got the same errors with `swift build`. 
And this is interesting.
I tried to include only iOS and Mac OS with conditional compilation block:
```
#if os(iOS) || os(macOS)
    @available(iOS 14.0, macOS 11.0, *)
    private func processUMPMessages(...){...}
#endif
```
But yes, you where right, we can't check OS version at compile time, only on runtime. So this is fails to compile as well.
But when I remove Mac OS from the compile time (and runtime) check - I am able to compile the code without errors. I think this might be because these new classes still marked as "Beta" in the documentation and not visible for compiler for some reason?
For example: https://developer.apple.com/documentation/coremidi/midieventlist/unsafesequence/iterator 

I think I can exclude Mac OS for now, since this is still beta on Mac OS. 
And I'll make another PR just for iOS?

PR #2303, opened 2020-10-16
New schedule function for AudioPlayer and exposed PlayerNode's volume 

warpling:
* Exposes a new function for scheduling files from URLs while also being able to pass options like `.looping`. I would recommend removing the file scheduler or modifying it to also use a buffer so options can be passed to it; if not using a buffer is desired by some we should perhaps document more why the two are different.

* Exposes volume variable


aure:
This PR comes with no tests but adds features and so cannot be merged into AK5. It doesn't mean its a bad PR, just needs tests and I'll leave it open with the hope that tests get added someone in the community, (maybe me, maybe the PR author, maybe someone else).

warpling:
@aure is it expected that the completion handler is called for a looping player at the end of each loop? I'm not sure I know how to add this.

aure:
> @aure is it expected that the completion handler is called for a looping player at the end of each loop? I'm not sure I know how to add this.

You wrote the method signature, you tell me :)

PR #2306, opened 2020-10-19
Fixed AKMIDIInstrument conformance to AKMIDIListener

antonypoderechin:
This Pull Request contains 2 fixes:

**1. An issue with AKMIDIInstrument**
AKMIDIInstrument has broken conformance to AKMIDIListener. Because of these `noteOff` events were received by the AKMIDIInstrument, but `noteOn` events were received by the default implementation of `AKMIDIListener`.

For example, the following pseudocode works wrong:
```
let midi = AKMAnager.midi
midi.openInput()
let instrument: AKAudioUnitInstrument = createInstrument()
midi.addListener(instrument)
```
After the proposed fix everything works correct and `noteOn` and `noteOff` events are handled in AKMIDIInstrument.

**2. Debugging improvement**
In order to debug the AudioKit code, I removed optimization from the Debug configuration. With optimizations an xcode debugger sometimes can't make step-in to the AudioKit.


aure:
Cool, the midi listener was already fixed on v5, but it never made it back here.

PR #2320, opened 2020-11-03
added sample rate setter test

ryanfrancesconi:


ryanfrancesconi:
will try to see why those tests fail

PR #2321, opened 2020-11-03
V5 samplerate changes

ryanfrancesconi:
Here's some work on fixing the problem of a static 44.1 rate. I think the main issue overall is using the default engine.mainMixerNode and letting the engine create it itself. I'm not sure but that seems to stay at 44.1 in my tests.

This isn't super elegant in terms of the changes to AudioEngine.output, but I believe it works. All of the tests are assuming a sample rate of 44.1k in their md5 hashes, so it's not possible to add a test that changes Settings.audioFormat at this time. I do have the test in there but it's commented out so the rest of the tests don't fail after that one is run.


ryanfrancesconi:
my xcode indenting is set to tabs not spaces. I'm unclear if this isn't translating

wtholliday:
With the changes to connections, does this break our "can't create an invalid graph" invariant? That was a really nice property to have. I think it would be a shame to loose it. What happens when you call detach while it's running? I'm having trouble understanding this PR because it makes various formatting changes (should be a separate PR), and other changes which don't seem to be central to its goal (what do connection changes have to do with sample rate?).

Sources/AudioKit/Nodes/Node.swift
```diff
@@ -65,34 +65,35 @@ open class Node {
         }
     }
 
-    func makeAVConnections() {
+    public func makeAVConnections() {
```
wtholliday, line 68: Why is this public?

ryanfrancesconi: sorry, the sample rate changes got mixed into the other access experiments. that doesn't need to be. though subclassing Node from outside of AudioKit may have some challenges.

let me quarantine just the sample rate thing

PR #2322, opened 2020-11-04
V5 engine output samplerate

ryanfrancesconi:
this is mostly about the AudioEngine.avEngine.mainMixerNode that was previously being used for AudioEngine.output. This mixer isn't adopting Settings.audioFormat. I'm unclear other than manually managing this final mixer instance ourselves how you would control the format of that mainMixerNode, but perhaps there is a much simpler solution that I missed.


ryanfrancesconi:
actually. not ready.

PR #2327, opened 2020-11-10
removing the Settings.RampType enum.

ryanfrancesconi:
probably left in there as it looks important.
but it's not any longer.


aure:
Thanks for your diligence!

PR #2332, opened 2020-11-21
Default parameter for MIDISampler init

btfranklin:
Added default parameter value to MIDISampler initializer to enable instance creation without a name being provided explicitly. The enableMIDI method already works like this, but this extends the same intent to the initializer itself.


aure:
Thanks!

PR #2340, opened 2020-12-03
Add a name to mixer to assist in debugging

hoopes:
Hello,

This is one of my few PRs in the real world, so be gentle.

My intent was to start using the `NamedNode` protocol introduced recently by @btfranklin , and expand it to more nodes, starting with the `Mixer` node.

A project often has a bunch of `Mixer` nodes in the tree, and using the `connectionTreeDescription` is less useful if they're all just called `Mixer` in the resulting tree string. If their name was also printed (and the programmer took the time to name each mixer node), this would allow a user to understand what their tree looked like more easily.

Additionally, please note that I tried to make the `name` optional - if you don't want to set it, you shouldn't need to, and the default behavior will remain.

My swift knowledge isn't the deepest I have, so for the `name! as CFString` bits I followed xcode's suggestions, and I'm not sure if that's the best move.

In any case, please let me know if this is an avenue I could/should continue pursuing, adding to other nodes along the way.

Thanks!

### DIFF
--- a/Sources/AudioKit/MIDI/MIDIInstrument.swift
+++ b/Sources/AudioKit/MIDI/MIDIInstrument.swift
@@ -15,7 +15,7 @@ open class MIDIInstrument: PolyphonicNode, MIDIListener, NamedNode {
     open var midiIn = MIDIEndpointRef()
 
     /// Name of the instrument
-    open var name = "AudioKit MIDI Instrument"
+    open var name: String? = "AudioKit MIDI Instrument"
 
     /// Active MPE notes
     open var mpeActiveNotes: [(note: MIDINoteNumber, channel: MIDIChannel)] = []
@@ -38,8 +38,8 @@ open class MIDIInstrument: PolyphonicNode, MIDIListener, NamedNode {
     ///   - name: Name to connect with
     ///
     open func enableMIDI(_ midiClient: MIDIClientRef = MIDI.sharedInstance.client,
-                         name: String = "AudioKit MIDI Instrument") {
-        CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
+                         name: String? = "AudioKit MIDI Instrument") {
+        CheckError(MIDIDestinationCreateWithBlock(midiClient, name! as CFString, &midiIn) { packetList, _ in
             for e in packetList.pointee {
                 e.forEach { (event) in
                     self.handle(event: event)
--- a/Sources/AudioKit/MIDI/MIDINode.swift
+++ b/Sources/AudioKit/MIDI/MIDINode.swift
@@ -14,7 +14,7 @@ open class MIDINode: Node, MIDIListener, NamedNode {
     open var midiIn = MIDIEndpointRef()
 
     /// Name of the instrument
-    open var name = "MIDINode"
+    open var name: String? = "MIDINode"
 
     private var internalNode: PolyphonicNode
 
@@ -41,8 +41,8 @@ open class MIDINode: Node, MIDIListener, NamedNode {
     ///   - name: Name to connect with
     ///
     public func enableMIDI(_ midiClient: MIDIClientRef = MIDI.sharedInstance.client,
-                           name: String = "MIDINode") {
-        CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
+                           name: String? = "MIDINode") {
+        CheckError(MIDIDestinationCreateWithBlock(midiClient, name! as CFString, &midiIn) { packetList, _ in
             for e in packetList.pointee {
                 let event = MIDIEvent(packet: e)
                 guard event.data.count > 2 else {
--- a/Sources/AudioKit/MIDI/MIDISampler.swift
+++ b/Sources/AudioKit/MIDI/MIDISampler.swift
@@ -17,7 +17,7 @@ open class MIDISampler: AppleSampler, NamedNode {
     open var midiIn = MIDIEndpointRef()
 
     /// Name of the instrument
-    open var name = "MIDI Sampler"
+    open var name: String? = "MIDI Sampler"
 
     /// Initialize the MIDI Sampler
     ///
@@ -38,8 +38,8 @@ open class MIDISampler: AppleSampler, NamedNode {
     ///   - name: Name to connect with
     ///
     public func enableMIDI(_ midiClient: MIDIClientRef = MIDI.sharedInstance.client,
-                           name: String = "MIDI Sampler") {
-        CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
+                           name: String? = "MIDI Sampler") {
+        CheckError(MIDIDestinationCreateWithBlock(midiClient, name! as CFString, &midiIn) { packetList, _ in
             for e in packetList.pointee {
                 e.forEach { (event) in
                     if event.length == 3 {
--- a/Sources/AudioKit/Nodes/Mixing/Mixer.swift
+++ b/Sources/AudioKit/Nodes/Mixing/Mixer.swift
@@ -4,9 +4,13 @@ import AVFoundation
 import CAudioKit
 
 /// AudioKit version of Apple's Mixer Node. Mixes a varaiadic list of Nodes.
-public class Mixer: Node, Toggleable {
+public class Mixer: Node, Toggleable, NamedNode {
+
     /// The internal mixer node
     fileprivate var mixerAU = AVAudioMixerNode()
+
+    /// Name of the mixer
+    open var name: String?
 
     /// Output Volume (Default 1)
     public var volume: AUValue = 1.0 {
@@ -33,17 +37,18 @@ public class Mixer: Node, Toggleable {
     }
 
     /// Initialize the mixer node with no inputs, to be connected later
-    public init(volume: AUValue = 1.0) {
+    public init(volume: AUValue = 1.0, name: String? = nil) {
         super.init(avAudioNode: mixerAU)
         self.volume = volume
+        self.name = name ?? self.name
     }
 
     /// Initialize the mixer node with multiple inputs
     ///
     /// - parameter inputs: A variadic list of Nodes
     ///
-    public convenience init(_ inputs: Node...) {
-        self.init(inputs.compactMap { $0 })
+    public convenience init(_ inputs: Node..., name: String? = nil) {
+        self.init(inputs.compactMap { $0 }, name: name)
     }
 
     // swiftlint:enable force_unwrapping
@@ -52,8 +57,8 @@ public class Mixer: Node, Toggleable {
     ///
     /// - parameter inputs: An array of Nodes
     ///
-    public convenience init(_ inputs: [Node]) {
-        self.init()
+    public convenience init(_ inputs: [Node], name: String? = nil) {
+        self.init(name: name)
         connections = inputs
     }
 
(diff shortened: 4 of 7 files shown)


hoopes:
The CI is failing for an unrelated bug.

Sources/AudioKit/MIDI/MIDIInstrument.swift
```diff
@@ -15,7 +15,7 @@ open class MIDIInstrument: PolyphonicNode, MIDIListener, NamedNode {
     open var midiIn = MIDIEndpointRef()
 
     /// Name of the instrument
-    open var name = "AudioKit MIDI Instrument"
+    open var name: String? = "AudioKit MIDI Instrument"
```
btfranklin, line 18: I don't agree with making `name` optional. Since a default is provided, this would only be `nil` if the user expressely _removed_ the name and set it to `nil`, which doesn't seem like a case that should be supported or, at the very least, encouraged. Optionals should only be used when there's a good reason for them, and I don't see any good reason here.

Sources/AudioKit/MIDI/MIDIInstrument.swift
```diff
@@ -38,8 +38,8 @@ open class MIDIInstrument: PolyphonicNode, MIDIListener, NamedNode {
     ///   - name: Name to connect with
     ///
     open func enableMIDI(_ midiClient: MIDIClientRef = MIDI.sharedInstance.client,
-                         name: String = "AudioKit MIDI Instrument") {
-        CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
+                         name: String? = "AudioKit MIDI Instrument") {
```
btfranklin, line 41: Here again, it seems very strange and undesirable to making `name` optional. A default is already provided, so if no value is passed in, the default is used. This would mean the only way to get `nil` here would be for the caller to explicitly pass in `name: nil`.

Sources/AudioKit/MIDI/MIDINode.swift
```diff
@@ -14,7 +14,7 @@ open class MIDINode: Node, MIDIListener, NamedNode {
     open var midiIn = MIDIEndpointRef()
 
     /// Name of the instrument
-    open var name = "MIDINode"
+    open var name: String? = "MIDINode"
```
btfranklin, line 17: Would not make this optional. See previous comments for rationale.

Sources/AudioKit/MIDI/MIDINode.swift
```diff
@@ -41,8 +41,8 @@ open class MIDINode: Node, MIDIListener, NamedNode {
     ///   - name: Name to connect with
     ///
     public func enableMIDI(_ midiClient: MIDIClientRef = MIDI.sharedInstance.client,
-                           name: String = "MIDINode") {
-        CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
+                           name: String? = "MIDINode") {
```
btfranklin, line 44: Would not make this optional. See previous comments for rationale.

Sources/AudioKit/MIDI/MIDISampler.swift
```diff
@@ -17,7 +17,7 @@ open class MIDISampler: AppleSampler, NamedNode {
     open var midiIn = MIDIEndpointRef()
 
     /// Name of the instrument
-    open var name = "MIDI Sampler"
+    open var name: String? = "MIDI Sampler"
```
btfranklin, line 20: Would not make this optional. See previous comments for rationale.

Sources/AudioKit/MIDI/MIDISampler.swift
```diff
@@ -38,8 +38,8 @@ open class MIDISampler: AppleSampler, NamedNode {
     ///   - name: Name to connect with
     ///
     public func enableMIDI(_ midiClient: MIDIClientRef = MIDI.sharedInstance.client,
-                           name: String = "MIDI Sampler") {
-        CheckError(MIDIDestinationCreateWithBlock(midiClient, name as CFString, &midiIn) { packetList, _ in
+                           name: String? = "MIDI Sampler") {
```
btfranklin, line 41: Would not make this optional. See previous comments for rationale.

Sources/AudioKit/Nodes/Mixing/Mixer.swift
```diff
@@ -4,10 +4,14 @@ import AVFoundation
...
 
 /// AudioKit version of Apple's Mixer Node. Mixes a varaiadic list of Nodes.
-public class Mixer: Node, Toggleable {
+public class Mixer: Node, Toggleable, NamedNode {
+
     /// The internal mixer node
     fileprivate var mixerAU = AVAudioMixerNode()
 
+    /// Name of the mixer
+    open var name: String?
```
btfranklin, line 13: I do like the idea of adding name support here. However, rather than making this optional, I think a better solution would be to follow the existing convention of providing a meaningful default. Following the existing convention, I would recommend using "Mixer" for the default name.

Sources/AudioKit/Nodes/Mixing/Mixer.swift
```diff
@@ -33,17 +37,18 @@ public class Mixer: Node, Toggleable {
     }
 
     /// Initialize the mixer node with no inputs, to be connected later
-    public init(volume: AUValue = 1.0) {
+    public init(volume: AUValue = 1.0, name: String? = nil) {
```
btfranklin, line 40: I would recommend making this non-optional, but provide the default for the node. "Mixer" is a good choice that follows the existing conventions.

Sources/AudioKit/Nodes/Mixing/Mixer.swift
```diff
@@ -33,17 +37,18 @@ public class Mixer: Node, Toggleable {
...
         super.init(avAudioNode: mixerAU)
         self.volume = volume
+        self.name = name ?? self.name
```
btfranklin, line 43: I'm not sure what the goal was here, but this approach can be simplified to just `self.name = name` if you provide a default on the parameter and make it non-optional.

Sources/AudioKit/Nodes/Mixing/Mixer.swift
```diff
@@ -33,17 +37,18 @@ public class Mixer: Node, Toggleable {
...
     }
 
     /// Initialize the mixer node with multiple inputs
     ///
     /// - parameter inputs: A variadic list of Nodes
     ///
-    public convenience init(_ inputs: Node...) {
-        self.init(inputs.compactMap { $0 })
+    public convenience init(_ inputs: Node..., name: String? = nil) {
```
btfranklin, line 50: Same feedback here about making this non-optional with a good default.

Sources/AudioKit/Nodes/Mixing/Mixer.swift
```diff
@@ -52,8 +57,8 @@ public class Mixer: Node, Toggleable {
     ///
     /// - parameter inputs: An array of Nodes
     ///
-    public convenience init(_ inputs: [Node]) {
-        self.init()
+    public convenience init(_ inputs: [Node], name: String? = nil) {
+        self.init(name: name)
```
btfranklin, line 61: Same feedback here.

Sources/AudioKit/MIDI/MIDINode.swift
```diff
@@ -41,8 +41,8 @@ open class MIDINode: Node, MIDIListener, NamedNode {
...
+        CheckError(MIDIDestinationCreateWithBlock(midiClient, name! as CFString, &midiIn) { packetList, _ in
```
btfranklin, line 45: Making `name` non-optional will remove the need to force-unwrap this, and if it's `nil` (a possibility introduced by making `name` an optional value) this will fail anyway. That is actually a pretty good example of why to avoid using optionals when possible.

Sources/AudioKit/MIDI/MIDISampler.swift
```diff
@@ -38,8 +38,8 @@ open class MIDISampler: AppleSampler, NamedNode {
...
+        CheckError(MIDIDestinationCreateWithBlock(midiClient, name! as CFString, &midiIn) { packetList, _ in
```
btfranklin, line 42: See earlier comment about the force-unwrap here.

btfranklin (changes requested):
I think using the `NamedNode` facility on more nodes is the right move, and I agree that it'll make the connection tree output much more useful. It will also encourage good organization of node trees with some semantic information attached, which is almost always a good thing.

However, I recommend avoiding the use of optionals throughout this approach, and instead, following the existing convention that has been established by the node types that were already using `NamedNode`, which is to provide well-formed non-optional default name values.

Sources/AudioKit/MIDI/MIDIInstrument.swift
hoopes: The intent here was to avoid the `connectionTreeDescription` printing something like `Mixer ("Mixer")` if i were to provide `"Mixer"` as the default. The `connectionTreeDescription` (which i cannot thank you enough for getting the ball rolling on), prints the type by default, and the name if provided. Giving a default to the `Mixer` class would be fine, but do you think something like `"Mixer"` would be a reasonable default? That would then remove the optional name, and make it required.

btfranklin: Yeah, I feel you on that concern about stuff like `Mixer("Mixer")` but I think that's a pretty small price to pay, and it's really just a debugging tool anyway. The existing other `NamedNode` types have the same behavior as a result of their defaults, so it's not a step backwards. The other thing that gives me pause about making this optional is that `MIDIDestinationCreateWithBlock` which is clearly expecting some name to be provided. The ! force-unwrap will definitely fail as soon as one of these is set to `nil`, since ! means "I promise this won't be `nil` when you use it".

btfranklin: The is another possible approach we could think about here, which is to change the behavior of connectionTreeDescription to _only_ show the name if the node is a `NamedNode`. That would allow the defaults to be useful by themselves. The only concern I have about that is that it might make some named values be confusing since it would then hide their type, but I guess that would be up to the developer to avoid. For example, if a `Mixer` was named "1", the developer would have to understand that it would be better to name it "Mixer 1" or something like that if they want the tree to be readable.

What do you think about that alternative? It seems like it solves all the problems.

I'd like to know what @aure thinks about all this, too. I'm not actually an official AudioKit team member, and these are just my opinions.

hoopes: https://github.com/AudioKit/AudioKit/compare/v5-develop...hoopes:mixer-name-redux?expand=1

What do you think of this instead?

btfranklin: I think that's better, but I would not recommend having the name parameters be optional with defaults of nil in the initializers. If you make them non-optional but give them good defaults, then existing code can still ignore their addition. I would recommend modeling the approach closely around what's done in `MIDISampler`.

Your thoughts on this have really gotten me thinking that updating `connectionTreeDescription` will make the adoption of `NamedNode` much less painful. I think I'll play with that some, and it may feed into your changes here and make everything much smoother overall. But in the meantime, there's no reason why `Mixer` couldn't adopt naming using the exact same pattern as `MIDISampler`.

aure: Regardless of whether you guys are core team members, you are the ones who are using and developing and to some extent advocating for this useful (I like it too though!). What would be kind of cool is if the default name could always be unique. Perhaps the pointer location of the node would be a nice unique identifier to use?

Sources/AudioKit/MIDI/MIDIInstrument.swift
aure: I agree lets avoid optionals.

Sources/AudioKit/MIDI/MIDINode.swift
aure: AudioKit 5 did away with as many optionals as we could. Optionals have their purpose but they should be optional for a reason, not just by default.

aure:
When @btfranklin gives the go ahead I'll merge.

hoopes:
Closed in favor of the simpler https://github.com/AudioKit/AudioKit/pull/2342

PR #2342, opened 2020-12-04
Mixer name redux

hoopes:
My intent was to start using the NamedNode protocol introduced recently by @btfranklin , and expand it to more nodes, starting with the Mixer node.

A project often has a bunch of Mixer nodes in the tree, and using the connectionTreeDescription is less useful if they're all just called Mixer in the resulting tree string. If their name was also printed (and the programmer took the time to name each mixer node), this would allow a user to understand what their tree looked like more easily.

Updated note from the last one: @btfranklin is this more along the lines of what you would be looking for? Feels a little awkward to repeat the default a couple times, but whatever, like you said, it's just a debugging tool. And if you have updates that make all this easier, then even better.

### DIFF
--- a/Sources/AudioKit/Nodes/Mixing/Mixer.swift
+++ b/Sources/AudioKit/Nodes/Mixing/Mixer.swift
@@ -4,9 +4,12 @@ import AVFoundation
 import CAudioKit
 
 /// AudioKit version of Apple's Mixer Node. Mixes a varaiadic list of Nodes.
-public class Mixer: Node, Toggleable {
+public class Mixer: Node, Toggleable, NamedNode {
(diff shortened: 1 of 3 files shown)


btfranklin (approved):
I agree with you that it feels strange to have the default repeated in multiple places, especially since it's kind of just overriding the same exact value from place to place. We can re-examine that aspect of things, I think, in a future update. Other than that quirk, I think this looks great!

hoopes:
@aure is there any other ceremony to this? At work, i have to create a JIRA ticket, name the branch a certain way, blah blah blah - if there's anything else you need, please let me know.

PR #2344, opened 2020-12-10
AudioPlayer Seek File, Playback Progress Block, Convenience Methods

willsmillie:
- AudioPlayer seek to play audio from specific time
- Added playback progress block via a DisplayLink class (useful for updating UI components)
- And some helper methods like duration, sampleRate, and current (playback time)

TO DO:

- Seek buffers

### DIFF
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
@@ -66,11 +66,13 @@ public class AudioPlayer: Node {
     /// This also will be called when looping from disk,
     /// but no completion is called when looping seamlessly with a buffer
     public var completionHandler: AVAudioNodeCompletionHandler?
-    
+
+    #if !os(macOS)
     /// Progress block to be called via a CADisplayLink.
     /// Use this method to update your UI with info like progress bars, times, etc.
     public var progressBlock: AVAudioNodeCompletionHandler?
     private var displayLink: DisplayLink?
+    #endif
 
     /// The file to use with the player. This can be set while the player is playing.
     public var file: AVAudioFile? {
@@ -222,11 +224,13 @@ public class AudioPlayer: Node {
             schedule(at: when)
         }
 
-        displayLink = DisplayLink.init({
+        #if !os(macOS)
+        displayLink = DisplayLink {
             DispatchQueue.main.async { [weak self] in
                 self!.progressBlock?()
             }
-        })
+        }
+        #endif
         
         playerNode.play()
         isPlaying = true
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
@@ -299,8 +299,7 @@ extension AudioPlayer {
     @available(*, deprecated, renamed: "schedule(at:)")
     public func scheduleBuffer(_ buffer: AVAudioPCMBuffer,
                                at when: AVAudioTime?,
-                               options: AVAudioPlayerNodeBufferOptions = [])
-    {
+                               options: AVAudioPlayerNodeBufferOptions = []) {
         self.buffer = buffer
         isLooping = options == .loops
         schedule(at: when)
@@ -309,8 +308,7 @@ extension AudioPlayer {
     @available(*, deprecated, renamed: "schedule(at:)")
     public func scheduleBuffer(url: URL,
                                at when: AVAudioTime?,
-                               options: AVAudioPlayerNodeBufferOptions = [])
-    {
+                               options: AVAudioPlayerNodeBufferOptions = []) {
         guard let buffer = try? AVAudioPCMBuffer(url: url) else {
             Log("Failed to create buffer", type: .error)
             return
@@ -320,8 +318,7 @@ extension AudioPlayer {
 
     @available(*, deprecated, renamed: "schedule(at:)")
     public func scheduleFile(_ file: AVAudioFile,
-                             at when: AVAudioTime?)
-    {
+                             at when: AVAudioTime?) {
         self.file = file
         schedule(at: when)
     }
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
@@ -31,7 +31,7 @@ public class AudioPlayer: Node {
 
     /// The current sample rate of the buffer or file
     public var sampleRate: Double {
-        return (isBuffered ? buffer?.format.sampleRate : file!.processingFormat.sampleRate)!
+        return (isBuffered ? buffer?.format.sampleRate : file?.processingFormat.sampleRate) ?? 0
     }
 
     /// The duration of the buffer or file
@@ -54,9 +54,10 @@ public class AudioPlayer: Node {
     private var elapsedTimeOffset: Double = 0
     public var currentTime: TimeInterval {
         if isPlaying {
-            let nodeTime = playerNode.lastRenderTime
-            let playerTime = playerNode.playerTime(forNodeTime: nodeTime!)
-            return Double(elapsedTimeOffset) + (Double(playerTime!.sampleTime) / playerTime!.sampleRate)
+            if let nodeTime = playerNode.lastRenderTime {
+                let playerTime = playerNode.playerTime(forNodeTime: nodeTime)
+                return Double(elapsedTimeOffset) + (Double(playerTime!.sampleTime) / playerTime!.sampleRate)
+            }
         }
         return 0
     }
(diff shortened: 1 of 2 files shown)


aure:
Hi there, this looks like a good start. There are a number of things to attend to. How would you like to proceed? I can either just tell you stuff to change, or you can add me as a collaborator to your fork and I can make commits to it and have it appear in this PR. 

The first issue is that this PR does not pass tests because you included UIKit which does not exist on macOS. To use UIKit you have to check that you're on appropriate operating system. This is done in several places in AudioKit. 

The next issue is that you're adding a feature to AudioKit without adding a test of that feature.  The test suite is pretty easy to use, I can help you out if you need me to.

Finally if you want to join the AudioKit slack group, send me an email and I can add you.

willsmillie:
Thanks @aure! I don't mind giving it a shot myself, however if it's easier and quicker for you I can add you. I just resolved the UIKit issue. 

Next would be the tests which I don't actually have any experience with but wouldn't mind giving a shot.

Lastly, email for slack is will@smillie.me

Thanks!

KieranHarper:
Drive by comment – I'm new to AudioKit but these changes just helped me hack seeking into my local project. However I wonder if an even more useful approach might be a current time property you can set, regardless of playing state? Then if you play() later that's where it starts. I had to adjust these changes a little to get seeking while paused to not resume by itself, and also meanwhile report the seeked time as the current time.

royherma:
@KieranHarper How were you able to use this PR in your code? I want to as welL!

KieranHarper:
> @KieranHarper How were you able to use this PR in your code? I want to as welL!

I have a local checkout of the repo, which I'm able to edit directly and try in my project. You could check out this specific branch from willsmillie's fork if you want it as-is, but in my case I just applied that parts of the changes here that I wanted to try. If you've installed the package into your project using SPM, you can get it to point at your local repo by editing the project.pbxproj file and changing the path, as per Jared Sinclair's comment [in this thread](https://forums.swift.org/t/how-to-add-local-swift-package-as-dependency/26457/24).

aure:
People really want this, can we make this pass tests?

willsmillie:
Yes, I'll have some time today.

aure:
@ryanfrancesconi I got this to at least pass tests, but could you have a look and if you approve, I will merge. It would be great if there were something that worked with macOS.

ryanfrancesconi:
> @ryanfrancesconi I got this to at least pass tests, but could you have a look and if you approve, I will merge. It would be great if there were something that worked with macOS.

Two main things would be that the display link shouldn't be in a player class. If you have 200 players you want 200 display links?

 I would prefer that the player have a delegate that could send progress events if desired. In general this looks very much like a hack for a specific project rather than a clear addition. I use a CADisplayLink timer in my own work, but that is external to all players, and I only need one to update the screen with.

Also, force unwrapping all those optionals isn't good. It's better to let them silently fail with ? or strong check them to make sure the properties are valid.

Thanks!

ryanfrancesconi (commented):
Two main things would be that the display link shouldn't be in a player class. If you have 200 players you want 200 display links?

I would prefer that the player have a delegate that could send progress events if desired. In general this looks very much like a hack for a specific project rather than a clear addition. I use a CADisplayLink timer in my own work, but that is external to all players, and I only need one to update the screen with.

Also, force unwrapping all those optionals isn't good. It's better to let them silently fail with ? or strong check them to make sure the properties are valid.

Thanks!

willsmillie:
Good call @ryanfrancesconi 

I can replace it with a delegate, or could also do a block. Any preference or reason to go with one over the other? I'm thinking just one for playerDidStart, and playerDidStop. Does that sound like a good alternative?

ryanfrancesconi:
@willsmillie honestly, you shouldn't need these types of progress methods in the player. What exactly are you making?

Basically a player is generally controlled from a larger set of logic for playback, so it's preferable to put your delegate methods in your application level. For the most part you know if a player is started or stopped. What you don't know is the completion event, which is already handled in the player.

If you are wanting to update a playhead in a waveform type of editor, I recommend keeping this timing logic outside of your player class. The reason that hostTime is a thing is for the purpose of syncing different types of play events. For example, you can base your timer class off of hostTime and use AVAudioTime addition to track the position of the playhead rather than looking at the playerTime as you have added. It's not a bad thing, but you probably will no longer have the correct time if you are looping in a buffer after it goes past the end - the numbers will keep going up, not go back to 0.

ryanfrancesconi:
I wouldn't dispatch the completion handler to main like that personally. Though the completion handler comes in on a thread other than main, you're introducing an unknown amount of latency by dispatching async. 

I would remove all the display link stuff. I really don't think that belongs in this class. I can give you more info on how to handle this situation differently if interested.

ryanfrancesconi:
Don't use AVAudioNodeCompletionHandler as a type for a progress block. That's pretty confusing to read. Just use the underlying () -> Void

ryanfrancesconi:
on macOS the analog to CADisplayLink is CVDisplayLink, which is a bit more complicated to manage via swift (as it's really C) but works well. Should Ak wanto use use this method of timer I'd recommend a generic cross platform protocol that goes to CA or CV depending on the platform. But the api would be the same.

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -284,3 +323,42 @@ extension AudioPlayer {
         schedule(at: when)
     }
 }
+
+// Extension for seeking to a specific AVAudioTime
+extension AudioPlayer {
+    /// Play the audio player from a specific time
+    /// - Parameters:
+    ///   - time: The time at which the player will seek to
+    public func seek(time: Double) {
```
thejeff77, line 332: This is great.

One comment/suggestion: The way I read this, the seek function has a side affect of always playing. If the song was paused or stopped, seeking will play the song at the specified position. 

Personally with the way this function performs, I'd name it "seekAndPlay", which will fully describe what this function is doing. 

Question: Is it possible to make this function state aware? I.E. seek to the specified position without playing only when paused, so the next call to player.play() would start at the given position? This way the function caller won't need to hold onto the desired seek time until they are ready to play.

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -40,15 +50,30 @@ public class AudioPlayer: Node {
...
+    /// Used to get the correct current time, after seeking
+    private var elapsedTimeOffset: Double = 0
+    public var currentTime: TimeInterval {
+        if isPlaying {
+            if let nodeTime = playerNode.lastRenderTime,
+               let playerTime = playerNode.playerTime(forNodeTime: nodeTime) {
+                return Double(elapsedTimeOffset) + (Double(playerTime.sampleTime) / playerTime.sampleRate)
+            }
+        }
+        return 0
```
thejeff77, line 62: Consistency: 

If the player is paused, Its currentTime can be a positive non-zero number. 

If the code you've written for the "isPlaying" block is unreliable when the player is paused, consider storing the currentTime in an internal temp variable in the pause() method before actually pausing the player.

thejeff77 (approved):
I'd like to try this out ASAP so a quick merge would be great. I've added some comments for your consideration.

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
ryanfrancesconi: that isn't really a seek, as jeff says, that is a play. Generally a seek would just set the player up at that time, similar to AVPlayer.seek: https://developer.apple.com/documentation/avfoundation/avplayer/1385953-seek

Honestly, these methods are still just replacing what was already made in AudioKit 4.

ryanfrancesconi: if you do want a seek, rather than calling node.play() call 

`node.prepare(withFrameCount frameCount: AVAudioFrameCount)`

aure:
Honestly this PR is getting old and out of date, and the original author hasn't responded, so I'll close it for now. I think it shouldn't be reopened but if parts of it were resubmitted separately we could get some in. Or if someone wanted to look at this code and make their own smaller PRs, that would be fine.

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -97,11 +122,15 @@ public class AudioPlayer: Node {
...
-        guard isPlaying, engine?.isInManualRenderingMode == false else { return }
+        guard isPlaying, !isPaused, engine?.isInManualRenderingMode == false else { return }
 
         scheduleTime = nil
-        completionHandler?()
         isPlaying = false
+        elapsedTimeOffset = 0
+
+        DispatchQueue.main.async { [weak self] in
+            self?.completionHandler?()
```
warpling, line 132: Is this something worth backporting until this PR is fixed? (or any other fixes in here for that matter)

Matt54:
PR for seeking the AudioPlayer:
https://github.com/AudioKit/AudioKit/pull/2428

PR #2352, opened 2020-12-18
Added a name to AudioKit Engine's built-in Mixer node

btfranklin:
Added a name to AudioKit Engine's built-in Mixer node. This makes it look much better in the connection tree, and it's nice to have a name for it anyway. Updated the unit tests to match.


aure:
Cool.

PR #2354, opened 2020-12-18
sequencer realtime safety

wtholliday:
Use circular buffer instead of mutex for sequencer engine events.

### DIFF
--- a/Sources/CAudioKit/Sequencing/SequencerEngine.mm
+++ b/Sources/CAudioKit/Sequencing/SequencerEngine.mm
@@ -9,9 +9,15 @@
 #include <stdio.h>
 #include <atomic>
 #include <mutex>
+#include "TPCircularBuffer.h"
 
 #define NOTEON 0x90
 #define NOTEOFF 0x80
+
+struct SequencerEvent {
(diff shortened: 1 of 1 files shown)


Sources/CAudioKit/Sequencing/SequencerEngine.mm
```diff
@@ -9,10 +9,16 @@
...
 #include <atomic>
 #include <mutex>
+#include "TPCircularBuffer.h"
 
 #define NOTEON 0x90
 #define NOTEOFF 0x80
 
+struct SequencerEvent {
+    bool notesOff = false;
+    double seekPosition = NAN;
```
eljeff, line 19: thats cool i've never seen NAN used on purpose

eljeff (approved):
simple, easy to understand change, thanks

Sources/CAudioKit/Sequencing/SequencerEngine.mm
wtholliday: First encountered NAN as a sentinel in this book: https://www.oreilly.com/library/view/21st-century-c/9781491904428/

PR #2355, opened 2020-12-18
Fixed error "Index out of range" for VirtualInputs and VirtualOutputs

Thomasv46:
Originally, virtualInput and virtualsOutputs where MidiPortRef with a value of 0
In the previous commit, I replaced `MidiportRef()` by `[MidiPortRef]()` wich create an empty array of MidiPortRef.
Replacing `[MidiPortRef]()` by `[MidiPortRef()]` should solve the issue by creating an array of MidiPortRef whose first value is 0


aure:
Thanks!

PR #2358, opened 2020-12-20
Changed use of word "framework" to "package"

btfranklin:
Changed use of word "framework" to "package" in references to AudioKitUI.


aure:
Thanks!

PR #2365, opened 2020-12-27
V5 develop

emurray2:
I am working on creating an Audio Unit effect built with AudioKit nodes. When I was working on this, I thought it might be useful to have a way to automatically add parameters for each node inside a custom Audio Unit. I ran into some trouble with parameters having the same address, so I added an offset to distinguish parameters if needed and organize them into groups based on the AudioKit nodes.

### DIFF
--- a/Sources/AudioKit/Internals/CoreAudio/AudioUnitBase.swift
+++ b/Sources/AudioKit/Internals/CoreAudio/AudioUnitBase.swift
@@ -66,10 +66,10 @@ open class AudioUnitBase: AUAudioUnit {
     }
 
     private var _parameterTree: AUParameterTree?
-    
+
     /// Parameter address offset - useful for grouping multiple AudioKit node parameters in a custom Audio Unit
     public var addressOffset: AUParameterAddress?
-    
+
     /// Parameter tree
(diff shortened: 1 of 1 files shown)


aure:
This seems like an okay solution to your problem, but since its not used in AudioKit elsewhere, it kind of falls into the category of "why?" Can you show me how you use it and maybe I can figure out something clearer?

emurray2:
@aure That makes sense. Yes, I would love to share what I'm working on. Let me organize my code a bit and set up a repository so you can check it out. I'll send you an invite once the repository is up.

aure:
I think for now I'll leave this as an open pr and then if by the time your demo project is totally finished, if this is needed, we'll merge it. I suspect you might find a few more things you'd like to change in AudioKit before you're done.

emurray2:
@aure That's a good idea. I've been studying some Apple documentation, and it looks like I may be able to setup a custom parameter `implementorValueObserver` which calls the `setValue` method for the underlying Audio Units. That may be a better way because I'm still trying to figure out how to get the Apple Audio Units which don't have an `internalAU` configured. I'll keep you updated on how things go.

Issue #2366, opened 2020-12-27
v5 Mixer crashes when adding 8+ inputs to it while playing


fquarters:
**Confirm you should be adding the bug here**
Not really sure If it should be posted here, but unless I'm doing something completely wrong, the problem is in the AudioKit since it crashes in extension method `AVAudioNode.connect(input:bus:format:)` defined in AudioEngine.swift

**Describe the bug**
When adding new nodes to a mixer that has already been started, engine crashes as soon as the mixer has 8+ connections. However, it doesn't crash if nodes were added before the mixer has started. Type of nodes being added doesn't matter.

**To Reproduce**
Execute function `play()` twice
```swift
    var testEngine: AudioEngine?
    var oscGroup: Mixer?
    let oscBatchSize = 5
    
    @IBAction func play(_ sender: Any) {
        
        if let oscMixer = oscGroup {
            
           for _ in 0..<oscBatchSize {
                
                let osc = Oscillator()
                osc.start()
// crashes here when oscMixer has 8+ connections
                oscMixer.addInput(osc)
                
            }
            
            
        } else {
            
            let engine = AudioEngine()
            let mainMixer = Mixer()
            engine.output = mainMixer
            
            let oscMixer = Mixer()
            mainMixer.addInput(oscMixer)
            
            for _ in 0..<oscBatchSize {
                
                let osc = Oscillator()
                osc.start()
                oscMixer.addInput(osc)
                
            }

            oscMixer.start()
            mainMixer.start()
            try! engine.start()
        
            self.oscGroup = oscMixer
            self.testEngine = engine
        }  
    }
```
**Expected behavior**
It doesn't crash.

**Additional context**
Stacktrace:
```
0   CoreFoundation                      0x00007fff37bfdb57 __exceptionPreprocess + 250
	1   libobjc.A.dylib                     0x00007fff70a445bf objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff37cac59e -[__NSCFString characterAtIndex:].cold.1 + 0
	3   CoreFoundation                      0x00007fff37b6e861 -[__NSArrayM objectAtIndexedSubscript:] + 169
	4   AVFAudio                            0x00007fff335a06ab _ZN17AUInterfaceBaseV39GetFormatEbj + 71
[...]

PR #2368, opened 2020-12-28
V5 develop: allow to manually resize Mixer's input busses array

fquarters:
Proposed fix (or rather a workaround) to a crash caused by a adding multiple new inputs to a mixer on the fly. See issue #2366


aure:
A test should be written to make sure this is a tested feature that doesn't get broken over time, but I'll accept the PR anyway.

PR #2375, opened 2021-01-03
FFTTap: Add Ability To Turn Off Normalization

Matt54:
By setting the isNormalized property to false, the fftData returned will no longer be normalized. Default keeps functionality the same as before.


aure:
Looks great. It would be good if this had a simple test with it. I'll do it later today unless you want to give it a shot. I'll wait to merge it until one of us does that though.

aure:
Great! Thanks!

PR #2376, opened 2021-01-03
Support for multiple virtual inputs and outputs and send midi to specific physical and virtual output

Thomasv46:
- Refactored creation of virtual inputs and outputs
- Fixed numerous bug introduced in previous pull request (add support for multiple virtual inputs and outputs)
- add support to send midi to only specific physical ans virtual output (default send to every open outputs and virtual outputs)

### DIFF
--- a/Sources/AudioKit/MIDI/MIDI+VirtualPorts.swift
+++ b/Sources/AudioKit/MIDI/MIDI+VirtualPorts.swift
@@ -102,7 +102,7 @@ extension MIDI {
            let result = MIDISourceCreate(client, virtualPortName as CFString, &virtualOutputs[virtualPortIndex])
             if result == noErr {
                 if virtualInputs.count <= virtualPortIndex {virtualInputs.append(0)}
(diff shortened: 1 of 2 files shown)


Sources/AudioKit/MIDI/MIDI+VirtualPorts.swift
```diff
@@ -160,30 +118,34 @@ extension MIDI {
...
+            }
+            virtualInputs.append(0)
+            return true
             }
-        }
         return false
     }
 
-    /// Closes the virtual output port, if created one already.
+    /// Closes the virtuals output ports, if created one already.
```
aure, line 137: I can fix this later, but virtual isn't pluralizable.

Thomasv46: Done. Sorry for that, english isn't my mother tongue

aure:
Looks good, thanks!

PR #2378, opened 2021-01-04
Make it possible to define custom node taps outside of module

alexandrefrancois:
Just made Basetap.doHandleTapBlock open so that custom taps can be properly defined outside of the AudioKit module.
Also had to update the access control for overrides in the already defined taps.


aure:
Seams reasonable. Thanks.

PR #2381, opened 2021-01-06
FFTTap: scale output based on frameCount

Matt54:
We should be returning fftData values that result in strictly negative amplitudes. After looking into Apple's documentation, it seems that a scaling factor is required. I used 1/n (n = frameCount) for this factor and after performing my amplitude calculations in FFTView they matched the same amplitude values that I was getting from Ableton Live's spectrum analyzer.


aure:
This looks great. Again as we're early in on this, I'm not so worried about hounds and stuff, and will do a bit of style review soon.

eljeff (commented):
Awesome thanks for continuing to improve this

Issue #2290, opened 2020-10-04
Feature Request: Setter for Oscillator Waveform


Matt54:
With the removal of AKOscillatorBank and the other banks, we seemingly now no longer have the ability to dynamically swap the waveform of our sound generator on the fly.

Can we simply add a setter for the waveform of an oscillator? If not, is there some alternative approach we could do to have this ability?

Removing the oscillator banks, with their waveform setters, breaks the core functionality of my wavetable synthesizer application.

aure:
AudioKit 4 is not disappearing, your app does not have to follow the V5 trajectory. The AK4 code for banks can be used as a model of functionality, but needs to be implemented in a thread safe and cleanly code way. I do not have time for this right now so someone else will have to step up.

Matt54:
We can close this issue with the addition of the DynamicOscillator.

PR #2391, opened 2021-01-19
Created DynamicOscillator

Matt54:
New oscillator that allows for dynamic setting of its wavetable by storing two tables and crossfading between them on wavetable set.


Matt54:
Not sure how to link an issue, but it should link to this:
https://github.com/AudioKit/AudioKit/issues/2290

aure:
I think you know what I'm going to say next... this definitely needs a test suite as it is a new feature.

PR #2392, opened 2021-01-19
 DynamicOscillator: Add methods to get the wavetable's float array

Matt54:
Added method of getting the stored floating point value array stored in the dynamic oscillator's wavetable.

The onSetWavetable callback allows for views to monitor changes in the wavetable array.

The getWavetableValues allows for a manual method of getting the wavetable array.


aure:
Maybe a short set of test functions for these new accessors? I'm not in love with onSetWavetable as a callback name. Seems to me that wavetableUpdateHandler may be more Swifty. Or something else.

Matt54:
@aure what's going on with the check being unsuccessful?

aure:
so, yeah, for posterity, let me just say that sometimes continuous integration can fail for no fault of the pull request code and then a maintainer just has to come in and hit the "re-run all jobs" button.  Thanks for adding the test and doing the name update, looks great.

PR #2396, opened 2021-01-22
Transient Shaper

emurray2:
I made my first node for AudioKit! It's a transient shaper which can be used as an Audio Unit. I implemented the code based off this Transient Designer Max 4 Live patch made by David Braun: https://maxforlive.com/library/device/1775/transient-designer.

To implement this as a node in AudioKit, I turned to recreating some of the Max objects from the `cyclone` library which are used in the Transient Designer patch: https://github.com/porres/pd-cyclone.

I'm not sure how the license works because I'm a beginner at this stuff, but I thought it would be worth it to sort through it. I would love to have this effect be part of AudioKit if possible! Please let me know what you think.

-Evan


aure:
I think the BSD License is okay as you have added it, but if someone complains, perhaps I'd have to remove it. All new nodes should also have a corresponding test file that puts the node through its paces. Let me know if you want to do this. Perhaps we could do it in office hours if you're not sure how to do it yourself?

emurray2:
> I think the BSD License is okay as you have added it, but if someone complains, perhaps I'd have to remove it. All new nodes should also have a corresponding test file that puts the node through its paces. Let me know if you want to do this. Perhaps we could do it in office hours if you're not sure how to do it yourself?

That sounds like a good plan! I would love to do this. I'll try coming to the office hours tomorrow if I have some time. If I can't make it, I can maybe look at an example of how the test files are set up and follow that.

emurray2:
I'm going to debug it some more after my classes and I'll test it again before I submit another PR.

emurray2:
Once I'm happy with how the test performs/how the audio sounds, should I add the `md5` to `ValidatedMD5s`?

aure:
> Once I'm happy with how the test performs/how the audio sounds, should I add the `md5` to `ValidatedMD5s`?

Yes

Sources/AudioKit/Nodes/Effects/Dynamics/TransientShaper.swift
```diff
@@ -0,0 +1,126 @@
+//
+//  TransientShaper.swift
+//  AudioKit
+//
+//  Created by Evan Murray on 1/6/21.
+//
```
aure, line 6: Can you update your headers to be the current 1 liner AudioKit headers. I hate these giant, useless, always getting out of date headers.

Sources/CAudioKit/Nodes/Effects/Dynamics/TransientShaperDSP.mm
```diff
@@ -0,0 +1,614 @@
+//
+//  TransientShaperDSP.m
+//  CAudioKit
+//
+//  Created by Evan Murray on 1/6/21.
+//
```
aure, line 6: Here too

Sources/CAudioKit/Nodes/Effects/Dynamics/TransientShaperDSP.mm
```diff
@@ -0,0 +1,614 @@
...
+
+#include "SoundpipeDSPBase.h"
+#include "ParameterRamper.h"
+#include "soundpipe.h"
```
aure, line 10: Do we need to include soundpipe.h? It seems like you could just include DSPBase.h and make it inherit from that.

Sources/CAudioKit/Nodes/Effects/Dynamics/TransientShaperDSP.mm
```diff
@@ -0,0 +1,614 @@
...
+// MARK: END Cyclone Slide
+
+enum TransientShaperParameter : AUParameterAddress {
+    TransientShaperParameterInputAmount,
+    TransientShaperParameterAttackAmount,
+    TransientShaperParameterReleaseAmount,
+    TransientShaperParameterOutputAmount,
+};
+
+class TransientShaperDSP : public SoundpipeDSPBase {
```
aure, line 269: public DSPBase should be plenty

aure:
Would love to see a cookbook demo of this.

emurray2:
> Would love to see a cookbook demo of this.

That's a great idea! That will be a good project for me to implement next weekend.

PR #2402, opened 2021-01-29
Porting android windows

sseyod:
Code changes that allow the Core AudioKit code to be compiled and linked for both Windows and Android.

I've added a file called PORTING which outlines the changes, and gives pointers on how Windows and Android developers can deploy the AudioKit code to their own projects.


Sources/CAudioKit/CoreAudio/DSPBase.mm
```diff
@@ -1,14 +1,22 @@
...
-#import <map>
-#import <string>
-
+#if __APPLE__
+#include "DSPBase.h"
+#include "ParameterRamper.h"
+#else // __APPLE__
+#include "DSPBase.h"
+#include "ParameterRamper.h"
+#endif // __APPLE__
```
aure, line 9: I don't understand how the if and else are different?

Sources/CAudioKit/CoreAudio/SoundpipeDSPBase.mm
```diff
@@ -1,6 +1,10 @@
 // Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
 
-#import "SoundpipeDSPBase.h"
+#if __APPLE__
+#include "SoundpipeDSPBase.h"
+#else // __APPLE__
+#include "SoundpipeDSPBase.h"
+#endif // __APPLE__
```
aure, line 7: Here as well?

Sources/CAudioKit/Nodes/Callback Instrument/CallbackInstrumentDSP.mm
```diff
@@ -1,8 +1,14 @@
 // Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
 
 #include "DSPBase.h"
-#import "TPCircularBuffer.h"
 
+#if __APPLE__
+#include "TPCircularBuffer.h"
+#else // __APPLE__
+#include "TPCircularBuffer.h"
+#endif // __APPLE__
```
aure, line 9: Ditto

Sources/CAudioKit/Nodes/Generators/Physical Models/ClarinetDSP.mm
```diff
@@ -1,6 +1,10 @@
 // Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
 
-#import "STKInstrumentDSP.hpp"
+#if __APPLE__
+#include "STKInstrumentDSP.hpp"
+#else // __APPLE__
+#include "STKInstrumentDSP.hpp"
+#endif // __APPLE__
```
aure, line 7: ?

sseyod: Hi! Thanks for the feedback. Yep, well spotted (this and the above)! I'm guessing those are easy for you to change after doing a merge? Not really sure how handling merge requests issues work yet!

NB: I have another set of improvements that I'll pull together as a separate pull request, once these are merged, that I think you'll find useful, as I've experimented more on iOS/macOS/tvOS/WIndows/Android (I write cross-platform audio apps).

Best wishes, Pete

sseyod: That includes some example cross-platform C++ code I'll add to the PORTING file.

megastep: FYI you can just keep pushing commits to your branch and they will show up in this pull request.

sseyod: @megastep Aha - many thanks, I didn't actually know that :)

Pete

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -247,20 +251,24 @@ extension AudioPlayer {
     /// or the player will call it when play() is called to load the audio data
     /// - Parameters:
     ///   - when: What time to schedule for
-    public func schedule(at when: AVAudioTime? = nil) {
+    ///   - completionCallbackType: Constants that specify when the completion handler must be invoked.
+    public func schedule(
+        at when: AVAudioTime? = nil,
```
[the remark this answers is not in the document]
sseyod: Fixed!

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -195,7 +195,11 @@ public class AudioPlayer: Node {
     /// - Parameters:
     ///   - when: What time to schedule for. A value of nil means now or will
     ///   use a pre-existing scheduled time.
-    public func play(at when: AVAudioTime? = nil) {
+    ///   - completionCallbackType: Constants that specify when the completion handler must be invoked.
+    public func play(
+        at when: AVAudioTime? = nil,
```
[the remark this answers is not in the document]
sseyod: Fixed!

Sources/CAudioKit/CoreAudio/DSPBase.mm
sseyod: Fixed!

Sources/CAudioKit/CoreAudio/SoundpipeDSPBase.mm
sseyod: Fixed!

Sources/CAudioKit/Nodes/Callback Instrument/CallbackInstrumentDSP.mm
sseyod: Fixed!

Sources/CAudioKit/Nodes/Generators/Physical Models/ClarinetDSP.mm
sseyod: Fixed and pushed!

sseyod:
Hi! Not sure how I can resolve that from GitHub? Pete

Sources/CAudioKit/Nodes/Generators/Synth/SynthDSP.mm
```diff
@@ -1,6 +1,11 @@
 // Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
 
-#import "SynthDSP.h"
+#if __APPLE__
+#include "SynthDSP.h"
+#else // __APPLE__
+#include "SynthDSP.h"
+#endif // __APPLE__
+
```
aure, line 8: Another one of these identical if/elses

aure:
Why did some of the files get renamed to from xxx.c to xxx_c.c?  Seems like that could have been an interim step that can now be reverted?

aure:
I'd love to see a repo of this in action on Windows and Androids projects. Or at least, maybe we could do a screen share and you can show me?

Sources/CAudioKit/AudioKitCore/Synth/CoreSynth.cpp
```diff
@@ -107,7 +127,7 @@ int CoreSynth::init(double sampleRate)
     data->segParameters[0].finalLevel = 0.2f;
     data->segParameters[0].seconds = 0.01f;
     data->segParameters[1].initialLevel = 0.2f;   // hold at 0.2 for 1 sec
-    data->segParameters[1].finalLevel = 0.2;
+    data->segParameters[1].finalLevel = 0.2F;
```
aure, line 130: This is minor but I've always seen float designated with a lowercase f, and all the other nearby values also have lowercase. There a few these added in this pr.

aure:
I also don't see any reference as to why all the files in the ugens directory were renamed to ugengs/ugen_xxx. Can they be renamed back? Just seems redundant.

Sources/CAudioKit/AudioKitCore/Synth/CoreSynth.cpp
sseyod: Will fix. Just a matter of coding style, really. All other code bases I've used stick to upper case. Doesn't matter to me!

sseyod:
Hi! ugengs/ugen_xxx. quesiton - this is critical for Windows.

If you have files called (say)
fred.c
Fred.cpp
FrED.cxx
etc.

You get a build error - even if they're in different folders. Just a peculiarity of Visual Studio!

So - when that happens, the solution is just to rename one of them.

I've had to rename a couple of files in my own codebase to accommodate AudioKit, which is no problem for me of course.

MSB8027 is the warning, I think.

Hoping that clarifies things,

Pete

Sources/CAudioKit/Nodes/Generators/Synth/SynthDSP.mm
sseyod: Aha, thanks. Fixed.

sseyod:
Hi @aure "I'd love to see a repo of this in action on Windows and Android projects. Or at least, maybe we could do a screen share and you can show me?"

I'm the developer of Wotja - https://intermorphic.com

I've had my eye on AudioKit for a while, and thought I should have a go at porting and trying out some of your units!

I've done a *lot* of porting / cross-platform work, so wasn't daunted by it, and you have a nice organisation to your code; only took me a day and a half or thereabouts; so it was a pleasure to share my porting changes back to you all.

Anyhow, I'm currently just trying to deploy one of your effects in my Wotja development stream. Hasn't taken too long, and hopefully will have it running in a cross-platform UI in a day or so.

I can keep you posted if you like - maybe invite to the Slack channel?

Best wishes, Pete

aure:
> I've had my eye on AudioKit for a while, and thought I should have a go at porting and trying out some of your units!
> I've done a _lot_ of porting / cross-platform work, so wasn't daunted by it, and you have a nice organisation to your code; only took me a day and a half or thereabouts; so it was a pleasure to share my porting changes back to you all.

Definitely, and its appreciated. My one concern is that these changes, while not breaking anything we are currently testing, are not themselves tested yet. For instance, when I merge this PR and then we continue developing, we have no way of testing whether our new changes have broken the Windows or Android build process. A test project, or ideally, a part of the continuous integration, testing the new targets would be great. I currently have no Windows or Android hardware to test on myself.

Happy to invite you to Slack, but I need your email address and full name to do so.

megastep:
My main concern with this is that there's a bit too much of bending over to please Visual Studio specifically. There should be other ways to build Windows libraries without relying on the idiosyncrasies of this specific tool, as popular as it is. It's been a long time since I've looked at this, but surely we could have a Windows build system that can produce a compiled library you can then just drop into VS? I'm sure you could get a LLVM setup that would allow to keep most of our code structure unchanged while producing Windows-compatilble binaries.

Porting should not need to involve major changes beyond what is strictly necessary IMHO. This is still first and foremost a project for Apple platforms.

megastep:
And Aure makes a good point about continuous integration - Github Actions definitely can handle Windows builds so I would be in favor of having a working build system in that environment so we can have tests running. At the same time that would force to make a clean build system instead of relying on VS specifically.

sseyod:
Hi Folks!

Thanks so much for your feedback, it is very interesting.

"""My one concern is that these changes, while not breaking anything we are currently testing, are not themselves tested yet."""

Yeah, I agree. One advantage of bringing-in developers from other platforms is that you'll attract devs who, for example, have experience in creating such tests for Windows / Android (not something I know about - I rely on Xcode for my unit tests; after all, the bulk of Wotja's code is 100% cross-platform C++; even the user interface!).

"""Happy to invite you to Slack, but I need your email address and full name to do so."""

Pete Cole - pete @ intermorphic dot com

"""My main concern with this is that there's a bit too much of bending over to please Visual Studio specifically"""

Sure, I take your point of course.

Looking back at the porting changes for both Android and Windows, they're really quite minor:
- renamed a very small number of files (in a minor way). FWIW, I have had a thought about how I might be able to avoid the renaming, and will make a point of looking at it. If as and when I can figure out a solution, my plan would be to share that with you all.
- added Windows compiler flags to hide a very large number of warnings - if the iOS code code were compiled with stricter flags, you'd have cleared these years ago I suppose!
- a couple of one-line bug-fixes shown-up by the MSVC compiler
- changing #import to #include (only where strictly required by the MSVC compiler)

"""Porting should not need to involve major changes beyond what is strictly necessary IMHO. This is still first and foremost a project for Apple platforms."""

Understood! Which platforms you choose to focus on is entirely your choice of course, it is your project! I would humbly suggest that by attracting the input from developers from other platforms, we might all be able to work together to make AudioKit even more powerful. That is just my own opinion, of course :)

I would suggest that involving a developer from another platform has already brought some concrete benefits:
- a couple of core bug fixes
- revealed some code (via the Windows compiler pragmas...) which might want to be "fixed" in the core code at some point

With an associated trade-off:
- a couple of files renamed (one-off change)
- some changes in #import -> #include

And some opportunities:
- more devs might get involved in the AudioKit project!
- increased future exposure etc.

Anyhow, thanks again for looking at my suggested code changes.

Best wishes,

Pete

PR #2405, opened 2021-01-30
Add case harmonic(partialAmplitudes: [Float])

niklassaers:
As discussed on Slack in #general, I'm looking for an alternative to the old AKFourierSeriesTableGenerator. This is my attempt to recreate it - add a type with an array of amplitudes as an option

To make this happen, I removed Int, Codable, CaseIterable from enum TableType and Codable from class Table. I could not see they were used as this in the repo, so I’m curious if this is a breaking change and for the motivation of adopting these. Making it codable again should probably be fine if needed, but if the enum needs to be represented by an Int, then I should probably find another approach

### DIFF
--- a/Sources/AudioKit/Internals/Table/Table.swift
+++ b/Sources/AudioKit/Internals/Table/Table.swift
@@ -3,7 +3,7 @@
 import AVFoundation
 
 /// Supported default table types
-public enum TableType: Int, Codable, CaseIterable {
+public enum TableType {
     /// Standard sine waveform
     case sine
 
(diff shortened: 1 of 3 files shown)


aure:
Since your removal of Int, Codable, CaseIterable, did not break any tests, whoever added them did not write a test case that would fail if they were not there, so I can only assume your change is okay and when/if someone comes back to complain, it will be on them to write the appropriate tests. Correstpondingly, you have added a feature with the new method that is not tested, so you'll need to add to this PR some sort of unit test for the new functionality. Hit me up on Slack if you need help writing the test.

niklassaers:
Added a test like the other render-tests. I tried it out in Cookbook and was pretty happy with the sound. What do you all think?

aure:
Okay I'll merge and see what breaks :)

PR #2407, opened 2021-02-01
Update PreparingSampleSets.md

maksutovic:
I added an example of a working SFZ file that works with AudioKit in order to help new users more easily/quickly begin working with Sampler. I have spent many, many, many hours chasing ghosts trying to get my sampler to work, only to find that I had added opcodes like <loval> AFTER my sample declaration in my <region> opcodes. I only discovered this after looking at Sampler+SFZ file and stepping through line by line to discover that the order of your opcodes matter, and I could not find any documentation in AudioKit to determine this. Also, other methods of generating SFZ files will not assume opcode order. I want to do an actual pull request in the code, however I find this is the lowest hanging fruit before messing with code that already works, that just needs to help inform the user that the order of <region> opcodes matter.


aure:
More docs are always welcomed, thank you.

PR #2419, opened 2021-02-07
bug fix -uninitialized AUMIDIEvent caused stream of unwanted midi …

misterinterrupt:
I introduced a bug!
By not creating the AUMIDIEvent, the initial values were some noise.
the event.length was not [nil] so the callback interpreted the event bytes.
it was sending these phantom bytes on every callback
I didn't notice it at first with a simple app that I was using to test my other code since I was only printing events that I was scheduling..

as penance, I created a test to assert that there are in fact no notes coming from a callback when there are none sequenced..


misterinterrupt:
reopened, I squashed and tried to fix those hound violations.. just lmk if the style comments are important.

aure:
Good catch, and great to provide the test.

PR #2422, opened 2021-02-08
Added mappedExp and mappedInverted

Matt54:
I added a method that does a mapping but then inverts the range. This helps with UI stuff where 0 is the top and height is the bottom.

Also, I added a method that goes from the logarithmic scale back to the linear scale. I called it mappedExp, but I'm not sure that this is an appropriate name. Suggestions for a new name are welcome


aure:
I'm not in love with these names either but I'll accept them for now.

PR #2428, opened 2021-02-15
AudioPlayer: Added Seek and Time Keeping

Matt54:
Added ability to seek the audio player as well as get the current time in the audio file of the player even after it seeks or pauses.


aure:
How about two tests for the new additions?

warpling:
Should stoppedTime be pausedTime since stopping seeks to 0 as far as I understand.

Matt54:
@warpling - I agree with the rename
@aure - I added some tests

royherma:
@aure So we have seeking available now?! 🙏

Issue #2409, opened 2021-02-02
V5 Sequencer cannot play notes at same position programmatically. 


maksutovic:
**Confirm you should be adding the bug here**
If your issue is implementing AudioKit in your own code, post your problem to [StackOverflow](https://stackoverflow.com/search?q=AudioKit) with the #AudioKit hashtag.

If you can make the issue happen on [AudioKit 5's Cookbook app](https://github.com/AudioKit/Cookbook/) then you can create an issue directly on that repository as well.

GitHub Issues should only be used for confirmed problems with AudioKit and
well-defined feature requests.

Sponsors will get more attention: https://github.com/sponsors/aure

**Describe the bug**
I posted this [issue on stack overflow](https://stackoverflow.com/questions/65997886/audiokit-v5-sequencer-cannot-play-notes-simultaneously-at-the-same-position-i). In essence, using the `add(noteNumber:, velocity:, channel:, position: , duration: , trackIndex: )` method of V5 `Sequencer`, if you add a note at the same position (to play a chord for example), only the first note will play. This is not the behavior exhibited by V4 AKAppleSequencer or AKSequencer. 

Much to my dismay, I tried loading a Midi file, and that will indeed play simultaneous notes at the same position.

Also, curiously, if you nudge the position of a note, say by 0.1, the effect of polyphony will sound with the sequencer.
**To Reproduce**
I have made a very simple sample project exhibiting the issue: https://www.dropbox.com/sh/vubyb9l4p1rxtrf/AAB8sQaKkpX5dPlD8TQjFq3qa?dl=0

**Expected behavior**
Using the `add` method on `Sequencer` a user is able to place multiple notes at the same position for the sequencer to playback.

**Screenshots**

**Details (please complete the following information):**
 - Type: all
 - OS: iOS 14.4
 - AudioKit Version 5.0 v5-main branch

**Additional context**
If this is intended behavior, I apologize, it is just a departure from how the previous iterations of Sequencer worked.

aure:
I don't think this is intended behavior. We will have a look.

misterinterrupt:
the issue was that in the sequencer's `process` call was sending note on msgs before note off msgs  

so if the same note is triggered in succession, the note on for the second msg comes before the note off from the first

e.g. trigger 4 whole notes one after the other, we hear the first note,  but never the other 3

in the original uploaded test, this was happening, but the chords sometimes played the same notes in succession like this.

[...]

PR #2431, opened 2021-02-18
fix simultaneous note-off/note-on msg ordering

misterinterrupt:
This is to assure that all note-offs that happen simultaneously with a note-on of the same note
I have set the sample offset to *1 sample earlier* than the sample time calculated from it's note's duration 

The note msg ordering was tested in an app and the tests have been adjusted to the new note ordering.
New tests were added to test a single repeating note and the exact scenario in [this issue](https://github.com/AudioKit/AudioKit/issues/2409)


misterinterrupt:
hmm, this shakerTest passes on my local machine

misterinterrupt:
alright, force-pushed the removal of an empty line to my branch and now the shaker test passes. 🍶

aure:
Thanks. We'll have to revisit this at some point I guess, but thanks for addressing it as is now.

PR #2434, opened 2021-02-20
Sequencer modifications

misterinterrupt:
- Use a BitSet for Running Status (add, remove, and check playing notes) 
- - removes the memory allocation
- - 
- Consolidate note processing loops
- Sort the MIDI Events by earliest time when the sequence is updated
- - not very expensive
- - done in the time domain of the ui/main thread
- - easily optimized, made lazier, cached as needed if it causes any kind of performance issue
- - easily customizable to emulate the behavior of preferred DAW note overlap style (re: what would ableton/garageband do?)
- SequencerEngine's process function only schedules events
- - code does not guess about or waste time on the message type or order
- - less complex, bug prone, and unmaintainable code (those hard to reason about conditionals in loops over events)
- A range of easy to re-introduce note ordering bugs are pushed into a swift function which is less error prone and more able to be customizable


misterinterrupt:
I have tested this in an app as well as in unit tests.
Open to adding more tests if requested.

PR #2435, opened 2021-02-20
Let controller present from UIBarButton too

niklassaers:
Make it so that BluetoothMIDIButton can present BTMIDICentralViewController from a UIBarButtonItem too.

Instead of just assuming the .next responder is a UIViewController, go down the chain until a UIViewController is found, or bail out

Not sure how to write a test for this as it fixes an internal issue when presenting a view controller, and there are no UITests in the project yet.

Also, not sure touchesEnded is the right place to handle the action - UIButtons tend to prefer actions rather than custom touch handling.


niklassaers:
I should add a bit to make it present nicely from popups on the iPad too. The origin CGRect needs to be set

aure:
Thanks for thinking about tests. But, you're kind of off the hook since its UI Stuff :)

PR #2439, opened 2021-03-02
Add engine running check to AudioPlayer

warpling:
This bug is the bane of my existence and my #1 bug in production still but perhaps this is the fix? 😅

It's often unclear what stops an engine (Siri, route change, etc) but I noticed there was already a check for `isRunning` in the looping code so it seems it would make sense to have it in `play()` as well.

### DIFF
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
@@ -208,8 +208,13 @@ public class AudioPlayer: Node {
     ) {
         guard !isPlaying || isPaused else { return }
 
-        guard playerNode.engine != nil else {
+        guard let engine = playerNode.engine else {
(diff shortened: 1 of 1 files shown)


ryanfrancesconi:
I don't think that is a full solution, but it might catch some cases. have you found it to help? What really needs to happen is on the application level and knowing what is stopping your engine.

warpling:
@ryanfrancesconi it definitely doesn't solve the root of the issue but it can prevent a fatal crash which seems like something AK should be trying to do whenever possible, right? I'm still mixed on how much of this very common audio interruption stuff AK should be handling (since every app needs to handle Siri, route changes, etc) but logging a warning and preventing a uncatchable exception seems preferable at least to me!

I have struggled with this "engine isRunning false" crash for years now. I try to handle engine restarts and interruptions in every way imaginable but there still seems to be some sort of race condition (at least that's my best guess) where an engine can get stopped between the time when it is checked for `isRunning` and a player being played.

PR #2447, opened 2021-03-22
Fix AudioPlayer scheduling when buffering

maxmuermann:
`AudioPlayer.schedule(at:completionCallbackType:)` used `nil` for the scheduled time when in buffering mode, rather than the provided AVAudioTime.


aure:
Thanks. This should have been under test coverage.

PR #2449, opened 2021-03-22
more robust sample time check

eljeff:
I was getting a low level AVAudio error about nodetime being invalid.
This addresses that before it gets that far.


warpling:
Nice! I've been getting some a rare and strange crashes recently… I wonder if this could be related to either?
* `Crashed: AVAudioPlayerNodeImpl.CompletionHandlerQueue`
* `Fatal Exception: com.apple.coreaudio.avfaudio
player did not see an IO cycle.`

eljeff:
@warpling I'm not sure these are related because the error I got was specifically 'nodeTime is invalid'.

warpling:
Ah 😅 well always happy to see the players get more robust regardless!

PR #2456, opened 2021-04-12
#2554 MIDI note offs for playing notes when sequencer stops

misterinterrupt:
I am forgoing any points for tests this time, since there is currently no precedent to test this correctly in code.
I have done some QA with an app I have for testing starting stopping, loops, no loops, and note trigger / loop boundary edge cases


aure:
I'm going to merge so @btfranklin has a chance to test.

PR #2457, opened 2021-04-13
Updated FormatConverter

ryanfrancesconi:
I wanted to bring this up to date as the previous version had the potential to leak `ExtAudioFileRef`'s from the `convertToPCM` function if there was an early exit due to an error or if the file didn't need to be converted.

The other change is to make it so that compressed -> compressed conversions can handle sample rate conversion as well.

Also added the test i use and made a base class for generating a more varied audio file.


aure:
Thank you!

PR #2458, opened 2021-04-14
Make Node.detach public

davebang:


davebang:
Apologies, created PR on wrong repo!

PR #2459, opened 2021-04-16
updates for MultiChannelInputNodeTap

ryanfrancesconi:


Sources/AudioKit/Internals/Utilities/Atomic.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    func unlock() {
+        pthread_rwlock_unlock(&lock)
+    }
+}
+
+/// A property wrapper that ensures atomic access to a value, meaning thread-safe with implicit serial read/write access.
+/// Multiple read accesses can potentially read at the same time, just not during a write.
+/// By using `pthread` to do the locking, this safer then using a `DispatchQueue/barrier` as there isn't a chance of priority inversion.
+@propertyWrapper
```
misterinterrupt, line 50: what do you think about warning or erroring out if this property wrapper is used on the main thread?

Sources/AudioKit/Internals/Utilities/Atomic.swift
```diff
@@ -0,0 +1,81 @@
...
+public final class Atomic<T> {
```
misterinterrupt, line 51: Could we consider using a name that is more specific to the underlying implementation, e.g. `ThreadLockedAccessor`, `LockingThreadAccessor`, `PThreadLockAcccess`, `ThreadLockProperty`, `PThreadBarrier`  ?

These names are not as pretty, but here are few thoughts that I think deserve some consideration:

- It is reasonable to assume the name`Atomic` would represent the full / commonly expected set of features/uses that atomics provide

- It may confuse the very useful mechanism used here with the atomic concepts from C, C++, or even Swift / possible future Swift (I've seen some [swift evolution work](https://swift.org/blog/swift-atomics/), but have not dug into it deeply).

- overloading the name `Atomic` may preclude the use of standardized or 3rd party packages that an AK user may want to include that use `Atomic`

- The property wrapper may/will show up in xcode autocomplete leading to possible unintended use resulting in undefined behavior / headaches for fast-movers or those with less knowledge of multithreading (maybe I missed something in the commits that makes it so this won't happen)

Perhaps you have already thought of all these or other things.
Even so, its pretty cool and instructive to have some github or inline documentation of AK's "framework design thinking"

ryanfrancesconi: these are good comments, thank you. i didn't actually write that wrapper, but was using it in the Tap class as it seemed preferable to a dispatch queue in this case. it's fine to keep it private to my class as well.

ryanfrancesconi: Changed to internal `ThreadLockedAccessor` - I would agree that `Atomic` shouldn't be a public name in AK.

misterinterrupt: Gotchya, sounds good!

ryanfrancesconi: thanks!

Issue #2460, opened 2021-04-16
AudioKit Tests do not run


eljeff:
Trying to run the tests locally - failing with no error.

1) new project
2) add ak v5-develop branch via SPM
3) drag in audiokit folder to debug locally
4) change target to AudioKit
5) click test tabs, and run tests
6) XCode says 'Build failed' with no error. Parent app builds fine, c-audiokit tests build and pass.

**Expected behavior**
tests should run

**Details (please complete the following information):**
MacBook Pro (Retina, 15-inch, Late 2013)
macos big sur  
I can get the tests to run if I roll back my branch to 22 Mar - haven't explored anything beyond that.

eljeff:
Found this in the logs: 
**bundle format unrecognized, invalid, or unsuitable**

Full log:
```
CodeSign /Users/jeffcooper/Library/Developer/Xcode/DerivedData/ak5tests-ezrjdtvhkqigsnadeukaxbuvdsry/Build/Products/Debug-iphonesimulator/AudioKit_AudioKitTests.bundle (in target 'AudioKit_AudioKitTests' from project 'AudioKit')
    cd /Users/jeffcooper/Documents/code/audiokit/audiokit\ v5/AudioKit
    export CODESIGN_ALLOCATE\=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/codesign_allocate
    
Signing Identity:     "-"

    /usr/bin/codesign --force --sign - --timestamp\=none /Users/jeffcooper/Library/Developer/Xcode/DerivedData/ak5tests-ezrjdtvhkqigsnadeukaxbuvdsry/Build/Products/Debug-iphonesimulator/AudioKit_AudioKitTests.bundle

/Users/jeffcooper/Library/Developer/Xcode/DerivedData/ak5tests-ezrjdtvhkqigsnadeukaxbuvdsry/Build/Products/Debug-iphonesimulator/AudioKit_AudioKitTests.bundle: bundle format unrecognized, invalid, or unsuitable
Command CodeSign failed with a nonzero exit code
```

aure:
But the test do run on CI and on my machine. CI is a fresh checkout every time.

aure:
Fixed.

PR #2462, opened 2021-04-16
fix tests using bad folder name 'Resources'

eljeff:
fix for https://github.com/AudioKit/AudioKit/issues/2460
You can't use a folder called 'Resources'


aure:
Do you have a reference as to why Resources is a bad folder name? Never had a problem with that folder name on my local machine or CI.

eljeff:
I don't have any references beyond googling the stuff I mention in the issue, specifically 'bundle format unrecognized, invalid, or unsuitable'.

Simple fix though

aure:
It had nothing to do with the folder name so I am going to close this "fix"

eljeff:
It's a fine alternative to remove reference to the exact folder that was causing the issue, so I wouldn't say it has 'nothing to do with it'.

Did you just leave the files in there anyway? Are they used?

Renaming the folder fixed it for me, so I bet it will fix it for someone else when they encounter this same problem.

aure:
They are resources we could use and they're in the repo already so I don't want to remove them for now. Renaming the folder was just coincidentally a fix. Whatever that folder is named, if it is added to the resources in Package.swift as a .copy reference, will not work. The name was not the issue.

eljeff:
`Whatever that folder is named, if it is added to the resources in Package.swift as a .copy reference, will not work. The name was not the issue.`
I added it via copy reference using a different name though, and it worked fine.
I'm happy w/ the fix you've done, I just want to include as much info as possible.

eljeff:
If it just works w/out the copy reference in the package, and you can still access those files, that is def a cleaner fix.

aure:
I'm fine either way too.

eljeff:
Are the files still accessible w/out the copy reference?

aure:
I don't know, since they aren't used 🤷

eljeff:
Well then yea either keep this fix or inscribe in the dark arcanum of audiokit lore that you can't have a folder named Resources in your copy reference in a swift package. 🧙‍♀️

eljeff:
Muchas gracias!

PR #2464, opened 2021-04-17
Fix midi input offsets

eljeff:
MIDI hardware receiving is failing because the timestamp is being treated as an offset

This adds a 'timeStamp' property to midievents and uses that where appropriate instead of offset.

Offset is used more in AU event processing and is unrelated, so this separates that

Open to testing / fixes - I'm not entirely sure how or why this was broken.


aure:
Its a big. change to the protocol, but its a good one so glad to get this in now, thanks.

aure:
How about if we're allowing this to default nil, we do the same with port ID?

PR #2465, opened 2021-04-17
WIP: Polyphonic samplers

gertjanjansen:
Hi,

I noticed a few things in v5 that were bothering me in v4 as well, so I though this would be a good moment to create a pull request with a suggestion. I haven't fully tested this but would like your opinion, if you agree with the direction I'm taking here.

I'm trying to make AppleSampler inherit PolyphonicNode instead of Polyphonic. Because I believe that works fine and is more useful. Is there a reason it was not polyphonic before?

Here's what I did:
- I made AppleSampler inherit from PolyphonicNode
- Then I noticed that stop and play methods of AppleSampler were throwing, but those of PolyphonicNode are not. I had 2 options here: (1) make methods on PolyphonicNode throw or (2) catch errors in AppleSampler and don't throw. I chose for the latter. But let me know if you think otherwise.
- For the same reason, I added a channel argument to the `stop` method of Polyphonic. I think it's strange that `play` receives a channel but `stop` doesn't.

I will try to add a few tests for this as well.


aure:
I don't see a problem with your approach. Make sure its well tested and lets try to get it in before AK5 is more officially launched.

eljeff:
Without looking at the code I have a first question... Apple Sampler is polyphonic by default. So what exactly do you mean by making it polyphonic?

eljeff:
Most of your changes look fine and warranted tho - so good job. I just don't understand the Polyphonic thing

gertjanjansen:
Thanks! I'll add tests.

It behaves polyphonic indeed, but it inherits from a Node instead of a PolyphonicNode. In my project I ran into an issue where I had a function that receives an `instrumentNode: PolyphonicNode` argument, but I couldn't pass an AppleSampler instance to it. So that's why I changed:
`open class AppleSampler: Node {`
to:
`open class AppleSampler: PolyphonicNode {`

gertjanjansen:
I added two tests now, just for a simple sample start and stop. I'm not exactly sure how to test the inheritance of AppleSampler. Doing something like this would be a bit useless:

`
func testPolyphonicNodeInheritance() {
    let polyphonic = AppleSampler() as? PolyphonicNode // Conditional cast from 'AppleSampler' to 'PolyphonicNode' always succeeds
    XCTAssertNotNil(polyphonic)
}
`

gertjanjansen:
I now added a test that uses an inline function that needs a PolyphonicNode as its first argument. Let me know what you think of that. I'm not sure why the CI tests fail though, they succeed on my machine.

Also the Hound keeps biting me. 🐶 Which is a good thing of course.

gertjanjansen:
I added some more tests. Do you have any idea why my the `Resources/drumloop.wav` file cannot be loaded? When I run the tests on my Mac it works fine, but somehow the CI tests can't find it. It's this line that fails:

```
let sampleURL = Bundle.module.url(forResource: "Resources/drumloop", withExtension: "wav")!
```

aure:
Recently this folder was renamed "TestResources", try that.

Issue #2470, opened 2021-04-20
V5 missing ability to play audio file from and to a specific time.


leamars:
**Is your feature request related to a problem? Please describe.**
Comparing V4 and V5, I've noticed some things missing from the previous AKAudioPlayer into the new AudioPlayer. In particular, the method to start and end playback of a file at a specific time.

This method is still mentioned in the docs:
https://audiokit.io/docs/Classes/AKAudioPlayer.html#/c:@M@AudioKit@objc(cs)AKAudioPlayer(im)playFrom:to:
![image](https://user-images.githubusercontent.com/2446390/115391551-81a44880-a1d7-11eb-8c95-f4dbd6602a86.png)

**Describe the solution you'd like**
I'd like to know whether it's going to be re-implemented, is it gone forever, and what the workaround is at the moment - if there is one?

leamars:
Just read through all the docs to see that you mentioned AudioPlayer's functionality is limited in V5, so I'll just switch to V4 (:

aure:
Actually this particular function does seem like something we could implement in v5. Would really like to have you using v5 if possible.

ryanfrancesconi:
I'll try to make some time asap to add this to the v5 player.

PR #2475, opened 2021-04-28
Updating AudioPlayer to support edited playback

ryanfrancesconi:
I think AudioPlayer is ok, but this does change the API a little bit as it is no longer either file based or buffer based. In order to have edited buffers you must keep a valid file object. That is the biggest change. So.. things like isBuffered changed.

• Also added isReversed back in cause it was easy. 
• New tests.
• and some misc loose ends like improving the comments on the recording class

Issue reference: https://github.com/AudioKit/AudioKit/issues/2470


aure:
Looks good!

PR #2486, opened 2021-05-05
Resource cleanup

wtholliday:
This adds the Stk "rawwaves" to the bundle resources, instead of storing the waves in headers and writing out tmp files.


wtholliday:
Judging from the errors, it looks like the CI is trying to build `resource_bundle_accessor.m` (auto-generated by SPM) as a C file (not ObjC). This seems like a bug in SPM. Works fine in Xcode.

wtholliday:
Well seems like the SPM support for resources in ObjC pretty half-assed. I'll work on this later.

megastep:
I think the main reason we tried to avoid actual files as resources is that it becomes a problem for static frameworks. Not really a problem for dynamic ones since you have to keep the bundle around, but the point of static is that all of it ends up in your compiled binary so not having to worry about figuring out the paths to the file in the bundle is easier.

Also as you can see SPM is still pretty half-assed in some ways.

megastep:
If we could feed the data from memory directly instead of relying on a temp file that would be ideal though.

wtholliday:
I filed a bug with Apple:

image

wtholliday:
> If we could feed the data from memory directly instead of relying on a temp file that would be ideal though.

We should use resources as they are intended. Feeding the data to STK from memory would require modifying STK in various places.

wtholliday:
With merging in the change to CI to use xcodebuild, this is now passing.

aure (approved):
This looks good to me now that it passes CI>

PR #2493, opened 2021-05-09
Make Node a protocol

wtholliday:
This makes `Node` a protocol and cleans up various things. Removes Toggleable (we've already got `shouldBypassEffect` on every AU).

We could also remove `NodeBase` by regenerating files.


aure (approved):
What can I say? Looks great.

PR #2497, opened 2021-05-10
Merge develop to prep for AudioKit 5.0.2

aure:
Lots of internal code cleanup:
-  Fixed MIDI NoteOff to Callback instrument
- Fixed Memory Leaks
- Fixed Thread issues
- Converted Node to a Protocol.
- Heavily simplified node initialization 
- More code is now autogenerated from scripts


aure:
Probably going to do a few other things and do a 5.1.0 release instead

PR #2489, opened 2021-05-07
Fix sample rate for buffered audio player

ChaseStas:
When I use buffered audio player, `playerNode` and audioFile stay with different sample rate and audio file sounds not as It should.
When I try to play, I receive this warning in debug area. My file is 48000, playerNode is 44100 Hz.
```
2021-05-07 11:46:23.486769+0300 AudioKitBugDemo[7339:95802] [general] AudioPlayer+Scheduling.swift:scheduleBuffer(at:completionCallbackType:):68:Format of the buffer doesn't match the player (AudioPlayer+Scheduling.swift:scheduleBuffer(at:completionCallbackType:):68)
2021-05-07 11:46:23.487850+0300 AudioKitBugDemo[7339:95802] [general] AudioPlayer+Scheduling.swift:scheduleBuffer(at:completionCallbackType:):69:Player <AVAudioFormat 0x600000426120:  2 ch,  44100 Hz, Float32, non-inter> Buffer <AVAudioFormat 0x600000425a40:  2 ch,  48000 Hz, Float32, non-inter> (AudioPlayer+Scheduling.swift:scheduleBuffer(at:completionCallbackType:):69)
```

It fix by adding `file?.processingFormat` in `makeAvConnections` function, when connecting playerNode to mixerNode.

I created an example project with buffered and normal player for you to see what I'm talking about.
[example repository](https://github.com/ChaseStas/AudioKitPlayerBug)

PR #2505, opened 2021-05-14
Buffer and mixed sample rate player tests

ryanfrancesconi:
For issue in PR: https://github.com/AudioKit/AudioKit/pull/2489

This completes what he started


aure:
Do you think 12345.wav could be converted to 48K by FileConverter on the fly instead of including it as a resource asset to AudioKit?

ryanfrancesconi:
it certainly could. i can do that

aure:
You might have to make a new branch off of develop in order to keep the 48K Version of the file from entering the git history.

ryanfrancesconi:
how's that?

ryanfrancesconi:
> You might have to make a new branch off of develop in order to keep the 48K Version of the file from entering the git history.

is that really so important? i deleted the file

PR #2506, opened 2021-05-15
Draft: Refactoring MIDISampler, AppleSampler and MIDIInstrument

gertjanjansen:
Hi,

The 5.1 AudioKit release broke quite a few components of my app. It was specifically this change that was an issue for me: https://github.com/AudioKit/AudioKit/commit/b38069b7965ccc9242c6f78d0cbf9a10ae106bf1#diff-fb7560882c334efaa2772be40163bacf6bb652b87b83151ee492a2fbc033fc03  

What I'm missing now is a uniform way to identify and interact with Nodes that respond to MIDI playback events (play a MIDI note, stop a MIDI note). I understand that MIDINode wasn't used by AudioKit internally, but that doesn't mean that someone using AudioKit wasn't using it. ;)

One of the things that I find a bit problematic when using AudioKit is that different classes use different method signatures for starting and stopping MIDI notes. There is no clearly defined protocol for that. Also, I believe that having two classes for MIDISampler and AppleSampler is not necessary. And I didn't understand why they did not inherit from MIDIInstrument. Why can't a sampler be a midi instrument?

So I tried to fix all of this! 

I haven't tested all of these changes yet! But I wanted to share it with you to make sure I'm not doing some work that somebody else is already doing.

There are a few things I still need to do:
- Add tests, of course! I will work on that.
- Have another look at `handleMIDI`, with respect to a race condition mentioned by Tayler and these commits: 02c7b3e763b85720bb7c6131775b9b89f1ce0705 and 830cb6240340f6596582a5520c3b67c8d3c9aa26. I might have removed something important so will double check that
- I'm not sure about the name of the new `MIDIPlayable` protocol I introduced. Might be confusing because there's also a MIDIPlayer that represents something very different. Maybe I should name it something like `MIDINotePlayable`? 
- I'm not always sure when to use public vs open. I subclass some AudioKit classes in my own app, which doesn't always work when the parent (AudioKit) class is public. I notices some classes in AudioKit are public while others are open.

### DIFF
--- a/Sources/AudioKit/MIDI/MIDIConnectable.swift
+++ b/Sources/AudioKit/MIDI/MIDIConnectable.swift
@@ -1,0 +1,27 @@
+// Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
+import AVFoundation
+
+#if !os(tvOS)
+
+/// Protocol for a node that can be connected to a MIDI client.
+protocol MIDIConnectable {
+    /// MIDI Input
+    var midiIn: MIDIEndpointRef { get set }
+    
+    /// Enable MIDI input from a given MIDI client
+    ///
+    /// - Parameters:
+    ///   - midiClient: A reference to the midi client
+    ///   - name: Name to connect with
+    ///
+    func enableMIDI(_ midiClient: MIDIClientRef, name: String?)
+    
+    /// Discard all virtual ports
+    func destroyEndpoint()
+    
+    func showVirtualMIDIPort()
+    
+    func hideVirtualMIDIPort()
+}
+
+#endif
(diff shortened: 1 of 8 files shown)


aure:
Definitely write tests, especially as we fix things and refactor them, its important that we have tests that cover the use cases we want to maintain.

aure:
In regards to open vs. public, I think the distinction is as follows. If we make something open, and we change things in AudioKit, people inheriting from that class may be adversely affected. So, we try to just be public when possible. I've come to believe that its better if we as a framework provide protocols rather than classes inheritance. Its definitely an on-going learning experience.

gertjanjansen:
> Definitely write tests, especially as we fix things and refactor them, its important that we have tests that cover the use cases we want to maintain.

Yes, I will. I plan on writing test for the develop branch, so without these changes, and then make sure the tests still run with these changes.

aure (changes requested):
The develop branch has been moving pretty fast. I think your changes are in general good, but they have gotten pretty stale. I would suggest you might close this particular PR for now and then you can make a new one once you've updated. Sound okay?

gertjanjansen:
Okay, I'll close it for now. I couldn't spend much time on it the last 2 weeks.

I believe I read somewhere that you have a Slack channel, is that correct? Could I join that? It would be nice if I could discuss changes like this up front, so I'm not doing things that you're doing as well.

aure:
What's your email? I can invite you to the Slack. I will delete the email after I use it if you post it here, or you can email me at aure at AudioKit dot io.

PR #2509, opened 2021-05-20
Use CryptoKit to calculate md5s.

wtholliday:
Removes reliance on C code in Soundpipe to calculate md5s.

This PR changes platform requirements. I'd like to be sure that's ok.


aure:
Great!

Issue #2495, opened 2021-05-09
Virtual MIDI ports, inputs and outputs are swapped


ShowtimeSoft:
When I list my virtual MIDI ports, input ports and output ports are swapped.
Virtual input ports are listed as output ports and vice versa.

**To Reproduce**

1. I create my ports:
let midi = MIDI()

midi.createVirtualInputPorts(numberOfPort: 1, [Int32(1000000)], names: ["MIDI Test Input Port"])
midi.createVirtualOutputPorts(numberOfPort: 1, [Int32(2000000)], names: ["MIDI Test Output Port"])

2. I list them, using:
midi.inputNames
and
midi.destinationNames

3. I see, that 
midi.inputNames includes "MIDI Test Output Port"
midi.destinationNames includes "MIDI Test Input Port"

**Expected behavior**
I would expect that my input port is listed in midi.inputNames, and output in outputNames.

**Screenshots**
![Artboard](https://user-images.githubusercontent.com/77680597/117583475-9d7c7980-b107-11eb-879b-b0c7d1b0baed.png)

**Details (please complete the following information):**
 - Type: Mac
 - OS: Big Sur 11.2.3
 - AudioKit Version 5

**Additional context**
StackOverflow user "Eggsalat" suggested:
In MIDI+VirtualPorts.swift, createVirtualInputPorts calls MIDIDestinationCreateWithBlock, and createVirtualOutputPorts calls MIDISourceCreate. Maybe that's where the issue is?

aure:
Bad commit message, but this should be addressed here c7438568b468eeefcda855d8ddce1e3b9ac4f239

Let me know if that works as expected.

ShowtimeSoft:
I do like 'count' rather than 'numberOfPort'.
I also like that we got rid of an unwanted whitespace in front of port names.

However, on core functionality, I now have two issues that relates to ports being swapped.

1) In the application that I send FROM (Logic Pro), the destination port now shows up as "MIDI Test Output Port".
I would expect the INPUT port to show up.

2) I can create only one port, either input or output. Only the last port created show up.

If I:
        midi.createVirtualInputPorts(count: 1, uniqueIDs: [Int32(1200000)], names: ["MIDI Test Input Port"])
        midi.createVirtualOutputPorts(count: 1, uniqueIDs: [Int32(1500000)], names: ["MIDI Test Output Port"])

Only the Output port is available

If I:

[...]

PR #2511, opened 2021-05-21
Addressing issue #2495

emurray2:
I was able to fix some of the MIDI virtual port swaps and deletions mentioned in issue #2495. I also made an example using virtual MIDI ports. Let me know if you would like me to make it a Cookbook example.


emurray2:
It seems like the workflow runner is having trouble with the PlaygroundOscillator.swift. I'm not sure how to fix it.

PR #2512, opened 2021-05-22
MIDI Fix: Adding Multiple Virtual Ports

emurray2:
Possibly addressed one of the MIDI improvements listed in MIDI+VirtualPorts.swift -- having the ability to create more than one virtual port. This works with both the virtual inputs and virtual outputs.


aure:
Thanks!

PR #2513, opened 2021-05-22
AudioKitSequencing

wtholliday:
This doesn't necessarily have to be exposed as a separate library, but I think it would be nice to at least have a separate target to show the separation of concerns in AK.

We could do this similarly for automation and the AU (i.e. AudioKitAutomation and AudioKitAU).

All three (Sequencing/Automation/AU) could be exposed as a single library (AudioKitExtended?) which would be used outside of pure-Swift playgrounds.


aure:
We discussed this and decided to go a different way.

PR #2514, opened 2021-05-23
 Resolving Split Connections (implemented by fquarters)AudioKit#2370 

emurray2:
A way to address the `required condition is false: numSplitterConnections == 0 || numSplitterConnections > 1` thrown by `AVAudioEngineGraph` when removing an input node from a `Mixer()` which is also connected to another mixer (split connection).


wtholliday (changes requested):
- Needs tests
- I suspect we can do this with a cleaner API

emurray2:
That didn't go well. It's a good thing there are tests. I'll keep looking at this and see if I can get any of them to pass.

wtholliday:
I always thought this sort of behavior of disconnecting other things to be a bit odd... why not have it always reconnect the split connections? (i.e. expectSplitConnections is always true)

Sources/AudioKit/Nodes/Mixing/Mixer.swift
```diff
@@ -90,6 +90,18 @@ public class Mixer: Node, NamedNode {
         avAudioNode.disconnect(input: node.avAudioNode)
     }
 
+    /// Remove input from a mixer who's input is connected to another mixer (a split connection)
+    /// - Parameter node: Node to remove
+    /// - Parameter expectSplitConnections: Does the input to remove also have another connection (is it splitting)?
+    public func removeInput(_ node: Node, expectSplitConnections: Bool = false) {
```
wtholliday, line 96: Do we need `expectSplitConnections`?

emurray2: That's what I was thinking, although whatever I did caused more of the node tests to fail. I'll keep looking at it some more though.

emurray2: Now I understand what you were saying. I tried something different before, but what I did just now seemed to work. Thanks for the advice, and let me know if there's anything else that needs to be fixed :) !

wtholliday (approved):

wtholliday:
We did some cleanup. Looks good! If it doesn't work for some reason, let's write another test :)

Thanks to @emurray2 and @fquarters for doing the hard work!

emurray2:
> We did some cleanup. Looks good! If it doesn't work for some reason, let's write another test :)
> 
> Thanks to @emurray2 and @fquarters for doing the hard work!

Awesome! You're welcome!

fquarters:
> We did some cleanup. Looks good! If it doesn't work for some reason, let's write another test :)
> 
> Thanks to @emurray2 and @fquarters for doing the hard work!

You're welcome! Glad to be of help :)

PR #2516, opened 2021-05-25
Organizing MIDI API

emurray2:
I edited some of the functions such as `createVirtualPorts()` so the parameters are more clear. Also, everything can now be automatically generated when there are already existing ports. This way, the caller only needs to call `destroyAllVirtualPorts()` when they want to. This may help eliminate confusion when running `createVirtualPorts()`. That way, people won't loose track of a port they expect to still be there after they run `createVirtualInputPort()` for example.

Let me know if this needs any adjustments or if it seems like a good idea.


aure:
This looks good to me so I'll merge at this stage, but keep 'em coming!

PR #2526, opened 2021-06-08
Add Docs on Enabling Modules

b3ll:
This adds a line to the migration guide explaining how to enable cxx modules to ensure that `@import AudioKit` will work within Objective-C or C++ targets.


b3ll:
hmm there's a better approach to fix this actually, I believe we can fix it via some directory changes in the targets

PR #2531, opened 2021-06-14
Update to migration guide

btfranklin:
Added a little additional info about types that stayed in the same package, but changed modules.


btfranklin:
I guess this should have been against 'develop'. I'll close this and redo it.

PR #2532, opened 2021-06-14
Update to migration guide

btfranklin:
Added a little additional info about types that changed modules while staying in the same package, which can be a little bit non-obvious.


aure:
Thanks!

PR #2545, opened 2021-07-09
FormatConverter update

ryanfrancesconi:
handle bitRate conversion for stereo vs mono given that the mono file is effectively 1/2 the bitrate of the stereo.
I've chosen to handle this by keeping one bit rate range and dividing by 2 if mono input.
This conversion previously failed if the bit rate was above the bounds allowed by m4a mono.

So, if bitrate is 320 (highest supported) : 320 bitrate for stereo and 160 for mono.


ryanfrancesconi:
kinda hard for me to keep this in sync due to difference in my own API vs AudioKit's

aure:
Thanks Ryan, how can we make the merge process easier? I don't make too many changes to your files so it should be okay from now on.

ryanfrancesconi:
i just need to pay closer attention if any commits were made!

Issue #2541, opened 2021-07-01
AudioPlayer: crash when loading audiofiles with different formats while playing


ShowtimeSoft:
**Describe the bug**
AudioPlayer
I get a crash when I try to load an audiofile while I'm playing back another.
I get the crash only when the two files are of different format, e.g. mp3 and m4a.

**To Reproduce**
Load an audiofile.mp3
audioPlayer.start()
Load an audiofile.m4a

-> crash

isBuffered can be true or false, no difference.

**Expected behavior**
I would expect to be able to load any kind of supported audiofile while playing back.

**Details (please complete the following information):**
 - Type: MacOS 
 - OS: Big Sur
 - AudioKit Version 5.2 main

PR #2558, opened 2021-07-25
Check if playerNode is already connected to mixerNode

emurray2:
Addresses #2541. Also added tests for this.

### DIFF
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
@@ -269,6 +269,20 @@ public class AudioPlayer: Node {
 }
 
 extension AudioPlayer: HasInternalConnections {
+    /// Check if the playerNode is already connected to the mixerNode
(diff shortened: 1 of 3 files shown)


Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -269,6 +269,19 @@ public class AudioPlayer: Node {
 }
 
 extension AudioPlayer: HasInternalConnections {
+    /// Check if the playerNode is already connected to the mixerNode
+    func playerNodeConnectedToMixerNode(engine: AVAudioEngine) -> Bool {
```
aure, line 273: Can we prefix that function with "is" to be "isPlayerNodeConnectedToMixerNode" to really drive home that its a boolean? Also (and I didn't check yet), does the engine need to be sent to this? Probably can get the engine from the node and this can be a computed variable instead of a function.

aure (changes requested):

aure:
Cool.

PR #2560, opened 2021-07-27
AudioPlayer Improvements

emurray2:
Addresses AudioKit#2539

Problem: the `_editEndTime` setter clamps the audio file to the lowest file length loaded, even when a new file is loaded. This prevents the buffer from updating if the file is loaded in buffer mode. Thus, the shortest audio file loaded will always be played over any other loaded file.

Improvements:
- Force the buffer to update every time when a new file is set (achieved through the player's `file` property setter)
- Add `isEditTimeEnabled` property to toggle the edit region of the file on/off
- Add `preserveEditTime` preference to `load()` function - default behavior is to load a new file without the previous file's edit time
- Add tests for toggling the `isEditTimeEnabled` property


aure:
Sweet! Thanks!

Issue #2559, opened 2021-07-26
Realtime linear ramping doesn't work (tapered does)


ryanfrancesconi:
I am working with the automation, and had to make some fixes on my end for scheduled events. I don't use realtime ramping myself, but was concerned I broke it, so I made this test in the AK5 tests to double check. This doesn't ramp, though I would assume it should. Is this supposed to work like this?

If I change it to a curve, it does work. I'm assuming it has something to do with the limited points in the automation when linear. I found that linear in general doesn't work right - scheduled as well. That said, I don't use linear, but I'm wondering what is wrong there.

```swift
extension FaderTests {
    func testRealtimeRamp() {
        let engine = AudioEngine()
        let url = Bundle.module.url(forResource: "12345", withExtension: "wav", subdirectory: "TestResources")!
        let player = AudioPlayer(url: url)!
        let fader = Fader(player)

        engine.output = fader

        try? engine.start()

        fader.start()
        player.play()
        
        // Doesn't work:
        fader.$leftGain.ramp(from: 0, to: 1, duration: Float(player.duration))
        fader.$rightGain.ramp(from: 0, to: 1, duration: Float(player.duration))

        wait(for: player.duration + 0.5)

        player.play()
        // Does work:
        fader.$leftGain.taperedRamp(from: 0, to: 1, duration: Float(player.duration))
        fader.$rightGain.taperedRamp(from: 0, to: 1, duration: Float(player.duration))
        
        wait(for: player.duration + 0.5)
    }

    // for waiting in the background for realtime testing
    func wait(for interval: TimeInterval) {
        let delayExpectation = XCTestExpectation(description: "delayExpectation")
        DispatchQueue.main.asyncAfter(deadline: .now() + interval) {
            delayExpectation.fulfill()
        }
        wait(for: [delayExpectation], timeout: interval + 1)
    }
}
```

Added to NodeParameter+Automation:

```swift
/// Audio Taper Ramp from a source value (which is ramped to over 20ms) to a target value
    ///
    /// - Parameters:
    ///   - start: initial value
    ///   - target: destination value
    ///   - duration: duration to ramp to the target value in seconds
    ///   - rampTaper: Taper, default is 3 for fade in, 1/3 for fade out
    ///   - rampSkew: Skew, default is 1/3 for fade in, and 3 for fade out
    ///   - resolution: Segment duration, default 20ms
    public func taperedRamp(from start: AUValue,
                            to target: AUValue,
                            duration: Float,
[...]

PR #2563, opened 2021-07-29
Automation update

ryanfrancesconi:
In order to get the way that I use automation to work correctly, I needed to make these changes. For me, this works correctly now both offline and in realtime. I'm reluctant to mess with what you made, but you can always just ignore this if you want.

This also fixes the issues i saw with the realtime ramp not working correctly for linear which i previously submitted an issue and closed by adding an additional point to create a `AUEventSampleTimeImmediate`. It is possible that could be addressed in `ParameterAutomationGetRenderObserver` instead of how I did it?

review requested: @wtholliday


wtholliday:
- Spurious formatting changes makes the PR hard to understand
- I'm not clear on what the actual problem is (not denying there is one, I just don't know)
- It would be great to have a smaller, more focused PR.
- "previously submitted an issue" which one?
- the PR creates coupling between `NodeParameter` and `AutomationCurve`

aure:
I brought in the formatting changes so if you could redo the PR they won't appear as changes.

ryanfrancesconi:
> * Spurious formatting changes makes the PR hard to understand
> * I'm not clear on what the actual problem is (not denying there is one, I just don't know)
> * It would be great to have a smaller, more focused PR.
> * "previously submitted an issue" which one?
> * the PR creates coupling between `NodeParameter` and `AutomationCurve`

The issue submitted is this one: https://github.com/AudioKit/AudioKit/issues/2559
I was not seeing the real time linear ramping working, but then i just decided to fix it myself in this PR. If you're interested to check that yourself, you might try those original tests in that issue with the current codebase.

The actual issue for me was that basically automation didn't work in the contexts I use. Mainly - scheduled automation, matched automation for both offline and realtime, starting mid way through an automation sequence, and so on. To recreate all the problems at this point in the original code would take me a moment to do, but if i can find time to do that then i can describe exactly. It's not so easy for me to create these types of situations in audio kit tests as they involve a UI in order to really see / hear what is going on in terms of scheduled ramps. I tried to created tests that show that these things work now in my PR anyway. It's all a bit abstract without a UI for me when dealing with complicated automation data.

It's fine to remove the taperedRamp addition from NodeParameter if you don't want it. I don't use it myself, but it's a good example on how to do that for someone wondering how the curves should work. I'm sure there is a better place to show that, or maybe it's already somewhere.

ryanfrancesconi:
Part of the reason the PR is large and less focused, is that my priority was getting it working for me (outside of audiokit) - which i did, then i've just moved it to here. I'm not using AudioKit, but have adapted this automation code to my project.

No problem if you just want to close this, was just trying to help.

ryanfrancesconi:
A visual example of how I'm testing this:

Starting mid way through a complicated curve sequence:
image

Starting mid way through a tapered curve:
image

Scheduled fade in and out, tapered:
image

and so on...

David4noy:
Ok, it's a little too much for me to understand all what you wrote above, but I have an issue too who might be connected.

I'm trying to do an auto band filter with this code and I get an error.

        
        let osc = Oscillator()
        let filter = EqualizerFilter(osc)
        
        var frequancy:Double = 500
        var startFrequency: AUValue = 500
        var endFrequency: AUValue = 3000
        var frequancyRampDuration: Float = 3.0
        
        filter.$centerFrequency.ramp(from: startFrequency, to: endFrequency, duration: frequancyRampDuration)
        
       print(filter.centerFrequency)
        

    If not - Should I start a new issue?

ryanfrancesconi:
> libc++abi.dylib: terminating with uncaught exception of type NSException
> *** Terminating app due to uncaught exception 'com.apple.coreaudio.avfaudio', reason: 'required condition is false: _engine != nil'

if that is all of your code, then you need an engine, but yes, best to not put that in here. this is it's own thing.

ryanfrancesconi:
> I brought in the formatting changes so if you could redo the PR they won't appear as changes.

cool, i'll redo it.

ryanfrancesconi:
Updated PR notes:

I moved the taperedRamp function to Fader -- Originally i only added that as the ramp() wasn't working and I wanted to see if taperedRamp did - which it did as there are many more points so it gets the valid start point.

Note: I believe the original issue for linear ramping is that the actual  ramping code wasn't registering the initial start point correctly (it might of been in the past but not set) so there wasn't anything to ramp to. Since it went from [0 -> 1] and if it doesn't get 0, then there is no ramp as it just automated 1.

David4noy:
> if that is all of your code, then you need an engine, but yes, best to not put that in here. this is it's own thing.

It's not all the code and I do have engine.
If I deleting this code all works good..
Thanks

ryanfrancesconi:
Is this getting too stale? Maybe just close and I'll submit an issue describing the problems I see?

aure:
> Is this getting too stale? Maybe just close and I'll submit an issue describing the problems I see?

Okay

PR #2566, opened 2021-08-02
Remove inaccurate gain comment

eljeff:
Fader is not limited to gain of 4


aure:
By removing the comment entirely you reduce our code coverage and it shows up as undocumented, so please add it back, just removing the incorrect limit of 4.

PR #2583, opened 2021-08-28
Fixed imaginary number for scaling by split complex in FFTTap

Matt54:
Addresses comments made in: https://github.com/AudioKit/AudioKit/pull/2581 by @aburgel

Set the imaginary element of complex split to zero for scaling

Renamed inputCount to binCount

Set initial array length of fftData to the bin count


aburgel:
looks good to me! :shipit:

Issue #2584, opened 2021-08-30
AVAudioPCMBuffer.append not working as expected


jasonayre:
AVAudioPCMBuffer.append does not seem to be working correctly, or at least not how I would expect it to, and I cannot find it being used anywhere in examples or other code paths

**To Reproduce**
Take the following code for example, in which I am just trying to use the extract method, to extract a copy of a slice of the original buffer, and append it to the original buffer, which I would expect would just put it on the end of the original buffer.

```
let fileURL = "somepath.wav"
let file = try AVAudioFile(forReading: fileUrlFor(path: fileURL))
let sampleRate = file.sampleRateForFile
let audioFormat = AVAudioFormat(standardFormatWithSampleRate: sampleRate, channels: 2)
let loopBuffer = AVAudioPCMBuffer(pcmFormat: audioFormat!, frameCapacity: AVAudioFrameCount(file.length))
try file.read(into: loopBuffer!)
let startFrame = AVAudioFramePosition(loopBuffer!.frameLength)
loopBuffer!.append(loopBuffer.extract(from: 0.0, to: 2.0)!, startingFrame: startFrame, frameCount: AVAudioFrameCount(file.length))
```

Always results in the "Insufficient audio in buffer" precondition being hit. - Using the append method without any parameters (just passing in the extracted buffer) appears to do the same thing. 

**Expected behavior**
I am expecting the original buffer to have the extracted buffer appended to it, and the length of the original buffer to be the sum of the original buffer + the extracted buffer

**Details (please complete the following information):**
 -Type: Mac 
 - OS: 10.15.7
 - Audio Kit V5, rev b1099859f6654c98f0f28e9fb1ed388ae96cb7d6

aure:
Could you PR a failing test? That might be the best way to give us a working (nonworking) example.

jasonayre:
Most definitely, https://github.com/AudioKit/AudioKit/pull/2585

aure:
I'm inclined to accept the now passing test and close this issue, but I do not know how or why this is not failing anymore. Possibly it was fixed in an OS update?

PR #2585, opened 2021-08-31
append failing test reproduction

jasonayre:
*Failing test for #2584*

Simplified example from #2584, converting a file to a AVAudioPCMBuffer, and trying to append to the file, throws the same error I was seeing in my app. Added a failing test per request.


aure:
So, this test is appearing to pass, I'm confused.

PR #2586, opened 2021-08-31
fix decimator

eljeff:
A few of the ak effects are based on singling out a single mode of apples multi-mode-distortion.
At some point, the setup for these nodes got trashed, so the decimator for example was always distorting via the other modes.

This creates private parameters and then uses those to turn the other effects off.

also fixed a weird line break somewhere else.


eljeff:
updated the decimator test but no idea why the other test is failing.
testNodeConnection()  in nodetests.swift

aure:
Thanks, this probably affects other stuff too.

eljeff:
Yea - any of the effects based off a subset of the distortion au will need something similar.
If there is a more direct way to set these params that we only use for setup, i'd love to know it.

PR #2587, opened 2021-08-31
more direct setting of hidden params in decimator

eljeff:
simpler way to set values.
aure did this.

updated test.


eljeff:
since this hasnt been merged yet, now it also uses correct values for some distortion parameters

aure:
I'll take your word for it, but just so I know, where did you get the new info for the parameter limits and defaults from?

eljeff:
> I'll take your word for it, but just so I know, where did you get the new info for the parameter limits and defaults from?

The documentation
Right click on the property name, jump to definition.

aure:
> > I'll take your word for it, but just so I know, where did you get the new info for the parameter limits and defaults from?
> 
> The documentation
> Right click on the property name, jump to definition.

ruh roh, lets talk about it on slack

PR #2590, opened 2021-09-09
support partial retrieve of waveform

wkpark:
This is a quick hack to support partial retrieve waveform.

 * `offset`: optional start offset to retrieve (`offset` = `0` is default. ~~`-1`~~ `nil` to retrieve from `currentFrame` position. (offset could be minus relative to `currentFrame`)
 * `length`: optional `UInt` - length of waveform. ~~`0`~~ `nil` means full length or remains

### DIFF
--- a/Sources/AudioKit/Audio Files/WaveformDataRequest.swift
+++ b/Sources/AudioKit/Audio Files/WaveformDataRequest.swift
@@ -42,17 +42,18 @@ public class WaveformDataRequest {
 
     /// will be returned on the queue you pass in or the global queue
     public func getDataAsync(with samplesPerPixel: Int,
+                             offset: Int = 0, length: Int = 0,
                              queue: DispatchQueue = DispatchQueue.global(qos: .userInitiated),
                              completionHandler: @escaping ((FloatChannelData?) -> Void)) {
         queue.async {
-            completionHandler(self.getData(with: samplesPerPixel))
+            completionHandler(self.getData(with: samplesPerPixel, offset: offset, length: length))
         }
     }
 
     /// Get waveform data
     /// - Parameter samplesPerPixel: Number of samples you want per point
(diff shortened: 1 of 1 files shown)


Sources/AudioKit/Audio Files/WaveformDataRequest.swift
```diff
@@ -42,17 +42,18 @@ public class WaveformDataRequest {
...
                              queue: DispatchQueue = DispatchQueue.global(qos: .userInitiated),
                              completionHandler: @escaping ((FloatChannelData?) -> Void)) {
         queue.async {
-            completionHandler(self.getData(with: samplesPerPixel))
+            completionHandler(self.getData(with: samplesPerPixel, offset: offset, length: length))
         }
     }
 
     /// Get waveform data
     /// - Parameter samplesPerPixel: Number of samples you want per point
```
aure, line 54: You have added parameters without updating the documentation comments. This is especially important since 0 and negative numbers do special things that are not obvious at all from the API. (The things that you mention in the pull request should work well for the documentation.)

aure (changes requested):

Sources/AudioKit/Audio Files/WaveformDataRequest.swift
wkpark: ok. I've fix parameters like as following for more clear behavior:
 * `offset`: `0` or not given (default) / `offset`: nil or minus - (from `currentFrame`) => optional `Int`
 * `length`: nil (default) or not given, for full length(or remains) / `length` >= 0 for partial retrieving. => optional`UInt`

aure:
Minor cleanup: https://github.com/AudioKit/AudioKit/commit/a1f39b71b75aa805317cdd22bb442986dba26a40

PR #2608, opened 2021-10-25
Make functions public

Truba:
Hi guys, I'm working on custom DryWetMix that can change inputs.
Making this functions public allows me to do the engine updates.

I know the updates should be pushed to develop, and I'll change it if you want it.
I've pushed them to main since they are just exposing already created logic.


aure:
The consensus, for now, is that making these functions public is exposing functionality that the user shouldn't need. It seems like it allows you to hack through a problem you're having, but in that case maybe it should just live in your fork of AudioKit? I'm just wary of exposing too many internal functions. Convince me otherwise or perhaps change AudioKit so that this isn't necessary?

PR #2613, opened 2021-11-05
Add time conversion functions to `AppleSequencer`

YuAo:
Add time conversion functions to `AppleSequencer` that convert between host time and beats time.

### DIFF
--- a/Sources/AudioKit/Sequencing/Apple Sequencer/AppleSequencer.swift
+++ b/Sources/AudioKit/Sequencing/Apple Sequencer/AppleSequencer.swift
@@ -839,4 +839,39 @@ open class AppleSequencer: NSObject {
     func modTime(_ time: Double) -> Double {
         return time.truncatingRemainder(dividingBy: length.beats)
     }
+
+    // MARK: - Time Conversion
+
+    public enum MusicPlayerTimeConversionError: Error {
+        case musicPlayerIsNotPlaying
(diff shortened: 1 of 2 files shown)


aure:
Could a test be written for the new functions?

YuAo:
I don't think we can write accurate end to end tests for these functions.

These functions do time conversion between host time and beats. We do not know the accurate host time the music player starts to play (If we do, we won't be needing these functions).

An "inaccurate" way to test these is to see if the result lays in a reasonable range (10ms, 100ms, 1000ms ?)

For example, test `hostTime(forBeats:)` using a 100ms threshold.

```swift
guard let newTrack = seq.newTrack() else {
    XCTFail()
    return
}
newTrack.replaceMIDINoteData(with: generateMIDINoteDataArray(beatCount: 4, noteNumber: 50))
seq.setLength(Duration(beats: 4))
seq.setTempo(90)
seq.play()
let estimatedPlayerStartTime = mach_absolute_time()
let time = try seq.hostTime(forBeats: 4)
seq.stop()
let diff = CMClockMakeHostTimeFromSystemUnits(time - estimatedPlayerStartTime).seconds - 4 * 60.0 / 90.0

XCTAssert(abs(diff) < 0.1)
```

@aure What do you think?

aure:
I think that test would be better than nothing, sure.

PR #2621, opened 2021-11-16
use builtin clamp, some refactoring

adamnemecek:
I realized that this should be targeting the develop branch too late. I'll do it next time.


aure:
A lot better, thank you! Keep em coming!

PR #2629, opened 2021-11-25
Update dryWetMix default value in Delay.swift

NickCulbertson:
Changed default for dryWetMix to a value in range.


aure (changes requested):
This can't be right. The range is still 0->1. Perhaps that should also be changed to 0->100?

sajj100:
Pls unsubscribe me from all email, as it's spamming my inbox and I can't  do it via unsubscribe option on mail due to login issues.RegardsSajjas On Nov 30, 2021 3:03 PM, Aurelius Prochazka ***@***.***> wrote:
Merged #2629 into develop.

PR #2632, opened 2021-11-30
CallbackInstrumentDSP: Replace NSTimer with GCD timer

Truba:
`CallbackInstrumentDSP` uses `NSTimer` which depends on `RunLoop` on the main thread.
If you were to try to instantiate / use `CallbackInstrument` on any other thread it would fail.

NSTimer in the current implementation also causes memory leaks, since it captures the self/this strongly, and the deconstructor that should invalidate it never gets called.

This PR aims to solve those issues by replacing the NSTimer with GCD timer and putting it on a separate queue.


aure:
Thanks!

Issue #2646, opened 2021-12-30
AudioPlayer play() & pause() not behaving as expected


kxmmeron:
**Describe the bug**
When using play() and pause() for toggle playback functionality (where play() is called if the audio is paused and pause() is called if audio is playing), calling pause() during the second toggle cycle does not pause the audio and the UI becomes out of sync.

**To Reproduce**
1. Create an instance of AudioPlayer
2. Load an audio file
3. Call AudioPlayer.play() to start audio
4. Call AudioPlayer.pause() to pause audio
5. Repeat step 3-4
6. See that audio isn't paused on second cycle and UI reports inaccurately

**Expected behavior**
I expect that calling pause() after play() would always stop the audio.  Not too sure but seems that it could be some sort of state issue.  I noticed that both `isPlaying` and `!isPaused` in `AudioPlayer+Playback.swift` Line 56 are false causing the function to return when this issue is encountered even though audio is playing.

There seems to be a workaround for this: calling play() right before pause() seems to give the behavior that I am looking for but doesn't seem natural to me.

**Details (please complete the following information):**
 - Type: macOS, iOS
 - OS: macOS Monterey 12.1, iOS 15.2
 - AudioKit 5.3

**Additional context**
It also seems that `getCurrentTime()` is still increasing after pausing audio as well.

### Commit demonstrating issue in Cookbook: https://github.com/kxmmeron/Cookbook/commit/3523f9063e0559571389c1d20199ac82c8737e3a

Use `Enable Workaround` toggle to see behavior before and after workaround (calling play() right before calling pause())

emurray2:
I believe `getCurrentTime()` was mentioned in #2620 as a feature, but it seems like enough people are expecting it to return a finite time. I'll open up another PR for that soon.

kxmmeron:
Thanks for the quick fix Evan, I appreciate it!

Issue #2576, opened 2021-08-10
AudioPlayer  doesn't stop after pause and volume decreases when pressing play


David4noy:
If I do pause and then stop or pause -> play -> stop, it doesn't stop.
Instead,  it keeping the last pause point and "stop" there (and play from there).
Also, if you do pause - > stop -> play - it won't play in the first tap - only in the second.

I use the simple methods:
    
    func playPlayback(){
        playback.play()
    }
    
    func pausePlayback(){
        playback.pause()
    }
    
    func stopPlayback(){
        playback.stop()
    }
    
I tried also with:

    func playPlayback(){
        
        if playback.isPaused {
            playback.resume()
        } else {
        playback.play()
        }
    }

But was the same

aure:
You're a good problem finder. Do you think you could write a test that produces the incorrect result that we could then fix?  It would be a good habit to get into.

David4noy:
> You're a good problem finder. Do you think you could write a test that produces the incorrect result that we could then fix? It would be a good habit to get into.

To be honest, I don't really know how to write and do test yet, I'm still a beginner.. I'll try to learn about it and do, I guess it will take me awhile but I want to help if I can.

David4noy:
Btw, I thought about something who might be helpful, to add to AudioPlayer:

enum StateMod {
    
    case looping
    case paused
    case playing
    case stopped
    case seeking
    case reversing
    case buffering
    case scheduled
    
}

This way we can interact more easily with the player state and to move between them (I think...)

We can also add some row value for some uses or debugging. for example:

enum StateMod: String {
    
    case looping = "looping"
    case paused = "paused"
    case playing = "playing"
    case stopped = "stoped"
    case seeking = "seeking"
    case reversing = "reversing"
    case buffering = "buffering"
    case scheduled = "scheduled"
    
}

Log(StateMod.RawValue)

David4noy:
Btw, I thought about something who might be helpful, to add to AudioPlayer:


enum StateMod: String {

    case looping
    case paused
    case playing
    case stopped
    case seeking
    case reversing
    case buffering
    case scheduled

}



This way we can interact more easily with the player state and to move between them (I think...)

We can also add some row value for some uses or debugging. for example:


enum StateMod: String {

    case looping = "looping"
    case paused = "paused"
    case playing = "playing"
    case stopped = "stoped"
    case seeking = "seeking"
    case reversing = "reversing"
    case buffering = "buffering"
[...]

Issue #2539, opened 2021-06-30
AudioPlayer: Can't load an audio file that is longer than the one currently loaded (buffered = true)


ShowtimeSoft:
**Describe the bug**
AudioPlayer:
If I load an audiofile B that is longer (duration wise) than the one currently loaded A, it does not seem to be loaded correctly, at least A is still being played back.

Happens only when isBuffered = true.

I have attached a compiled test app and source code.

The test app includes three audio files of various length.
You will see that you can only load files that are shorter (further down the list) than the one currently loaded.

**To Reproduce**

1) player.load(file: "a short file", buffered: true)
2) player.load(file: "a long file", buffered: true)

**Expected behavior**
I expect that the long file is now loaded and played back, but it is not. It is still the short file that is being played.

**Details (please complete the following information):**
 - Type: MacOS
 - OS: Big Sur
 - AudioKit Version 5.2 Main

As far as I can see from one of my other projects, this is not new to version AK 5.2, as I can also make it occur there.

<img width="646" alt="image" src="https://user-images.githubusercontent.com/77680597/123964333-9bba8e00-d9b3-11eb-86ec-50f702451322.png">

[AudioKitTest.zip](https://github.com/AudioKit/AudioKit/files/6741005/AudioKitTest.zip)

Source:
[AK Test Source.zip](https://github.com/AudioKit/AudioKit/files/6741067/AK.Test.Source.zip)

aure:
@ShowtimeSoft You may reopen the issue if @emurray2 's fix proves insufficient. Till then, this is closed.

PR #2651, opened 2022-01-08
Add an enum file for node states

emurray2:
This may help with state management in the future (i.e. only having one state for being .playing or .paused).

In the past, there seems to be scenarios where the player has two states at a time (#2646, #2576, #2539). For example, maybe it is set to `isPlaying == true` in `play()`, but then set to `isPaused == true` when running `seek()`. I know that's probably not the case. However, maybe some states of the player would be easier to manage in an enum. Not every state would, but the significant ones might. A node probably shouldn't be playing and paused at the same time, but it might be playing and seeking.

Also, this article was sort of an inspiration behind this idea: https://dev.to/mpocock1/state-management-how-to-tell-a-bad-boolean-from-a-good-boolean-260n . I know web development is different, but this thinking may be useful here. Please let me know if anyone else has any ideas, feedback or thoughts on this. Thanks :) !


aure:
DId you mean to include these tests in this PR?

aure:
I think the enum is a good idea, but I don't think it should be committed until it is used and tested.

Tests/AudioKitTests/Node Tests/Player Tests/AudioPlayerTests.swift
```diff
@@ -2,6 +2,66 @@ import AudioKit
...
+
+    let player = AudioPlayer()
+
+    init() {
+        mixer = Mixer()
+        mixer.addInput(player)
+        engine.output = mixer
+
+        do {
+            try Settings.setSession(category: .playAndRecord,
```
emurray2, line 19: Weird, this compiles in Xcode. Not sure why it doesn't compile here. Could possibly be related to this? https://stackoverflow.com/questions/67494326/swiftui-cannot-infer-contextual-base-in-reference-to-member

emurray2:
> DId you mean to include these tests in this PR?

I did, they seemed related to the same issue, but I think I'll open a separate one.

Sounds good, I'll work on testing them. Is there any way to get `AudioPlayer` to play sound during the tests? I can hear the oscillators, but I don't hear a sound when the player plays, so I don't know how to verify if it's working or not.

PR #2652, opened 2022-01-10
Add NodeStatus and use with AudioPlayer

emurray2:
This adds the `NodeStatus` `enum` and uses it with the `AudioPlayer` Playback Node.

### DIFF
--- a/Sources/AudioKit/Nodes/NodeStatus.swift
+++ b/Sources/AudioKit/Nodes/NodeStatus.swift
@@ -1,0 +1,18 @@
+// Copyright AudioKit. All Rights Reserved. Revision History at http://github.com/AudioKit/AudioKit/
+
+/// Keep track of a node status
+public enum NodeStatus {
+    // Node status for a playback node
+    public enum Playback {
+        // The node is stopped
+        case stopped
+        // The node is playing
+        case playing
+        // The node is paused
+        case paused
+        // The node is seeking
+        case seeking
+        // The node is scheduling
+        case scheduling
+    }
+}
--- a/Sources/AudioKit/Nodes/NodeStatus.swift
+++ b/Sources/AudioKit/Nodes/NodeStatus.swift
@@ -4,13 +4,13 @@
 public enum NodeStatus {
     /// Node status for a playback node
     public enum Playback {
-        /// The node is stopped
+        /// The player node is stopped.
         case stopped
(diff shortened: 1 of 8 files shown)


Sources/AudioKit/Nodes/NodeStatus.swift
```diff
@@ -0,0 +1,18 @@
...
+    /// Node status for a playback node
+    public enum Playback {
+        /// The player node is stopped.
+        case stopped
+        /// The player node is playing.
+        case playing
+        /// The player node is paused.
+        case paused
+        /// The player node is scheduling.
+        case scheduling
```
emurray2, line 14: This replaces the old property `isScheduled` and there is no need for `scheduleTime` anymore since the only thing dependent on it was the `isScheduled` property.

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
```diff
@@ -34,36 +23,47 @@ extension AudioPlayer {
...
+            status = .playing
+        case .playing:
+            // player is already playing
+            return
+        case .paused:
+            resume()
+        case .scheduling:
+            // player is already scheduling
+            return
+        case .completed:
```
emurray2, line 44: This replaces the logic inside the completion handler -- just moved most of it here.

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer.swift
```diff
@@ -26,14 +26,8 @@ public class AudioPlayer: Node {
...
-    /// Whether or not the playing is playing
-    public internal(set) var isPlaying: Bool = false
-
-    /// Whether or not the playing is paused
-    public internal(set) var isPaused: Bool = false
-
-    /// Will be true if there is an existing schedule event
-    public var isScheduled: Bool { scheduleTime != nil }
+    /// Status of the player node (playing, paused, stopped, or scheduling
+    public internal(set) var status = NodeStatus.Playback.stopped
```
emurray2, line 30: Initialization for the node status. This is what I was thinking for each type of node. Then more enums could be added to the whole status enum as they are used with other nodes.

aure:
Thanks!

PR #2667, opened 2022-02-05
NodeRecorder: write to specified directory

Matt54:
Keeps default functionality unchanged, but when fileDirectoryPath is provided then the NodeRecorder will write its files to this location. 

In other words, tell NodeRecorder "I want the files to get written to this folder - name them whatever" and then the dev can just keep track of the file names after each recording.

Does this sound like a good idea? I'm open to suggestions.

From what I could tell, if you wanted to do something like record audio files to the documents directory then you needed to either:
1) continuously hand off a new AVAudioFile to a new instance of a NodeRecorder
or
2) move/copy each file located in the temporary directory after it was recorded

1 seemed like a hack and 2 seemed like an added overhead with the file move/copy.

### DIFF
--- a/Sources/AudioKit/Taps/NodeRecorder.swift
+++ b/Sources/AudioKit/Taps/NodeRecorder.swift
@@ -49,7 +49,7 @@ open class NodeRecorder: NSObject {
             return nil
         }
     }
-    
+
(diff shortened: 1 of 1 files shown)


Matt54:
Can I rerun the CI test?

PR #2668, opened 2022-02-05
Midi: remove velocity from sendNoteOffMessage

Matt54:
I think it is confusing to have an inconsequential velocity value included for a note off message, but I could see leaving the argument and defaulting it to zero for backwards compatibility...

Unless there's some situation where note off has a velocity that I'm not aware of??


aure:
I've never seen midi note off velocity used anywhere.

PR #2677, opened 2022-02-17
Fix typo in variable name of internal var incomingUMPSysExMessage.

ronyeh:
Added a missing letter n to the variable name `incomingUMPSysExMessage`.

Impact to developers should be minimal since this is marked as an internal var.


aure:
Thank you!

PR #2679, opened 2022-02-18
Fix typos

rex4539:


aure (changes requested):
Mostly great but Apple uses the "Busses spelling" so we have to as well:

https://developer.apple.com/documentation/audiotoolbox/auaudiounit

rex4539:
Should be ok now.

Issue #2676, opened 2022-02-17
Bug Fix and Refactor Proposal: Midi note on/off + Time Stamps (Ableton Midi into tracks bug)


Matt54:
**Screenshot / Describe the bug**
![image (2)](https://user-images.githubusercontent.com/15333030/154376943-843ca885-6d7f-4cdd-9a07-cdd3110d5662.png)

I'll start with the solution and work backwards - if you send midi note on/off with an appropriate timestamp included, then Ableton will always forward the messages to the tracks.

(What I say below applies to both note on and note off - I will just refer to note on)
We have two main methods in AudioKit for sending note on:
- MIDI.sendNoteOnMessage
- MIDI.sendNoteOnMessageWithTime

The sendNoteOnMessage method calls sendMessage
The sendNoteOnMessageWithTime calls sendMessageWithTime

I cannot tell the difference between sendMessage and sendMessageWithTime at a glance (they appear a little different in implementation), so I could be wrong with what I'm about to suggest.

My suggested changes:
1) Refactor sendNoteOnMessage and sendNoteOnMessageWithTime into one method and refactor sendMessage and sendMessageWithTime into one method.

Reason:
It's safe to assume users of AK will always use sendNoteOnMessage as their first choice (I mean why include the time if you aren't keeping time of anything?? - just want to send a midi note NOW. Well apparently, that's not good enough...)

2) Document somewhere that a timeStamp is needed to avoid possible bugs with the way a DAW handles a midi event. 

AND/OR

Determine how to create a timeStamp when we are "self generating a midi note" as opposed to just getting it from a keyboard and forwarding it to the DAW (this is my current use case). If you are "self generating a midi note", where you don't actually have a timeStamp, then you will find that just sending 0 or MIDITimeStamp() will not work. If no timeStamp is included in the refactored sendNoteOn method, then we will add this "self generated timeStamp" as the default.

**To Reproduce**
This is a difficult to reproduce issue. It does not consistently occur, but as far as I can tell it tends to occur after Ableton has been open on your computer for a long time or your have not restarted your computer for a long time.

**Expected behavior**
Midi notes should always be forwarded to Ableton tracks (assuming appropriate preferences and I/O setup)

orchetect:
As per Apple docs, timeTag must not be 0 when a packet is sent with `MIDIReceived()`. It must be a proper timeTag, which would be passing in `mach_absolute_time()` for the timetag. Some software and hardware will not respond to messages if you pass in 0.

[...]

PR #2680, opened 2022-02-18
Refactored two midi send methods into one with timeStamp defaulted to mach_absolute_time()

Matt54:
Refactored two midi send methods into one with timeStamp defaulted to mach_absolute_time()

This is a solution to: #2676

### DIFF
--- a/Sources/AudioKit/MIDI/MIDI+Sending.swift
+++ b/Sources/AudioKit/MIDI/MIDI+Sending.swift
@@ -287,11 +287,11 @@ extension MIDI {
     ///   - noteNumber: MIDI Note Number
     ///   - velocity: MIDI Velocity
     ///   - channel: MIDI Channel (default: 0)
-    ///   - time: MIDI Timestamp (default: mach_absolute_time())
+    ///   - time: MIDI Timestamp (default: mach_absolute_time(), note: time should never be 0)
     public func sendNoteOnMessage(noteNumber: MIDINoteNumber,
                                   velocity: MIDIVelocity,
                                   channel: MIDIChannel = 0,
-                                  time: MIDITimeStamp = MIDITimeStamp(mach_absolute_time()),
+                                  time: MIDITimeStamp = mach_absolute_time(),
                                   endpointsUIDs: [MIDIUniqueID]? = nil,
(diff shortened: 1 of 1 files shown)


Sources/AudioKit/MIDI/MIDI+Sending.swift
```diff
@@ -352,27 +287,31 @@ extension MIDI {
...
         let message: [MIDIByte] = [noteCommand, noteNumber, velocity]
-        self.sendMessage(message, endpointsUIDs: endpointsUIDs, virtualOutputPorts: virtualOutputPorts)
+        self.sendMessage(message, time: time, endpointsUIDs: endpointsUIDs, virtualOutputPorts: virtualOutputPorts)
     }
 
     /// Send a Note Off Message
     /// - Parameters:
     ///   - noteNumber: MIDI Note Number
     ///   - channel: MIDI Channel (default: 0)
+    ///   - time: MIDI Timestamp (default: mach_absolute_time())
```
orchetect, line 306: May want to add a note to the inline doc here that `time` should never be a value of 0

Sources/AudioKit/MIDI/MIDI+Sending.swift
```diff
@@ -409,50 +348,14 @@ extension MIDI {
...
-
     /// Send Message with data with timestamp
     /// - Parameters:
     ///   - data: Array of MIDI Bytes
-    ///   - time: MIDI Timestamp
-    public func sendMessageWithTime(_ data: [MIDIByte],
-                                    time: MIDITimeStamp,
-                                    endpointsUIDs: [MIDIUniqueID]? = nil,
-                                    virtualOutputPorts: [MIDIPortRef]? = nil) {
+    ///   - time: MIDI Timestamp (default mach_absolute_time())
```
orchetect, line 354: And also here - see other note about inline doc

orchetect:
`MIDITimeStamp` is a typealias of UInt64, so constructing it is redundant.

`MIDITimeStamp(mach_absolute_time())` can just be `mach_absolute_time()`

Matt54:
Thanks for the comments @orchetect! Committed the changes

orchetect:
Thanks for the PR!

PR #2682, opened 2022-02-22
NodeRecorder: swapped file directory path for URL to support iCloud drive

Matt54:
NodeRecorder: swapped file directory path for URL to support iCloud drive

NodeRecorder still works the same as before when no file directory url is provided. When a file directory url is provided, files are written to this location. I was unable to get it to work with the iCloud documents drive without using the url instead of the path.


aure:
Looks good to me.

Issue #2683, opened 2022-02-27
Calling `AudioPlayer#seek` fires completionHandler


osheroff:
After upgrading to AudioKit from 5.2.2 to 5.3.3, when I call `AudioPlayer#seek()` it fires the `completionHandler` method.  Is this an intentional change?  In previous versions the completionHandler was only called when the end of the track was reached naturally.

osheroff:
specifically this PR https://github.com/AudioKit/AudioKit/pull/2652/files removes the guard that prevents the completion handler from firing in a seek.  Doesn't look super intentionally removed, LMK if you want a PR to put back the seek-suppression-guard

aure:
Go for it, thanks in advance.

emurray2:
> specifically this PR https://github.com/AudioKit/AudioKit/pull/2652/files removes the guard that prevents the completion handler from firing in a seek.  Doesn't look super intentionally removed, LMK if you want a PR to put back the seek-suppression-guard

Yes, that I may have not intentionally removed. What I was trying to get rid of were a bunch of booleans to keep track of states. I switched to enums instead so the player sometimes only has one state (i.e. it can't be paused and playing at the same time).

Feel free to experiment with the code and add it back in somehow if you like.

PR #2685, opened 2022-03-02
restore behavior that stops calling CompletionHandler on seek

osheroff:
https://github.com/AudioKit/AudioKit/issues/2683

FYI I had to move the `isSeeking` guard out of the `internalCompletionHandler`, with the (more correct) approach of calling it on the main dispatch thread it now has a data race with the `isSeeking` guard.

### DIFF
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
@@ -87,8 +87,10 @@ extension AudioPlayer {
         let time = time.clamped(to: 0...duration)
 
         if status == .playing {
+            isSeeking = true
             stop()
             play(from: time, to: duration)
+            isSeeking = false
         } else {
             editStartTime = time
             editEndTime = duration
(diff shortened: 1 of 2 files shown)


Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
```diff
@@ -87,8 +87,10 @@ extension AudioPlayer {
         let time = time.clamped(to: 0...duration)
 
         if status == .playing {
+            isSeeking = true
```
emurray2, line 90: This should work for now, although I'm wondering if it would be possible in the future to have this be part of the player status enums. Also, I wasn't able to find a data race you mentioned, but I'm curious now where that happens. Does Xcode say which line it happens on?

emurray2 (approved):

Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
osheroff: So the race, when I attempted to add the `if isSeeking { return }` logic back inside `internalCompletionHandler` is like:

- we set `isSeeking`
- we dispatch the call to `internalCompletionHandler` onto the main thread
- we unset `isSeeking`
- now `internalCompletionHandler` fires on the main thread, but we're not actively seeking.

emurray2: Ah I see. Yeah I found it best to avoid running certain tasks in the completion handler because it actually runs on a background serial queue. Synchronizing resources can be tricky. The way you did it in the PR should be fine now. I'm going to rework some more code and clean up that completion handler so we can be careful about async tasks.

Let me know if you see anything else which needs polishing, and thanks for catching this one 🙌🏻!

Issue #2684, opened 2022-02-28
CompletionHandler: Unrecoverable state when audio is buffered


ShowtimeSoft:
**Describe the bug**
When the audio is buffered and the completionHandler is triggered, the player gets into a state where PLAY cannot be triggered (player.play does not work).

Demonstration app and Xcode project attached.

I can see in the player.play function line 44, if status is .completed and the audio is buffered, the play command is not handled.
The comment reads:
line 45: // reset the status and play again if isLooping and not buffered

I can also see that though the player.status is .completed, the playerNode is still playing, not sure if that is by intention(?)

**To Reproduce**
1) Make sure to buffer the audio.
2) Play till the completionHandler is triggered.
3) Try to play again.
It doesn't play.

**Expected behavior**
I expect it to play.

**Details (please complete the following information):**
 - Type: macOS
 - OS: 12.2
 - AudioKit Version 5.3.3

**Additional context**
I have tried to manually call player.stop(), but since we have a "guard status == .playing else { return }", stop will not be called if player.status is .completed.

Compiled app for MAC:
[SectionPlayer.zip](https://github.com/AudioKit/AudioKit/files/8152560/SectionPlayer.zip)

Xcode Project (feel free to us it however you want):
[SectionPlayer.zip](https://github.com/AudioKit/AudioKit/files/8152564/SectionPlayer.zip)

emurray2:
@ShowtimeSoft I think it may be fixed now. Feel free to test once it gets merged, or you can do so with my fork. In the meantime, I'll start working on #2672 now :) .

emurray2:
There was a recent commit which also edited some of this logic to help with the `getCurrentTime()` method, so I deleted my fork. I'll refactor everything again with the changes I previously made.

ShowtimeSoft:
I can confirm that I can now let my completionHandler trigger a re-play.

PR #2686, opened 2022-03-02
Address audio player reset logic

emurray2:
Allow audio player to be reset to `.stopped` status after `play()` is called again if audio player is still in `.completed` status - AudioKit/AudioKit#2684


emurray2:
I think I'll close this and re-fork to avoid conflicts

PR #2687, opened 2022-03-03
stop the playerNode upon completion

osheroff:
Without this call the playerNode continues its timeline in the background and subsequent calls to `getCurrentPosition()` are incorrect. 

to reproduce the issue I'm talking about, simply do like:

```
player.play()
... wait for the player to stop naturally, then wait a few seconds

player.seek(0)
player.play()
print(player.getCurrentTime())
```

the call to getCurrentTime() will be off by the amount of seconds you waited (I think).


emurray2 (commented):
I also cleaned up this bit in #2686 because the last part with the `else if` wasn't needed and prevented the player from being able to play again after `.completed`

PR #2689, opened 2022-03-06
set .stopped status before stopping

osheroff:
If we don't do this, we can race with the completion handler, which says "ah, I'm stopping and the status is still `.playing`.  let's fire."  

we could also introduce an intermediate `.stopping` state, but I don't think it's necessary.


071088899 (commented):
Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift

PR #2699, opened 2022-04-14
Add a test case to show crash when adding sampler in a submix

jcavar:
Adding a sampler to a submix after some delay causes a crash in `SamplerState::Unref`.

Sampler is very sensitive to being deallocated while being attached to the audio engine. It seems that Mixer makes this even more tricky due to existing issues with adding nodes to an empty mixer.

I am not sure that I have a good idea of how to fix this, but I thought it is worth drawing your attention to it.

We've fixed this in our app by initializing Mixer with a dummy node.
I am not sure if this is something that we can do in AudioKit? Would it make sense to provide a version of Mixer that always has a dummy node as its input? That might simplify internal logic as well.

Side note:

While it seems to be ok to use empty `AVAudioUnitSampler` as a dummy node, I feel it might be a better idea to use `AVAudioUnitPlayer`.

### DIFF
--- a/Tests/AudioKitTests/Node Tests/Mixing Tests/MixerTests.swift
+++ b/Tests/AudioKitTests/Node Tests/Mixing Tests/MixerTests.swift
@@ -45,6 +45,18 @@ extension MixerTests {
         engine.stop()
     }
 
+    func testDoesntCrashWhenAddingSamplersInSubmix() throws {
+        let engine = AudioEngine()
+        engine.output = createMix()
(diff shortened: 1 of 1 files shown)


aure:
Sorry, by deleting the develop branch I've closed this pull request. Can you recreate this PR?

aure:
I tried to add it back myself but it lost your explanation so it would be better if you recreated it. I did merge your other change. Thanks!

PR #2704, opened 2022-05-03
Add a test case to show crash when adding sampler in a submix

jcavar:
Adding a sampler to a submix after some delay causes a crash in SamplerState::Unref.

Sampler is very sensitive to being deallocated while being attached to the audio engine. It seems that Mixer makes this even more tricky due to existing issues with adding nodes to an empty mixer.

I am not sure that I have a good idea of how to fix this, but I thought it is worth drawing your attention to it.

We've fixed this in our app by initializing Mixer with a dummy node.
I am not sure if this is something that we can do in AudioKit? Would it make sense to provide a version of Mixer that always has a dummy node as its input? That might simplify internal logic as well.

Side note:

While it seems to be ok to use empty AVAudioUnitSampler as a dummy node, I feel it might be a better idea to use AVAudioUnitPlayer.

Reopen of: https://github.com/AudioKit/AudioKit/pull/2699


aure (approved):
I just want to see CI run this

aure:
This PR is a year old now so its kinda behind and things are changing for v6 anyway.

Issue #2705, opened 2022-05-03
Approach to useful generic Node start/stop


jcavar:
Hey everyone,

Currently, `start` `stop` functions are extension on `Node`, but are not part of the protocol. 

This makes it tricky to use in a generic way when specific Node overrides `start` `stop` in its own way (e.g. `Reverb`).

For example:

```
let nodes: [Node] = [Reverb(), Chorus(), Fader()]
nodes.forEach { $0.start() }
```

This code sample will call `start` method from `Node` extension and "overridden" version will not get called. 

Using exact type solves this but it makes it difficult to handle in a generic way.

Before I start working on it, I wanted to hear your opinion. There are 2 options as I see them:
1. Add `start`, `stop` to `Node` protocol
2. Add another protocol (e.g. `ToggleNode`)

PR #2706, opened 2022-05-07
Fix start/stop not working properly when used on Node

jcavar:
As explained in https://github.com/AudioKit/AudioKit/issues/2705
start and stop functions are not working correctly when they are
not a part of protocol.


aure:
Looks great, thanks!

Issue #2691, opened 2022-03-12
AudioPlayer.getCurrentTime() does not seem to compensate for whenTime


ShowtimeSoft:
**Describe the bug**
It seems that getCurrentTime() does not compensate for the at:whenTime parameter.

**To Reproduce**
First:
Set whenTime to +2 seconds:
whenTime = AVAudioTime.now() + 2

Then play:
audioPlayer.play(from: playbackStartTime, to: playbackEndTime, at: whenTime)

The getCurrentTime() is now offset by 2 seconds.

**Expected behavior**
I would expect getCurrentTime() to report the actual time position within the audiofile

**Details (please complete the following information):**
 - Type: Mac
 - OS: macOS 12.2
 - AudioKit Version 5.3.2 and latest develop branch

emurray2:
Hmm, seems related to the failing test in #2692. I'll take a look at it this week.

ShowtimeSoft:
Hi,
Any news here, something I can do to help?

Thanks

emurray2:
Hi, my apologies -- I completely forgot.

I know the player's time is based off this:
[https://developer.apple.com/documentation/avfaudio/avaudioplayernode/1390449-playertime](https://developer.apple.com/documentation/avfaudio/avaudioplayernode/1390449-playertime)

which is set whenever `getCurrentTime()` is called.

What I don't know is if this `playerTime` returns the actual position of the playhead **or** the time since the node has been playing. If the second case is true, the behavior you expected will never be observed since `playerTime` is always set to `0.0`, regardless of the position of the player when it starts.

Does that make sense, and is it what you observe to happen? Please let me know. Thanks.

emurray2:
I'm also going to open this up for anyone else to look at, as I have final exams and projects coming up in school for the next 3 weeks.

ShowtimeSoft:
Thanks,
Yeah, that is kind of what I see: it seems that the "when" time is not compensated for, so if I set the "when" time to 2 seconds, getCurrentTime() is off by 2 seconds.

I'm trying to do a multi-track player, and I need to set the "when" time because I'm starting an array of players.
If I don't set the when time to, say 30ms, the players are not started simultaneously and the timing between the tracks is unstable.

Perhaps an array of AudioPlayers is not at all the way to do a multi track player (?).

emurray2:
[...]

PR #2708, opened 2022-05-08
Compensate for scheduled play

emurray2:
Addressing #2691 - Allows `getCurrentTime()` to compensate for the scheduled player time

### DIFF
--- a/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
+++ b/Sources/AudioKit/Nodes/Playback/AudioPlayer/AudioPlayer+Playback.swift
@@ -17,6 +17,10 @@ extension AudioPlayer {
 
         editStartTime = startTime ?? editStartTime
         editEndTime = endTime ?? editEndTime
+        if let whenTime = when, let renderTime = playerNode.lastRenderTime
+        {
+            scheduledTime = whenTime.timeIntervalSince(otherTime: renderTime) ?? 0.0
+        }
 
         guard let engine = playerNode.engine else {
             Log("🛑 Error: AudioPlayer must be attached before playback.", type: .error)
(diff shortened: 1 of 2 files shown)


aure:
Could you add a test for this?

emurray2:
> Could you add a test for this?

It seems the tests don't schedule the player. I get the output `false` for `audio.isSilent` when running the tests with a scheduled player.

```
func testScheduleTime() {
        guard let url = Bundle.module.url(forResource: "TestResources/12345", withExtension: "wav") else {
            XCTFail("Didn't get test file")
            return
        }
        let engine = AudioEngine()
        let player = AudioPlayer()
        engine.output = player
        player.isLooping = true

        let audio = engine.startTest(totalDuration: 2.0)

        do {
            try player.load(url: url, buffered: true)
        } catch let error as NSError {
            Log(error, type: .error)
            XCTFail(error.description)
        }
        var scheduledTime = 0.0
        player.play(from: 0.0, to: player.duration, at: AVAudioTime.now() + 2, completionCallbackType: .dataPlayedBack)
        audio.append(engine.render(duration: 0.1))
        print(player.getCurrentTime())
        print(audio.isSilent)
        audio.append(engine.render(duration: 0.1))
        print(player.getCurrentTime())
        print(audio.isSilent)
        audio.append(engine.render(duration: 0.1))
        print(player.getCurrentTime())
        print(audio.isSilent)
        audio.append(engine.render(duration: 0.1))
        print(player.getCurrentTime())
        print(audio.isSilent)

        let currentTime = player.getCurrentTime()
        //XCTAssertEqual(currentTime, 0.0)

        //testMD5(audio)
    }
```
Seems as if the player ignores `whenTime` during the tests and plays the audio right away. Is there a way I could write a test which waits for the player's schedule?

emurray2:
I'm going to put this on the back-burner and come back to it. I promise I will, but I think I found the jackpot and want to fix some of those realtime player tests. Having those work will help me write a test for this--in fact I think I saw something similar in there.

PR #2710, opened 2022-05-10
Compensate for scheduled play w/ tests

emurray2:
This is again to address #2691 and #2708, but this includes the tests for that. I also refactored the `getCurrentTime()` function so it matches the style of the `play()` function now, using the player's `NodeStatus` enums.


emurray2:
Looks like they pass now :) !

Issue #2634, opened 2021-12-05
Trying to record to m4a format using NodeRecorder


hoopes:
**Describe the bug**
I am attempting to use the `NodeRecorder` to record to a different file format - in this case, `m4a`. The use case is that it will trade cpu usage for disk activity, theoretically reducing battery usage.

**To Reproduce**

I tried to copy `NodeRecorder.createTempFile()`, ([here](https://github.com/AudioKit/AudioKit/blob/main/Sources/AudioKit/Taps/NodeRecorder.swift#L92)) and just replace the `AVFormatIdKey` in the settings:

```
    let filename = UUID().uuidString + ".m4a"
    let fileUrl = fm.tempDir.appendingPathComponent(filename)

    var settings = Settings.audioFormat.settings
    settings[AVFormatIDKey] = kAudioFormatMPEG4AAC
    settings[AVLinearPCMIsNonInterleaved] = NSNumber(value: false)

    let file = try! AVAudioFile(
      forWriting: fileUrl,
      settings: settings,
      commonFormat: Settings.audioFormat.commonFormat,
      interleaved: true)

    // Use the AK recorder
    recorder = try! NodeRecorder(node: node, file: file)
```

(Ignore the forced unwrapping, as it’s just a test so far)

This _appears_ to be doing the right thing - I see in the log:

```
[general] NodeRecorder.swift:record():136:⏺ Recording using format <AVAudioFormat 0x6000034dbde0:  2 ch,  44100 Hz, Float32, non-inter> (NodeRecorder.swift:record():136)
```

However, when I go to finish the recording, I get the error:
```
OpenFromDataSource failed
Open failed
[default]          ExtAudioFile.cpp:193   about to throw 'dta?': open audio file
[avae]            AVAEInternal.h:109   [AVAudioFile.mm:134:AVAudioFileImpl: (ExtAudioFileOpenURL((CFURLRef)fileURL, &_extAudioFile)): error 1685348671
```

That error code (see [here](https://www.osstatus.com/search/results?platform=all&framework=all&search=1685348671)) means `kAudioFileInvalidFileError`.

Further, if i add a breakpoint [here](https://github.com/AudioKit/AudioKit/blob/main/Sources/AudioKit/Taps/NodeRecorder.swift#L45) and look at the file on disk, it _appears_ to be a valid m4a file, but have zero duration and a `N/A` bitrate:

```
[...]

PR #2716, opened 2022-05-15
Add File Open/Close Functions for `NodeRecorder`

emurray2:
Changes:

1. Removed `file` parameter from initialization - this passes the value without a reference, leaving the user with the possible mistake of forgetting to close the file outside the scope of `NodeRecorder`

2. Added `deinit` so temp files can be deleted after the recorder is done being used

3. Added `openFile` function

4. Added `closeFile` function

5. Added a `recordedFileURL` to keep track of the url of the most recently closed file

6. Modified `audioFile` computed variable to return a file for reading based on the `recordedFileURL`

7. Changed the `AVAudioPCMBufferTests` to work with the new `NodeRecorder`

8. Added `NodeRecorder` tests which use `openFile` and `closeFile`

This should now make the test demonstrated in #2634 pass if the new `NodeRecorder` is used accordingly

### DIFF
--- a/Sources/AudioKit/Taps/NodeRecorder.swift
+++ b/Sources/AudioKit/Taps/NodeRecorder.swift
@@ -112,18 +112,7 @@ open class NodeRecorder: NSObject {
             // Keep track of file URL before closing
             recordedFileURL = fileURL
         }
-        guard let engine = self.node.avAudioNode.engine else {
-            Log("Error: Can't close file because NodeRecorder has no engine!")
-            return
-        }
-        // Need to pause rendering because closing the file can interfere with audio thread
-        engine.pause()
         file = nil
-        do {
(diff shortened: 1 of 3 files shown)


emurray2:
Is there a way to get the tests unstuck? I've noticed sometimes a test will keep running like an infinite loop does in Xcode. It seems to happen in CI too. I'm curious what causes this.

aure:
> Is there a way to get the tests unstuck? I've noticed sometimes a test will keep running like an infinite loop does in Xcode. It seems to happen in CI too. I'm curious what causes this.

I restarted the action.

emurray2:
I'm definitely doing something wrong. This doesn't happen with `main`. Don't merge yet, but stay tuned. I'm looking into it :)!

PR #2718, opened 2022-05-18
[AudioPlayer] - Tests for player status

emurray2:
There weren't any tests for the `AudioPlayer` status after two different files are loaded, so I added one.

Changes:
1.  Removed `completed` status.
2. Move looping check inside `completionHandler`
3. Fix `testFileLooping` to use a file which exists.
4. Add tests for AudioKit/Cookbook#97

### DIFF
--- a/Sources/AudioKit/Nodes/NodeStatus.swift
+++ b/Sources/AudioKit/Nodes/NodeStatus.swift
@@ -12,7 +12,5 @@ public enum NodeStatus {
         case paused
         /// The player node is scheduling.
         case scheduling
-        /// The player node completed playback.
-        case completed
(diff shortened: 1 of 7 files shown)


emurray2:
I think the workflow is going to get stuck again on `NodeRecorder`. Should we bypass these tests? Or should I try to get them unstuck again?

aure:
Always try to get tests to pass, but if not at least mark them as expecting to fail https://developer.apple.com/documentation/xctest/expected_failures

emurray2:
I forgot the CI won’t work because the changes are on a different branch. I’ll mark this as draft for now.

PR #2720, opened 2022-05-19
Add async test expectation to `testM4A()`

emurray2:
Use [`XCTestExpectation`](https://developer.apple.com/documentation/xctest/asynchronous_tests_and_expectations) to handle a task such as loading files to `NodeRecorder`. If the task gets stuck and takes too long, the test will fail itself. Next up: adding [Expected Failures](https://developer.apple.com/documentation/xctest/expected_failures) to this workflow.


emurray2:
The test still gets stuck on my Intel Mac. I'll mark this as a draft until I can fix that.

emurray2:
This is an issue deep down with the way the `main` and `IO` threads are communicating. I'll close and re-open a different ticket more related to the issue.

PR #2722, opened 2022-05-20
[`AudioPlayer`] - Tests for player status

emurray2:
Duplicate ticket for #2718, but now CI should be successful:

> There weren't any tests for the `AudioPlayer` status after two different files are loaded, so I added one.

> Changes:
> 1.  Removed `completed` status.
> 2. Move looping check inside `completionHandler`
> 3. Fix `testFileLooping` to use a file which exists.
> 4. Add tests for AudioKit/Cookbook#97

### DIFF
--- a/Sources/AudioKit/Nodes/NodeStatus.swift
+++ b/Sources/AudioKit/Nodes/NodeStatus.swift
@@ -12,7 +12,5 @@ public enum NodeStatus {
         case paused
         /// The player node is scheduling.
         case scheduling
-        /// The player node completed playback.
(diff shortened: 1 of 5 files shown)


aure:
Cool thanks, unless you are working on something you want to get in real quick, I will make a release for all these changes to be in Cookbook by default.

emurray2:
> Cool thanks, unless you are working on something you want to get in real quick, I will make a release for all these changes to be in Cookbook by default.

I think I'm finished now. Sounds like a good plan.

PR #2723, opened 2022-05-21
[`NodeRecorder`] - Pause recording feature

emurray2:
Changes:

1. Add `isPaused` property to `NodeRecorder`
2. Don't write to buffer if `isPaused`
3. Add `pause()` function
4. Add `resume()` function
5. Add tests for this functionality in `RecordingTests`

Also reverted a previous PR because it didn't do much other than break the Cookbook. Sorry 😣. But it works now 🙃!


aure:
Thanks!

Issue #2725, opened 2022-05-24
 Playing a clip does not start if one already playing


emurray2:
Reopening in AudioKit/AudioKit since this is an issue with `AudioPlayer` and not Cookbook
Original ticket: AudioKit/Cookbook#97

emurray2:
Hi @roboldham,

Can you test your example with AudioKit `main` branch and let me know if the new changes work. Thanks.

roboldham:
It does indeed! Yay... across all apps using AudioPlayer , .status() also reporting correctly in all cases

Thanks very much!

PR #2726, opened 2022-05-24
Add tests and resolution for AudioKit/AudioKit#2725

emurray2:
Add tests and resolution for AudioKit/AudioKit#2725

I also forgot to rename my branch again. Sorry about that.


aure:
I don't actually care if you name your branch for quick fixes like this.

PR #2735, opened 2022-06-08
Detach node from the engine on disconnect

jcavar:
Currently nodes are not detached from the engine if last connection is broken which causes memory leak.


aure:
Forget it, I'll just do it.

Issue #2730, opened 2022-05-27
Deadlock starting BaseTap after stopping it.


quantamrhino:
With the latest BaseTap fix (on the main branch) (the handler is no longer called on the main queue), 
doing a start() tap after it was stopped leads to a deadlock. 
The tap in question is the AmplitudeTap. 
Starting/stopping the tap from the main queue seems to solve this. Is this the right 'fix' for this problem?

emurray2:
Yes, stopping things from a background queue (specifically a serial queue in a handler) generally leads to deadlocks because the background queue is busy processing data. You can use the main queue or run the code asynchronously on another queue.

jcavar:
Hey @quantamrhino. Can you share a bit more context:
- Are you starting and stopping from the same background queue?
- What kind of queue is it?
- Do you have a stack trace? What is waiting on what?

aure:
Waiting on more info, reopen when/if that comes in.

PR #2736, opened 2022-06-08
Fix deadlock on tap stop

jcavar:
On removal, tap internally flushes callbacks,
which calls another handle block synchronously.
This causes deadlock as we are locking inside of another lock.

https://github.com/AudioKit/AudioKit/issues/2730


aure:
Merge in my changes to the main branch so the tests will run and hopefully pass.

PR #2738, opened 2022-06-14
Added support for syncing MultiSegmentAudioPlayers

Matt54:
Added support for syncing many MultiSegmentAudioPlayer
- added optional argument for referenceNowTime so that many players can schedule against a common AVAudioTime
- added optional argument for processingDelay so that many players can have enough time to process scheduling before playback


aure:
Nice work

PR #2743, opened 2022-06-23
NodeRecorder - optionally delete recordings

Matt54:
NodeRecorder - Made delete recordings on deinit optional (shouldCleanupRecordings property)

Use case:
When I use the NodeRecorder, I set the fileDirectoryURL so that I am recording to a storage location where I am intending to keep the files (and have them persist past the life of the NodeRecorder).

### DIFF
--- a/Sources/AudioKit/Taps/NodeRecorder.swift
+++ b/Sources/AudioKit/Taps/NodeRecorder.swift
@@ -81,18 +81,18 @@ open class NodeRecorder: NSObject {
     ///   - node: Node to record from
     ///   - fileDirectoryPath: Directory to write audio files to
     ///   - bus: Integer index of the bus to use
-    ///   - rawDataTapHandler: Raw audio data callback
     ///   - shouldCleanupRecordings: Determines if recorded files are deleted upon deinit (default = true)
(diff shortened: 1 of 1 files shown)


Sources/AudioKit/Taps/NodeRecorder.swift
```diff
@@ -80,22 +82,27 @@ open class NodeRecorder: NSObject {
...
     ///   - bus: Integer index of the bus to use
     ///   - rawDataTapHandler: Raw audio data callback
+    ///   - shouldCleanupRecordings: Determines if recorded files are deleted upon deinit (default = true)
     ///
     public init(node: Node,
                 fileDirectoryURL: URL? = nil,
                 bus: Int = 0,
-                rawDataTapHandler: RawAudioDataHandler? = nil) throws {
+                rawDataTapHandler: RawAudioDataHandler? = nil,
+                shouldCleanupRecordings: Bool = true) throws {
```
aure, line 91: I think that you should leave the rawDataTapHandler as the last item in the initializer so that it can be a trailing closure.

Matt54: Good call! Just switched the order

PR #2746, opened 2022-07-05
Implement matrix mixer

jcavar:
NOTE 1:

In order to use matrix mixer, it is required to use different connection format in the part of the chain. This ability was implemented in https://github.com/AudioKit/AudioKit/pull/2741. 
It was the simplest way to get the functionality but it is not very elegant on the use site. You need to create custom node and override `outputFormat`, which leads to boilerplate code.

I am not sure if there is better way around it apart from providing `outputFormat` on each of node implementations.

NOTE 2:

This is probably in some ways specific to our use case. But it should be in place where it might be useful for other people too.


aure:
Whoa cool, I'll have a look!

wtholliday (approved):
This looks great!

PR #2757, opened 2022-08-03
Handle case where BaseTap bufferSize exceeds frameCapacity of buffer.

nakajima:
I was using a tap but running into a crash when the `bufferSize` I specified was greater than the `frameCapacity` of the buffer passed to the callback. The crash was:

```
// *** Terminating app due to uncaught exception 'com.apple.coreaudio.avfaudio', reason: 'required condition is false: length <= _imp->_frameCapacity'
```

The test I added will crash without the change here.

I'm still pretty new to AudioKit and programming at this level in general so if this is totally wrong or a bad idea, no worries about closing it.

### DIFF
--- a/Sources/AudioKit/Taps/BaseTap.swift
+++ b/Sources/AudioKit/Taps/BaseTap.swift
@@ -107,7 +107,7 @@ open class BaseTap {
             self.unlock()
             return
         }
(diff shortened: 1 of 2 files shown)


aure (approved):
Looks good

nakajima:
Looks like CI is stuck? I don't think I have the ability to try to restart it.

PR #2762, opened 2022-08-12
potential approach for safe taps using actors

wtholliday:
This could fix the deadlock we're seeing on CI, while also being a more elegant/thread safe implementation of taps.


wtholliday:
Note that test failures are unrelated to these changes.

aure (approved):

PR #2763, opened 2022-08-13
Fix another deadlock in BaseTap

jcavar:
If we invalidate `handlerBlock` *after* the `lock()` in tap stop,
`handlerBlock`, might get called before being invalidated from internal
engine queue. Since `handlerBlock` acquires the same lock, the
result is a deadlock.


aure:
I'll leave this one to @wtholliday

wtholliday (approved):

PR #2766, opened 2022-08-22
Implement disconnect strategy for node

jcavar:
AudioKit's existing disconnect strategy traverses connections
recursively and reconnects nodes. This is generally preferrable,
but under certain conditions, it can cause problems.

This is mostly due to bugs on Apple's part.

One example of the problem is:

- AudioEngine creates internal multi splitter audio unit for connections
  to multiple nodes
- When dynamically reconfiguring such connection, the chain gets in an
  inconsistent state (Apple bug)
- When disconnecting problematic part of the chain, the app will crash

Under certain conditions, it is ok to simply recursively detach nodes
and engine will break the connections. This is also more optimised as we
don't need to reconnect the part of the chain that we know it is going away.


Sources/AudioKit/Nodes/Node+disconnectStrategy.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// This strategy will not keep any connections from downstream nodes
+    /// For example:
+    /// Mixer1
+    ///          ---> Node1
+    /// Mixer2
+    /// When disconnecting Node1 from Mixer1, Mixer2's connection to Node1
+    /// will not be preserved
+    /// Use only when you are sure that you have only one chain path to node
+    /// that you are disconnecting
+    case deatach
```
aure, line 28: Spelling should be "detach" everywhere this appears.

Sources/AudioKit/Nodes/Node+disconnectStrategy.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// This strategy will keep connections from downstream nodes
+    /// to nodes that are not in current disconnect path
+    /// For example:
+    /// Mixer1
+    ///          ---> Node1
+    /// Mixer2
+    /// When disconnecting Node1 from Mixer1, Mixer2's connection to Node1
+    /// will be preserved
+    case recursive
+    /// Recursively traverse node chain and deatach nodes
```
aure, line 17: Sometimes misspelled in comments which escapes renaming refactoring so check for the word anywhere.

jcavar: Done.

Sources/AudioKit/Nodes/Node+disconnectStrategy.swift
jcavar: Done.

aure:
I really want @wtholliday to have a look since he authored this connection mechanism.

jcavar:
Sounds good! More eyes, the better.

Please just note that existing default connection handling not touched and this is opt in.

wtholliday:
> This is mostly due to bugs on Apple's part.

Have you filed those bugs with Apple?

jcavar:
It involves couple of bugs, so let me provide some more context:

We have the following use case:

                  -> Effect   ->
Sampler                              Mixer .... -> Downstream Mixer
                  -----------> 


- We are dynamically changing effects while playing notes
- Effects are composed of various units and many include AudioKit's DryWetMixer


Problem 1: We can't use `Mixer`

The reason why we can't use `Mixer` unit is this bug: https://github.com/AudioKit/SoundpipeAudioKit/pull/19. At the time of writing we thought it was something in AudioKit, but digging deeper found it was Apple's bug and it happens with their units too.

Filed as http://openradar.appspot.com/radar?id=5614107591966720


Problem 2: Sampler cleaned up

To workaround Problem 1, instead of `Mixer`, we use custom version of `DryWetMixer` that have custom way of changing node on one of the input busses. This was working good, but there was a memory leak as we discovered. It was fixed by https://github.com/AudioKit/AudioKit/pull/2735, but that caused Sampler to get cleaned up on effect change.

It comes down to: http://openradar.appspot.com/radar?id=5514004654981120

Problem 3: Crash when removing whole chain from upstream mixer

This was all good, but when we remove whole this subchain from the Downstream Mixer, it crashes. The reason for the crash is that chain is in slightly inconsistent state since there is a memory leak. And AudioKit's connection mechanism tries to connect to a node that it shouldn't be there.

Detach strategy that we implemented here workarounds this.

Slightly related to: http://openradar.appspot.com/radar?id=5616162842869760



All bugs have associated sample code that reproduces the issue. Previously I was attaching them to bug report and I've just recently started putting them on github.

PR #2767, opened 2022-08-22
Re-encoded test resource

agilecollab:
Re-encoding the test resource in the same format as the original has resolved this issue.


agilecollab:
Oh, shoot, wrong account...

PR #2806, opened 2022-12-03
NodeRecorder AudioDataCallback

Matt54:
Added the AVAudioTime as an argument to the callback
(I'm using this to get the time that a recording started writing to the file)

Renamed callback to match new arguments


aure (approved):

Matt54:
@aure sorry about that - fixed the test

PR #2811, opened 2022-12-23
Add a workaround for connection format bug

jcavar:
There is a bug when using destination point API and providing only one point:
http://openradar.appspot.com/radar?id=5490575180562432

Workaround it by using `connect` API in this specific case.


aure (approved):

jcavar:
Please don't merge this yet. I think this broke something else.

jcavar:
@aure I think this is a good change and can be merged.

I got caught up on:
https://www.mail-archive.com/coreaudio-api@lists.apple.com/msg01143.html

and after setting channel layout "correctly" everything is working ok.

I've updated header docs for `MatrixMixer` to document that this needs to be done.

PR #2815, opened 2023-01-08
Introduce a callback queue on a tap

### LINKED ISSUE #2814: TSAN data race
### macOS Version(s) Used to Build

macOS 13 Ventura

### Xcode Version(s)

Xcode 14

### Description

image

To reproduce, run tests with TSAN enabled

@josi

### Crash Logs, Screenshots or Other Attachments (if applicable)

_No response_

jcavar:
Fixes #2814, but it feels clumsy as we have two syncronisation mechanisms - locks and queue.

We should probably rewrite a tap to work on one dispatch queue and remove all the locks.

### DIFF
--- a/Sources/AudioKit/Taps/AmplitudeTap.swift
+++ b/Sources/AudioKit/Taps/AmplitudeTap.swift
@@ -44,13 +44,14 @@ public class AmplitudeTap: BaseTap {
                 bufferSize: UInt32 = 1_024,
                 stereoMode: StereoMode = .center,
                 analysisMode: AnalysisMode = .rms,
+                callbackQueue: DispatchQueue,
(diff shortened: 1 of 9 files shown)


wtholliday:
Why don’t we call all the tap callbacks on the main queue?

jcavar:
Hey @wtholliday, I've added a comment on the issue, but copying it here too:

I think we should try to avoid dispatching on the main queue from the framework.

In case of e.g. `AmplitudeTap`, calculation will be happening on the main queue, which might not be desirable.

If framework does things on the main queue, you have no choice as a consumer. If it doesn’t, and you need it, it is easy to switch to main queue at a call-site.

But, putting that aside, I don’t think this solves it either. You might be doing operations on the tap (stop, start) from another queue (as we do).

PR #2818, opened 2023-01-16
MIDIKit conversion

aure:


Sources/MIDI/MIDIPlayer.swift
```diff
@@ -4,41 +4,7 @@
...
-
-    /// Rewind the sequence
-    public func rewind() {
-        currentPositionInBeats = 0
-    }
-}
+import MIDIKitIO
 
 /// Simple MIDI Player based on Apple's AVAudioSequencer which has limited capabilities
 public class MIDIPlayer: AVAudioSequencer {
```
wtholliday, line 10: Let's get rid of this. It uses AVAudioEngine. Plus subclassing... ew!

orchetect: @wtholliday MIDISampler has had a little work done but it's one of the bigger things I want to revisit this week and rebuild

wtholliday: There's no "if" ... we can't support AVAudioSequencer.

Issue #2819, opened 2023-01-25
Memory Leak in MIDI+Sending.swift


alenicol:
### macOS Version(s) Used to Build

macOS 13 Ventura

### Xcode Version(s)

Xcode 14

### Description

Hi,
apologies for not making the Bug Report as complete as requested, however it's such a small thing that I did not think it was worth it.

I found a memory leak in MIDI+Sending.swift

In line 359:
        let packetListPointer: UnsafeMutablePointer<MIDIPacketList> = UnsafeMutablePointer.allocate(capacity: 1)

This memory is never deallocated. By adding at line 384:
         packetListPointer.deallocate()

The problem should be solved. Please double check that this does not have ramifications or bad consequences in the rest of the code

orchetect:
MIDI code is completely stripped out and replaced for v6.

aure:
I closed the issue but then I thought, okay if someone wants to fix it they can, for v5.

orchetect:
Can close if this is now resolved.

PR #2821, opened 2023-01-28
Deallocate `packetListPointer`

emurray2:
#2819


emurray2:
Looks like the `AVAudioSourceNode` of `PlaygroundNoiseGenerator` got CI's attention. I don't think this is related to MIDI.

PR #2836, opened 2023-04-19
Remove files of FormatConverter

cgetc:


cgetc:
I created a PR with the wrong repository

PR #2842, opened 2023-05-23
silenced deprecation warnings

mlostekk:


aure (approved):

NickCulbertson:
I see where assign is deprecated. I'm not sure why `.update` is throwing an error. https://developer.apple.com/documentation/swift/unsafemutablepointer

mlostekk:
Oh, darn. Maybe its not backwards compatible. Which targetSDK are you compiling for?

aure:
I guess we can revert this change.

NickCulbertson:
Updating Xcode fixed it. Unfortunately, I'm not sure what version I was on previously. I think it was Xcode Version 14.1

mlostekk:
Revert PR opened [here](https://github.com/AudioKit/AudioKit/pull/2843). The [documentation](https://developer.apple.com/documentation/swift/unsafemutablepointer/assign(from:count:)?changes=latest_minor) is not very clear so lets revert to keep backward compat.

PR #2847, opened 2023-06-30
Fixes two control-flow typos in `func convertToPCM()`

iandundas:
## Two fixes in `func convertToPCM(){}` 
[file here](https://github.com/AudioKit/AudioKit/blob/618df54cf5a31ae267d3957cf8a2f03a137f16cf/Sources/AudioKit/Audio%20Files/Format%20Converter/FormatConverter%2BPCM.swift#L8C29-L8C29)

#### Fix 1: 
There is a `defer{}` which calls `closeFiles()`, but later it's called again further down the function - so executes twice.

CleanShot 2023-06-30 at 12 19 07@2x

CleanShot 2023-06-30 at 12 19 53@2x

#### Fix 2: 
This one is a bit serious: when there's an error (_see below_), the `completionHandler` is first called with the error, but then later executes the "no errors" path and calls `completionHandler?(nil)`, so both paths execute in the client code:

CleanShot 2023-06-30 at 12 22 39@2x

### DIFF
--- a/Sources/AudioKit/Audio Files/Format Converter/FormatConverter+PCM.swift
+++ b/Sources/AudioKit/Audio Files/Format Converter/FormatConverter+PCM.swift
@@ -169,7 +169,7 @@ extension FormatConverter {
                 {
                     completionProxy(error: Self.createError(message: "Error reading from the input file."),
                                     completionHandler: completionHandler)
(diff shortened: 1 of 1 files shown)


jordibruin:
Thanks for this!

aure (approved):

PR #2861, opened 2023-09-28
Add possibility to create AudioComponentDescription for non-realtime effects

silvansky:
Apple uses `kAudioUnitType_FormatConverter` type for non-realtime effects like time stretching ones. So I added missing `init`s with this type to `AudioComponentDescription`.


aure (approved):

Dilanvar (commented):
Ahh

PR #2862, opened 2023-09-28
Add flag to allow bypass of MIDI enabling on sampler

jcavar:
Enabling midi on sampler instance is not always desirable. If you have MIDI chain with MIDI effects, the sampler should receive transformed events and not the original ones.

This preserves backwards compatibility by defaulting to `shouldEnableMIDI` to `true`


jcavar:
Absolutely yes! Thank you!

PR #2883, opened 2023-12-06
Make recorded file URLs public in NodeRecorder.swift

NickCulbertson:
This allows us to access the recorded files from our main class.

### DIFF
--- a/Sources/AudioKit/Taps/NodeRecorder.swift
+++ b/Sources/AudioKit/Taps/NodeRecorder.swift
@@ -63,7 +63,7 @@ open class NodeRecorder: NSObject {
 
     private var recordedFileURL: URL?
 
-    private static var recordedFiles = [URL]()
+    public static var recordedFiles = [URL]()
 
     /// Callback type
     public typealias AudioDataCallback = ([Float], AVAudioTime) -> Void


misteu:
TL;DR

Consider making it `private(set) static var recordedFiles = [URL]()` instead.

Or would you need a public setter as well?

I think it general it makes sense to make it accessible, in the current state I am also not sure for what the `recordedFiles` are used for anyway. The newly recorded `URL`s are appended but nothing is done with them (apart from removing them via that cleanup method)

I'm actually only using `record()` which creates an `internalAudioFile` if needed. And access it's `URL` via `audioFile` and I never bothered about this array but it still might make sense to have it accessible.

Matt54 (approved):
The only reason I can think of to keep it private is to protect the url that's currently being recorded to since we append that url from the moment we create a new file in createAudioFile

I could see private(set) being more appropriate, but perhaps that is opinionated. I could see a situation where someone would want to remove a specific recording (some type of cancel operation).

I'd personally be fine with us just making it public and letting developers handle it how they would like.

@NickCulbertson I'm currently bookkeeping these files by grabbing a reference to nodeRecorder.audioFile?.url right before calling .stop() and .createNewFile(). It's a little clunky, but it works for me.

Matt54:
Just seeing this PR from December, sorry it slipped through the cracks @NickCulbertson! Got a notification for it this morning thanks to @misteu's comment.

aure (approved):

Issue #2881, opened 2023-11-16
Unit tests failing MD5 validation


ghislainleblanc:
### macOS Version(s) Used to Build

macOS 13 Ventura

### Xcode Version(s)

Xcode 14

### Description

(Note that this result is actually on MacOS 14.1.1, Xcode 15.0.1, which were not options that can be selected)

I imagine this means something has changed at the OS level? I'd be happy to update the MD5s accordingly.

The following UTs are failing MD5 validation:
testEffects(): XCTAssertTrue failed - AudioKit.Decimator
FAILEDMD5 5dc2828d3ae812e780e6778d2d075615

testFilters(): XCTAssertTrue failed - AudioKit.HighShelfFilter
FAILEDMD5 2c058606c0e5fb7c6e245eb0f098366d

testEffects(): XCTAssertTrue failed - AudioKit.Distortion
FAILEDMD5 0135e68b5e4a208d33b4ef44a4479e4c

testFilters(): XCTAssertTrue failed - AudioKit.LowShelfFilter
FAILEDMD5 197e9f80a2e1fecb7492845fb16ca8c8

testEffects(): XCTAssertTrue failed - AudioKit.RingModulator
FAILEDMD5 b1acb3ec51f58d18d0d5df2d2c073a50

testEffects(): XCTAssertTrue failed - AudioKit.Reverb
FAILEDMD5 ac8d2c81f0c74217d3fff003cbf28d68

testHeadRoom(): XCTAssertTrue failed - 
FAILEDMD5 "-[CompressorTests testHeadRoom]": "db27f010ec481cd02ca73b8652c4f7c1",

testNodeConnection(): XCTAssertTrue failed - 
FAILEDMD5 "-[NodeTests testNodeConnection]": "6df5064dfb23635516cc418cbebb2e0d",

aure:
The main issues with MD5s is that the Github Action continuous integration is indeed still passing, so it could be that the new MD5s are specific to certain OS or even certain processors. It probably wouldn't hurt to add the MD5s you're finding to the list of acceptable MD5s for any given test.

aure:
You create a fork and then branch off of that, then you can make a pull request.

PR #2886, opened 2023-12-21
Fix unit tests to run correctly on Apple Silicon

ligaz:
This change fixes the failing unit tests when run on Apple Silicon.

Fixes: #2881.


aure (approved):
If it passes CI this is fine by me

PR #2911, opened 2024-04-24
Copy packets to prevent AddressSanitizer: stack-use-after-return on address

kirsteins:
Enabling the "Address Sanitizer" with the "Detect use of stack after return" feature has exposed a stack-use-after-return issue in the following code example:

```swift
import AudioKit
import CoreMIDI

private func createTestMIDIPacketList() -> MIDIPacketList {
    let midiData: [UInt8] = [0x90, 60, 90]
    let packetListPointer = UnsafeMutablePointer<MIDIPacketList>.allocate(capacity: 1)
    var packet = MIDIPacketListInit(packetListPointer)
    let timeStamp: MIDITimeStamp = 0
    packet = MIDIPacketListAdd(packetListPointer, 512, packet, timeStamp, midiData.count, midiData)
    return packetListPointer.pointee
}

func triggerAddressSanitizerWithUseOfStackAfterReturn() {
    for packet in createTestMIDIPacketList() {
        // Accesses packet via iterator and cause
        // AddressSanitizer: stack-use-after-return on address
    }
}
```

This pull request copies all MIDIPacket instead of iterating them one by one and fixes the issue with address sanitizer.

Please discuss if this is a good idea.

### DIFF
--- a/Sources/AudioKit/MIDI/Packets/MIDIPacketList+SequenceType.swift
+++ b/Sources/AudioKit/MIDI/Packets/MIDIPacketList+SequenceType.swift
@@ -15,22 +15,27 @@ extension MIDIPacketList: Sequence {
     /// Create the sequence
     /// - Returns: Iterator of elements
     public func makeIterator() -> AnyIterator<Element> {
-
-        withUnsafePointer(to: packet) { ptr in
+        // Copy packets to prevent AddressSanitizer: stack-use-after-return on address
+        let packets = withUnsafePointer(to: packet) { ptr in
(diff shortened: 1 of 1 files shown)


kirsteins:
Tagging @wtholliday since you fixed the previous issue with address sanitizer in this file. Feedback welcomed and appreciated!

aure (approved):
I'll approve because MIDI is going to be very different in forthcoming versions of AudioKit anyway, so this seems fine to do for now, but I'll let @wtholliday chime in if he wants to, before merging later today.

Issue #2921, opened 2024-06-15
AudioKit use a deprecated Method that is no longer available in VisionOS 2.0.


Ivan0195:
### macOS Version(s) Used to Build

macOS 13 Ventura

### Xcode Version(s)

Xcode 14

### Description

AudioKit use a deprecated Method that is no longer available in VisionOS 2.0.
<img width="673" alt="339558652-8824f6d3-bbc7-41db-9af8-86bb013a99f3 (1)" src="https://github.com/AudioKit/AudioKit/assets/93259552/1b84599e-ecc8-4ca7-b7f8-257b465acdcd">

itsthisjustin:
Crap this is gonna be bad for us

1998code:
Same +1 🙈

itsthisjustin:
> Fixed @Ivan0195 @itsthisjustin 
> 
> https://github.com/AudioKit/AudioKit/pull/2923

Amazing! Idk if anyone else is having this issue but as of ios18 beta 1, the audio quality on recorded mics using AudioKit is reallllly bad. Press play here for example: https://www.cleft.ai/shared/4xDns6Nocy

1998code:
@itsthisjustin Not sure 🤔 I only used AudioKit to play audio which works fine.

Matt54:
@itsthisjustin - I have experience the same issue with mic input. I fixed this in my project by adding Settings.sampleRate = 48_000 during my audio session configuration, but I'll look into a solution for the AudioKit codebase

itsthisjustin:
Confirmed we just fixed it this way as well but would love a more official update maybe?

Matt54:
Hopefully coming very soon 🙂

kokashking:
what is the status on the issue?

antiero:
Looking to start some visionOS work and keen to know if this fixed/supported in the latest visionOS release?

PR #2923, opened 2024-06-24
Fix FormatConverter+Compressed on iOS 18, macOS 15, tvOS 15 and VisionOS 2

1998code:
Use export(to:as:) instead.

Remark: 
The old one was fully deprecated on visionOS 2 and Xcode 16. Failed to fallback.
As a workaround, I used ```#if !os(visionOS)...``` to prevent crashes.
Screenshot 2024-06-24 at 2 15 14 PM

Ref Doc:
https://developer.apple.com/documentation/avfoundation/avassetexportsession/4354409-export

### DIFF
--- a/Sources/AudioKit/Audio Files/Format Converter/FormatConverter+Compressed.swift
+++ b/Sources/AudioKit/Audio Files/Format Converter/FormatConverter+Compressed.swift
@@ -35,7 +35,8 @@ extension FormatConverter {
         session.determineCompatibleFileTypes { list in
             Task.init {
                 guard let outputFileType: AVFileType = list.first else {
-                    let error = Self.createError(message: "Unable to determine a compatible file type from \(inputURL.path)")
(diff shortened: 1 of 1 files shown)


1998code:
Link to issue: https://github.com/AudioKit/AudioKit/issues/2921

1998code:
@aure / @wtholliday Please help review this PR. Thanks.

warpling:
Bump :)
Thanks for doing this @1998code!

1998code:
Welcome @warpling 
I only tested on visionOS 2.0 beta.

Matt54 (changes requested):

Matt54:
I think a good stop gap solution here would be to hide the existing method from visionOS and then create a new method that works across all platforms using the modern concurrency version of AVAssetExportSession's export. This way we don't break the current api.

PR #2935, opened 2024-10-11
Update defaultAudioFormat for iOS 18 to 48_000

NickCulbertson:
This PR sets a better defaultAudioFormat for Settings.sampleRate on iOS 18 and newer. Tests have shown that the current sampleRate of 44_100 causes audio glitches when recording from the mic input. This PR changes the default in iOS 18 to `AVAudioFormat(standardFormatWithSampleRate: 48_000, channels: 2) ?? AVAudioFormat()` while keeping earlier versions, Mac, and iOS on Mac unchanged.

Tests show the following:

iOS Simulator
**48_000** & **44_100** both work

Sequoia macOS Catalyst 
**44_100** works

iOS 17 iPad 
**48_000** & **44_100** both work

iOS 18 iPad
**48_000** & **44_100** both work

iOS 17 iPhone
**44_100** works

iOS 18 iPhone
**48_000** works

### DIFF
--- a/Sources/AudioKit/Internals/Settings/Settings.swift
+++ b/Sources/AudioKit/Internals/Settings/Settings.swift
@@ -52,8 +52,22 @@ public class Settings: NSObject {
     }
 
     /// Default audio format
-    public static let defaultAudioFormat = AVAudioFormat(standardFormatWithSampleRate: 44_100,
-                                                         channels: 2) ?? AVAudioFormat()
+    public static let defaultAudioFormat: AVAudioFormat = {
+        if #available(iOS 14.0, *) {
+            if ProcessInfo.processInfo.isMacCatalystApp || ProcessInfo.processInfo.isiOSAppOnMac {
+                // iOS app on Mac
+                return AVAudioFormat(standardFormatWithSampleRate: 44_100, channels: 2) ?? AVAudioFormat()
+            }
+        }
+
+        // iOS 18 handling
+        if #available(iOS 18.0, *) {
+            return AVAudioFormat(standardFormatWithSampleRate: 48_000, channels: 2) ?? AVAudioFormat()
+        } else {
+            // Fallback default
+            return AVAudioFormat(standardFormatWithSampleRate: 44_100, channels: 2) ?? AVAudioFormat()
+        }
+    }()
 
     /// The sample rate in Hertz, default is 44100 kHz. Set a new audioFormat if you want to change this value.
     /// See audioFormat. This is the format that is used for node connections.


aure (approved):

quantamrhino:
This commit is causing problems on iPads with 44.1KHz sample rate. The Audio Engine fails to initialize with error "AVAEInternal.h:109   [AVAudioEngineGraph.mm:1389:Initialize: (err = AUGraphParser::InitializeActiveNodesInOutputChain(ThisGraph, kOutputChainOptimizedTraversal, *GetOutputNode(), isOutputChainActive)): error -10868"
There's no workaround for this problem (by setting Setting.audioFormat in init, for example).

This code, on iPads executes in the if #available(macOS 15) branch in this case! 

`public static let defaultAudioFormat: AVAudioFormat = {
        if #available(iOS 18.0, *) {
            if !ProcessInfo.processInfo.isMacCatalystApp && !ProcessInfo.processInfo.isiOSAppOnMac {
                /// Default AVAudioFormat for iOS 18 and newer (not on Mac)
                return AVAudioFormat(standardFormatWithSampleRate: 48_000, channels: 2) ?? AVAudioFormat()
            }
        }

        if #available(macOS 15.0, *) {
            /// Default AVAudioFormat for macOS 15 and newer
            return AVAudioFormat(standardFormatWithSampleRate: 48_000, channels: 2) ?? AVAudioFormat()
        }

        /// Fallback default
        return AVAudioFormat(standardFormatWithSampleRate: 44_100, channels: 2) ?? AVAudioFormat()
    }()`
Perhaps defaultAudioFormat should be put in the iOS/macOS variants that are protected by the #if !os(macOS) and #if os(macOS) respectively?

NickCulbertson:
I think the sample rate can also be different while using external hardware.

When I initially tested using the Cookbook on iOS 18, 48_000 was needed for iPhone, iPad, and Mac. I noticed that as iOS 18 progressed Mac could use either one, but now works better on 44_100. I haven't retested on iPad, but it sounds like Apple reverted whatever changes they made with the default sample rate. 

We tried setting the default sample rate to AVAudioSession's sample rate, but I remember that not working correctly which is why I set them explicitly.

At very least we could test out removing: ``` if #available(macOS 15.0, *) {
            /// Default AVAudioFormat for macOS 15 and newer
            return AVAudioFormat(standardFormatWithSampleRate: 48_000, channels: 2) ?? AVAudioFormat()
        }```

@quantamrhino So, are some iPads now on 44_100 and others on 48_000?

NickCulbertson:
iOS 17 and older use 44.1k. That was the default set for all sampleRates previously.

It's unclear which sampleRates now need to be set to 48k in iOS 18 and newer. I think the iPhone needs 48k, Mac works with both, but 44.1k works better while recording, and iPad was working with these changes, but I have not retested in iOS 18.1+.

What changes are you recommending? Ideally, we would pick the best default for each device AND match whatever the device's inputs and outputs are set to. This PR initially tried to solve that first problem.

quantamrhino:
I'm not really sure ... what would you recommend? My fear is this:
let's say older ipads with mic sampling rate of 44.1, always set to audioFormat of 48. will that cause issues? 
I thought it would change the pitch, but that doesn't seem to be happening in my test anymore ... odd!

quantamrhino:
If I set Settings.audioFormat before creating my first node, things dont work and the same error (in the first message) shows up again. But If I change the let to a var in public static var defaultAudioFormat, and assign to default audio format before creating my first node, it works! Can we change defaultAudioFormat to a var? what will that impact?

NickCulbertson:
Yeah, I've noticed more changes in the sampleRate requirements too. For example, I was convinced that 44.1k was working better on macOS 15, but when I just tested again 48k is now working and 44.1k is having issues. My best guess is that either I've done something system wide to change the sampleRate or **Apple is continuing to mess with the sauce**. 

```
let's say older ipads with mic sampling rate of 44.1, always set to audioFormat of 48. will that cause issues?
I thought it would change the pitch, but that doesn't seem to be happening in my test anymore ... odd!
```

When I tried setting older devices pre-iOS 18 to 48k it did cause issues. That was the reason for this PR in the first place. I think it has more to do with the OS than the device. My iPhone 12 Pro went from needing 44.1k to 48k after iOS 18. 

I didn't notice pitch shifting issues like you described, but instead constant audible pops if the sampleRate was a mismatch.

So, I'm open to optimizing this default picking, but I'm not sure what needs changing. I do think the pre-iOS 18 fallback should still be 44.1k

NickCulbertson:
```
public static var audioFormat = defaultAudioFormat
```

audioFormat is a var and it only uses defaultAudioFormat to set the default. Changing it to a var might be fine, but it doesn't address the larger issue and it might be creating some race condition. Can you share a code sample?

quantamrhino:
You are right .. setting it early enough works. But when I set the audioFormat to 48KHz, and connect to a bluetooth speaker, the Settings.session.sampleRate is seen to be 44.1Khz (same as mic sample rate). Is it possible to set output sample rate and input sample rate high? I also see a much drastic problem with my car play - the mic is set to 24Khz and that's the output format set too, even though i've set it to 48! In playAndRecord, with AudioKit, is it possible to accept input sample rate to 24 but keep output to 48? I'm tearing my hair out over this :-( Any help is appreciated.

NickCulbertson:
I don't see any AudioKit examples or tests that use separate sample rates, so your guess is as good as mine. You could create a separate issue to see if anyone chimes in. I'd just be bouncing things off of ChatGPT at this point. Good luck!

PR #2949, opened 2025-01-08
Fix MatrixMixer tests

jcavar:
Use constant sample rate and output format to avoid side-effects from
previous tests.


aure (approved):
Nice!

PR #2952, opened 2025-03-14
Fixed crash when ramping delay is longer than 4096 samples

maweibezahn:
If sample count of the delay was longer than 4096, it was capped to 4096, but after that assert() checked for < 4096; so this lead to a crash. Also made the warning log a bit more explanatory including delay length, sample rate and resulting original sample count.


aure (changes requested):
Use <= 4096 because 4097 is a weird number

aure (approved):

aure:
Thank you!

PR #2981, opened 2026-03-27
fix: increase FormatConverter test timeout from 9s to 30s

ghost:
Two fixes for CI failures:

1. FormatConverter test timeout bumped from 9s to 30s. AAC encoding takes ~17s locally and longer on CI runners, so the M4A tests always timed out.

2. All `weak let` changed to `weak var` in BaseTapTests and NodeTests. Swift 6.1 on CI rejects `weak let` since the reference can become nil at runtime.


systemblueteam:
@aure ill keep working on it! I can’t remember off the top of my head but there was one CI that I kept an eye on and it went on for over an hour (to be honest I didn’t expect that but given the library it makes sense just didn’t know what my expectation should be on that front). Sorry about the failures!

PR #2983, opened 2026-03-28
fix: test failures in testPlayAfterNonLoopingCompletion and testPlayThreeFiles

ghost:
Fixes the two test failures @aure flagged in #2981. testPlayAfterNonLoopingCompletion rewrote to validate through MD5 instead of asserting on player.status (unreliable in offline rendering). testPlayThreeFiles added engine.stop() before early return to prevent hang.


aure:
@systemblueteam You wanna check the last two commits I made to make sure they didn't bash what you had intended with some of your recent commits? At least the tests are passing and you should be able to use them to validate changes.

systemblueteam:
Looks good, no conflicts. Tests passing.

PR #2984, opened 2026-03-28
fix: prevent MixerTests hangs in CI

ghost:
Two tests in MixerTests hung when run headless. Switched both to offline rendering so they complete reliably without changing what they actually test.


systemblueteam:
Makes sense. Just opened #2985.

Issue #2900, opened 2024-02-26
FormatConverter discards 2nd channel when going from 2 to 1 channels 


quantamrhino:
### macOS Version(s) Used to Build

macOS 13 Ventura

### Xcode Version(s)

Xcode 14

### Description

Using formatConverter.Options as follows:

```
options.format = .caf
options.sampleRate = 48000
options.bitDepth = 16
options.channels = 1 // mono.
```

if I have a stereo file with 1 channel (left), this converts it to a mono file correctly. 
But a stereo file with 1 channel (right) converts it to an empty file.
is this expected behavior - to discard the second channel?

If so, how does one convert a 1-channel stereo file to a mono file (with the non empty channel) as the first channel? (effectively flatten it).

NickCulbertson:
It sounds like it is just grabbing the first channel (left) and making it into a mono file. I wouldn't think that is the expected behavior, but it would take some digging to find where this is occurring and how to mix the two channels. If you want to make a minimal GitHub example I can take a look. 


I think there is a Stereo Limiter AudioKit node that could be used to make an incoming stream Mono.

PR #2993, opened 2026-03-29
fix: mix all channels when converting stereo to mono (#2900)

ghost:
Fixes stereo-to-mono conversion dropping the right channel. Reads as stereo and averages the channels before writing instead of relying on ExtAudioFile's default behavior.


systemblueteam:
Fair. I'll be more thorough. Closing the PR for now.

PR #3004, opened 2026-04-02
Migrate recordingBufferLength from Settings to NodeRecorder parameter

jcavar:
## Summary
- Remove `Settings.recordingBufferLength` static property and add a `bufferLength` parameter to `NodeRecorder.init` instead (defaults to `.veryLong`, same as before)
- Part of the Swift 6 migration — `Settings` is generally problematic due to its nature of static shared mutable state
- This is a breaking change but simple to resolve at callsites by passing the desired buffer length directly to `NodeRecorder`


aure (approved):

aure (changes requested):
Not passing tests...

jcavar:
Sorry about that, passing now

PR #3008, opened 2026-04-22
Add async FormatConverter APIs under Swift6 trait

jcavar:
## Summary

- Part of the Swift 6 migration. To keep the existing library unchanged, async versions of the FormatConverter methods are introduced behind the `Swift6` package trait (`#if Swift6` / `#else`).
- Each async function is structurally identical to its non-async counterpart — same validation, same error messages, same control flow — with `throw` replacing completion-handler error delivery and `await` replacing callback chaining.
- `convertPCMToCompressed` uses a shared private `convertPCMToCompressedImpl` to avoid duplicating its ~200-line body; async and callback versions are thin wrappers around it.
- Fixes two pre-existing bugs where completion handlers were not called on error paths (`convertToPCM` copy-failure, and `convertPCMToCompressedImpl` unsupported-format default case), which would have caused leaked continuations in the async wrappers.
- Tests updated to use `async/await` under the Swift6 trait.

### Methods converted

| Method | Pattern |
|--------|---------|
| `FormatConverter.start()` | `#if Swift6` / `#else` with complete functions |
| `convertCompressed(presetName:)` | `#if Swift6` / `#else` (different platform APIs) |
| `convertCompressed()` | `#if Swift6` / `#else` with complete functions |
| `convertPCMToCompressed()` | Shared `Impl` + async wrapper / callback wrapper |
| `AVAudioFile.extract(to:...)` | `#if Swift6` / `#else` with complete functions |

### DIFF
--- a/Sources/AudioKit/Audio Files/Format Converter/FormatConverter+Compressed.swift
+++ b/Sources/AudioKit/Audio Files/Format Converter/FormatConverter+Compressed.swift
@@ -18,6 +18,30 @@ extension FormatConverter {
     ///
     /// This is no longer used in this class as it's not possible to convert sample rate or other
     /// required options. It will use the next function instead
+    #if Swift6
+    @available(macOS 15, iOS 18, tvOS 18, visionOS 2.0, *)
+    func convertCompressed(presetName: String) async throws {
+        guard let inputURL = inputURL else {
+            throw Self.createError(message: "Input file can't be nil.")
+        }
+        guard let outputURL = outputURL else {
+            throw Self.createError(message: "Output file can't be nil.")
+        }
+
+        let asset = AVURLAsset(url: inputURL)
+        guard let session = AVAssetExportSession(asset: asset,
+                                                 presetName: presetName) else {
+            throw Self.createError(message: "session can't be nil.")
+        }
+
+        let list = await session.compatibleFileTypes
+        guard let outputFileType: AVFileType = list.first else {
+            throw Self.createError(message: "Unable to determine a compatible file type from \(inputURL.path)")
+        }
+
+        try await session.export(to: outputURL, as: outputFileType)
+    }
+    #else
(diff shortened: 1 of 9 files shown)


Sources/AudioKit/Audio Files/Format Converter/FormatConverter+PCM.swift
```diff
@@ -98,6 +98,7 @@ extension FormatConverter {
                 completionHandler?(nil)
             } catch let err as NSError {
                 Log(err)
+                completionHandler?(err)
```
jcavar, line 101: This was existing bug where completion handler was not called in this case

Sources/AudioKit/Audio Files/Format Converter/FormatConverter+Compressed.swift
```diff
@@ -193,6 +240,7 @@ extension FormatConverter {
             formatKey = kAudioFormatLinearPCM
         default:
             Log("Unsupported output format: \(outputFormat)")
+            completionHandler?(Self.createError(message: "Unsupported output format: \(outputFormat)"))
```
jcavar, line 252: Existing bug where completion handler was not called

Tests/AudioKitTests/Tap Tests/RawDataTapTests.swift
```diff
@@ -60,6 +60,7 @@ class RawDataTapTests: XCTestCase {
         // the background Task.
         wait(for: [dataExpectation], timeout: 0)
 
+        engine.stop()
```
jcavar, line 63: This fixes a hanging test that results in a deadlock from engine deallocation

jcavar:
@aure this one is finally passing CI. There was a hanging test which took a bit to resolve

aure (approved):

PR #3009, opened 2026-05-17
fix AudioPlayer replay after completion without stale callbacks stopping playback

robbyt:
PR #2978 introduced a race condition that broke rapid scrubbing on my app.

This PR changes the behavior of the AudioPlayer replay after normal (non-looping) completion without letting stale completion callbacks stop newly rescheduled playback.

This PR:
- Allows `play()` to restart when status is still `.playing` but the underlying `AVAudioPlayerNode` is no longer rendering
- Remove destructive playback-state cleanup from completion delivery
- Add regression coverage for stale completion arrival after seek/reschedule
- Keep the existing non-looping replay behavior covered by tests


robbyt:
I found some other issues with this approach, redesigning...

PR #3014, opened 2026-07-15
Fix macOS build with Xcode 27

AvdLee:
## Summary
- preserve AudioKit's macOS 11 deployment target when compiling with Xcode 27
- access the Objective-C `AUAudioUnit` property through its original selector, avoiding the Swift overlay's incorrect macOS 13 availability
- route existing audio-unit access through the compatibility accessor without changing runtime behavior

Xcode 27 beta 3 imports `AVAudioNode.auAudioUnit` as macOS 13+, while the Objective-C SDK declaration remains available from macOS 10.13. This causes AudioKit to fail compilation because the package supports macOS 11.

## Test plan
- [x] `xcodebuild -scheme AudioKit -destination 'platform=macOS' build` with Xcode 27 beta 3
- [x] targeted compressor and FFT tap tests with Xcode 27 beta 3
- [x] full suite compiles and executes 295 tests with Xcode 27 beta 3; 284 pass, 9 skip, and two existing node-deallocation tests report four assertions under the beta runtime
- [x] those two node-deallocation tests pass unchanged on upstream `main` with Xcode 26.6

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@ import PackageDescription
 
 let package = Package(
     name: "AudioKit",
-    platforms: [.macOS(.v13), .iOS(.v13), .tvOS(.v13), .visionOS(.v1)],
+    platforms: [.macOS(.v11), .iOS(.v13), .tvOS(.v13), .visionOS(.v1)],
     products: [.library(name: "AudioKit", targets: ["AudioKit"])],
     targets: [
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@ import PackageDescription
 
 let package = Package(
     name: "AudioKit",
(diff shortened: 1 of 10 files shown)


aure (approved):

Vincent-Loi:
@aure will we do a release soon? Currently having to re-point SPM to the `main` branch rather than a fixed version

robbyt:
The `perform(selector)` accessor here changed object lifetime: the +0 result sits in the caller's autorelease pool and keeps the node's AVAudioUnit alive, which broke `NodeTests.testAllNodesInChainDeallocatedOnRemove` / `testInnerNodesThatHaveMultipleInnerConnectionsDeallocated` on main. #3019 keeps the Xcode 27 / macOS 11 compatibility goal and restores the original lifetime (direct `auAudioUnit` where available, pool-wrapped selector fallback for macOS 11–12).

Issue #2995, opened 2026-03-29
Preserve play/pause state when seeking in AudioPlayer


Anywhere-Music-Player:
## Summary

When seeking with `AudioPlayer`, the current playback state is not preserved correctly.

In our app we needed the following behavior:

- If the player is currently playing, seeking should keep playback running from the new position.
- If the player is currently paused, seeking should update the position but remain paused.
- Completion callbacks should not fire incorrectly during seek transitions.

We implemented this in a custom branch of AudioKit, and it significantly improved seek behavior in real app usage.

## Problem

The default behavior made seek handling difficult because it could force playback state changes during seek operations.

In practice, this caused issues such as:

- seeking while paused could resume playback unexpectedly
- seeking while playing could interfere with the expected play/pause flow
- completion handlers could be triggered in situations where the player was seeking or no longer in a valid playback state

This makes it harder to build a predictable audio player UX on top of `AudioPlayer`.

## Expected behavior

`seek(time:)` should preserve the existing playback state:

- playing -> seek -> still playing
- paused -> seek -> still paused

Also, completion callbacks should only run when playback actually completes in a valid playing state, not during intermediate seek/scheduling transitions.

## What we changed locally

In our fork, we adjusted the `AudioPlayer` playback/scheduling behavior roughly along these lines:

- preserve current play/pause state during seek
- avoid forcing `play()` after every seek
- only restart the node after seek if the player was already playing before the seek
- make completion handling more explicit for `play`, `schedule`, and `seek`
- prevent completion callbacks from firing while `isSeeking == true` or when the player is no longer in a valid `.playing` state

## Why this would be useful upstream

This behavior feels more natural for real-world music player apps and gives a more predictable API surface for apps that need custom scrubbing / transport controls.

Right now, preserving state correctly requires patching internals in a fork, but this seems like behavior that could reasonably live in upstream AudioKit as either:

- the default behavior, or
- an opt-in mode / alternate API if backward compatibility is a concern

## Suggested direction

Possible options:

1. Update `seek(time:)` so it preserves playback state by default.
2. Add a dedicated seek API that explicitly preserves state.
3. Clarify/document the intended behavior and expose hooks that make this pattern possible without forking.

## Additional note

Our branch originally came from a real playback UX issue around seek behavior, and after applying these changes the player behavior became much more consistent in production-style usage.

https://github.com/Anywhere-Music-Player/AudioKit/tree/play-pause-state-after-seek

### Describe Alternatives You've Considered

[...]

Issue #2971, opened 2025-12-08
AudioPlayer.play() doesn't restart playback after non-looping audio completes


warpling:
### macOS Version(s) Used to Build

macOS 13 Ventura

### Xcode Version(s)

Xcode 14

### Description

When using AudioPlayer for non-looping audio, calling `play()` after the audio has finished does not restart playback. The player remains in `isStarted=true` and `status=.playing` state even though audio has completed, and subsequent `play()` calls are no-ops.

  Environment:
  - AudioKit version: 5.6.5
  - Platform: iOS 17/18
  - Device: iPhone (physical device)

  Expected Behavior:

  Either:
  1. `isStarted` should return false after non-looping audio completes, OR
  2. `play()` should restart playback from the current position (or beginning) even if isStarted is true

  Actual Behavior:

  After non-looping audio completes:
  - player.isStarted returns true
  - player.status returns .playing
  - Calling player.seek(time: 0) followed by player.play() does not produce audio
  - Must call player.stop() first to reset the state before play() works again

  Minimal Reproduction:

```
  let player = AudioPlayer(url: shortSoundURL, buffered: false)!
  player.isLooping = false

  // First play - works
  player.play()

  // Wait for audio to complete…
  DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
      print("isStarted: \(player.isStarted)")  // prints: true
      print("status: \(player.status)")        // prints: playing

      // Attempt to replay - no audio
      player.seek(time: 0)
      player.play()  // Silent - no audio plays

      // Workaround - stop first
      player.stop()
      player.seek(time: 0)
      player.play()  // This works
  }
```

  Workaround:

  Call `stop()` before `seek()` and `play()` when replaying:

  if player.isStarted {
      player.stop()
  }
  player.seek(time: 0)
  player.play()

Recommendation:
* Fix players to seek back to 0 and stop when finished playing and potentially introduce a setting that allows multiple calls to play to interrupt and restart the player.

PR #3015, opened 2026-07-25
Fix spurious completion callback when seeking during playback

robbyt:
I'm using AudioKit in a macOS DJ app. The app relies on `AudioPlayer.completionHandler` to know when a track ends, and lets the user scrub the timeline while a track is playing (each scrub tick calls `seek(time:)`).

## The bug

Scrubbing during playback would sometimes make the app think the track had ended: the completion handler fired, my app's UI reset the playhead to 0 and marked the deck stopped, however the audio kept playing.

This is a race inside `seek(time:)`. Seek works by calling `playerNode.stop()` and then scheduling a new segment. But `AVAudioPlayerNode.stop()` calls the completion handler of the segment it just killed, and at that moment none of the existing guards catch it:

- the schedule generation hasn't advanced yet (`schedule()` bumps it later in seek),
- `isSeeking` is still `false` (it's currently set *after* `playerNode.stop()`),
- `status` is `.playing`.

So the killed segment's completion looks exactly like a real end-of-file. The user's `completionHandler` fires and `status` gets stomped to `.stopped` mid-seek.

## How the code got here

Three commits set this up, each was ok on its own:

- f98e1fbe0 (#2978) made the completion handler set `status = .stopped` for non-looping playback, so `play()` could restart after a track ended (#2971). Side effect: a stray completion is no longer harmless — it now mutates player state.
- da1165e9e (issue #2995) added the `!isSeeking` guard to the completion handler. This issue asked that completions *not* fire during seek transitions — but the same refactor moved `isSeeking = true` from *before* `playerNode.stop()` to *after* it. That's the window: before this refactor, the flag ordering prevented the stop-callback from getting through.
- dc0b113a2 (my previous PR #3010) added the schedule-generation gate for *stale* completions — callbacks that arrive late, after a new segment is already scheduled. It can't catch this case, because the stop-callback fires before the generation advances.

So this patch is the missing half of #3010, and it restores the pre- #2995 flag ordering — completing what that issue asked for.

## The fix

Set both guards before stopping the node:

```swift
if wasPlaying {
    isSeeking = true
    bumpScheduleGeneration()
}
```

The generation bump retires the old segment before it's stopped, so its completion is rejected no matter when it arrives. `isSeeking` covers the same window and keeps the public seeking state accurate. No public API or behavior changes otherwise — real end-of-file completions still fire exactly as before.

## Tests

Offline rendering (what CI uses) never fires the stop-callback, so the tests inject the killed segment's completion through a small internal test hook placed right before `playerNode.stop()`. Four new tests: the seek case for file, buffered, and buffered+looping playback (all assert the spurious completion is suppressed and the guards are set before the stop), plus one that proves a genuine end-of-file completion still fires exactly once after a seek.

Verified in my app as well: with this patch, scrubbing works reliably; without it, my app's double-seek regression test fails deterministically on current AudioKit `main`.

Refs: #2971, #2978, #2995, #3010.


aure (approved):

robbyt:
Heads-up: the `build-ios` failure on this merge commit (NodeTests deallocation tests) isn't from this change — it's an autorelease-pool lifetime issue introduced by the `audioKitAudioUnit` accessor in #3014, whose own CI run had been cancelled. Bisect + fix in #3019.

Issue #3007, opened 2026-04-17
tvOS support needs to increase to 14 (.isiOSAppOnMac not available in tvOS 13)


jazzychad:
### macOS Version(s) Used to Build

macOS 26.4.1

### Xcode Version(s)

Xcode 26.4

### Description

Currently both Package.swift files specify tvOS 13 as the minimum deployment target for tvOS

<img width="541" height="142" alt="Image" src="https://github.com/user-attachments/assets/e4ba2961-1738-4758-8c2e-b3657b294c2f" />

and

<img width="549" height="144" alt="Image" src="https://github.com/user-attachments/assets/aa097997-c4a4-491f-b4bb-350ac010106f" />

However, in `Sources/AudioKit/Internals/Settings/Settings.swift` line 57 there is a call to `ProcessInfo.processInfo.isiOSAppOnMac` 

https://github.com/AudioKit/AudioKit/blob/main/Sources/AudioKit/Internals/Settings/Settings.swift#L57

This API is not available in tvOS 13, and is only available starting in tvOS 14. 

<img width="442" height="233" alt="Image" src="https://github.com/user-attachments/assets/e1d4159c-c20f-4ad9-b514-efd551fc2869" />

and

<img width="807" height="119" alt="Image" src="https://github.com/user-attachments/assets/b89512bb-48ec-489b-a0c8-bfaa8bdc05ff" />

As a result, AudioKit will not compile for tvOS apps at all, currently.

The fix is to bump the minimum deployment target for tvOS to 14. This also means that the other AudioKit packages (such as AudioKitEX, AudioKitUI, and possible others) need to have their minimum deployment targets also bumped to match (because they depend on AudioKit. Otherwise, there is a dependency minimum platform version error.

The120thWhisper:
Similarly, macOS support string needs to increase to 13.0 to resolve "'auAudioUnit' is only available in macOS 13.0 or newer", but the package using it says it's for 11.0+, but there are no availability guards or raised platform requirements to prevent the package from being compiled against macOS 11/12.

<img width="716" height="250" alt="Image" src="https://github.com/user-attachments/assets/a2368349-b82e-4049-b235-4e310c418a3b" />

<img width="668" height="117" alt="Image" src="https://github.com/user-attachments/assets/2fb9051a-4dfe-41af-b7e7-c5d660e84f81" />

These are the locations where the auAudioUnit is being used.

[...]

PR #3019, opened 2026-08-23
Fix AVAudioUnit leak from the audioKitAudioUnit accessor

robbyt:
## Summary

Fixes a node deallocation regression introduced by #3014. Two `NodeTests` failed because a removed node's `AVAudioUnit` stayed alive until the surrounding autorelease pool drained.

## Cause

#3014 replaced direct access to `AVAudioNode.auAudioUnit` with `perform(_:)` so the project could still target macOS 11 with Xcode 27, which marks the Swift property as macOS 13+. The selector call extended the returned audio unit's lifetime to the caller's autorelease pool, so the test assertions ran before the node was released.

## Fix

- Use `auAudioUnit` directly on macOS 13+ and on non-macOS platforms.
- On macOS 11–12, use the Objective-C selector inside a local `autoreleasepool` so the temporary reference is released before the accessor returns.

## Verification

- Confirmed by bisecting: a0b08066b passes but 1615414ba (#3014) fails.
- `NodeTests` pass on macOS and the iOS simulator.
- The full iOS simulator suite passes: 304 tests, 0 failures.

I could not test the Xcode 27/macOS 11 deployment-target build because I only have Xcode 26. The macOS availability check matches the annotation that prompted #3014.

Refs: #3014, #3015


jcavar:
As I mentioned here https://github.com/AudioKit/AudioKit/issues/3007#issuecomment-5316189617, I think initial workaround can be reverted as it is not necessary anymore.

